PR #15, opened 2021-11-08
1 Ada limit TxOut for Marlowe

nau:
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marlowe/src/Language/Marlowe/Client.hs
+++ b/marlowe/src/Language/Marlowe/Client.hs
@@ -24,46 +24,51 @@
 
 module Language.Marlowe.Client where
 import           Control.Lens
-import           Control.Monad                   (forM_, void)
-import           Control.Monad.Error.Lens        (catching, throwing)
-import           Data.Aeson                      (FromJSON, ToJSON, parseJSON, toJSON)
-import qualified Data.List.NonEmpty              as NonEmpty
-import           Data.Map.Strict                 (Map)
-import qualified Data.Map.Strict                 as Map
-import           Data.Maybe                      (isNothing, maybeToList)
-import           Data.Monoid                     (First (..))
-import           Data.Semigroup.Generic          (GenericSemigroupMonoid (..))
-import qualified Data.Set                        as Set
-import qualified Data.Text                       as T
-import           Data.UUID                       (UUID)
-import           Data.Void                       (absurd)
-import           GHC.Generics                    (Generic)
+import           Control.Monad                    (forM_, void)
+import           Control.Monad.Error.Lens         (catching, throwing)
+import           Data.Aeson                       (FromJSON, ToJSON, parseJSON, toJSON)
+import           Data.Either                      (rights)
+import qualified Data.List.NonEmpty               as NonEmpty
+import           Data.Map.Strict                  (Map)
+import qualified Data.Map.Strict                  as Map
+import           Data.Maybe                       (isNothing, maybeToList)
+import           Data.Monoid                      (First (..))
+import           Data.Semigroup.Generic           (GenericSemigroupMonoid (..))
+import qualified Data.Set                         as Set
+import qualified Data.Text                        as T
+import           Data.UUID                        (UUID)
+import           Data.Void                        (absurd)
+import           GHC.Generics                     (Generic)
 import           Language.Marlowe.Scripts
 import           Language.Marlowe.Semantics
-import qualified Language.Marlowe.Semantics      as Marlowe
-import           Language.Marlowe.SemanticsTypes hiding (Contract, getAction)
-import qualified Language.Marlowe.SemanticsTypes as Marlowe
-import           Language.Marlowe.Util           (extractNonMerkleizedContractRoles)
-import           Ledger                          (CurrencySymbol, Datum (..), PubKeyHash, Slot (..), TokenName,
-                                                  TxOut (..), inScripts, txOutValue)
+import qualified Language.Marlowe.Semantics       as Marlowe
+import           Language.Marlowe.SemanticsTypes  hiding (Contract, getAction)
+import qualified Language.Marlowe.SemanticsTypes  as Marlowe
+import           Language.Marlowe.Util            (extractNonMerkleizedContractRoles)
+import           Ledger                           (CurrencySymbol, Datum (..), PubKeyHash, Slot (..), TokenName,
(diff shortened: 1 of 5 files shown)


marlowe/src/Language/Marlowe/Client.hs
```diff
@@ -367,11 +372,21 @@ marlowePlutusContract = selectList [create, apply, auto, redeem, close]
...
-                marlowePlutusContract
+                let closeContract = do
+                        marloweData <- SM.runStep theClient (slotRange, [])
+                        case marloweData of
+                            SM.TransitionFailure e -> throwing _TransitionError e
+                            SM.TransitionSuccess d -> do
+                                tell $ OK reqId "auto"
+                                marlowePlutusContract
 
+                catching _StateMachineError closeContract $ \err -> do
```
hrajchert, line 385: What type of error should make us try again in two slots? Shouldn't we explicitly match against a particular error?

marlowe/src/Language/Marlowe/Client.hs
```diff
@@ -240,13 +240,18 @@ marlowePlutusContract = selectList [create, apply, auto, redeem, close]
...
+        ownPubKey <- Contract.ownPubKeyHash
         (params, distributeRoleTokens, lkps) <- setupMarloweParams owners contract
         slot <- currentSlot
         let StateMachineClient{scInstance} = mkMarloweClient params
         let marloweData = MarloweData {
                 marloweContract = contract,
-                marloweState = emptyState slot }
-        let payValue = adaValueOf 0
+                marloweState = State
+                    { accounts = AssocMap.singleton (PK ownPubKey, Token adaSymbol adaToken) 1000000
```
hrajchert, line 250: If I understand this correctly, now each role token can have some balance and the contract creator will have the min ada assigned to its public key hash instead of his role, right? This would allow a user to create a contract even if it doesn't have a role assigned. Neat 👍

hrajchert: Could you add a comment and change `1000000` with a more descriptive constant, or even better, a parameter?

marlowe/src/Language/Marlowe/Client.hs
```diff
@@ -102,7 +102,7 @@ data PartyAction
              | WaitForTimeout Slot
              | WaitOtherActionUntil Slot
              | NotSure
-             | CloseContract
+             | CloseContract [Payment]
```
hrajchert, line 105: this is only needed for the `auto` endpoint, right?

hrajchert (commented):
Looking good, I have a couple of questions.

marlowe/src/Language/Marlowe/Client.hs
nau: Correct

marlowe/src/Language/Marlowe/Client.hs
bwbush: While working on [SCP-3099 branch](https://github.com/input-output-hk/marlowe-cardano/tree/SCP-3099) to upgrade `main` to use the latest `plutus-apps`, I encounted test failures for the zero-coupon bond because of the min-ADA requirement. These failures also occur on `anemish/one-ada-limit` with the latest `plutus-apps`, which now correctly enforces the min-ADA requirement.

This isn't well publicized, but min-ADA is greater that 1,000,000 lovelace *if there is a datum hash in the UTxO*, even if there are no native tokens in the UTxO.

    # Min-ADA w/o a datum.
    
    $ cardano-cli transaction calculate-min-required-utxo  \
      --alonzo-era --protocol-params-file testnet.protocol \
      --tx-out addr_test1vq9prvx8ufwutkwxx9cmmuuajaqmjqwujqlp9d8pvg6gupczgtm9j+10000000
    
    Lovelace 999978
    
    
    # Min-ADA w/ a datum.
    
    $ cardano-cli transaction calculate-min-required-utxo  \
      --alonzo-era --protocol-params-file testnet.protocol \
      --tx-out-datum-hash-file example.datum               \
      --tx-out addr_test1vq9prvx8ufwutkwxx9cmmuuajaqmjqwujqlp9d8pvg6gupczgtm9j+10000000
    
    Lovelace 1344798
    
    
    # Node and CLI version.
    
    $ cardano-cli --version
    cardano-cli 1.31.0 - linux-x86_64 - ghc-8.10
    git rev 2cbe363874d0261bc62f52185cf23ed492cf4859

The safest way to compute this is with `Cardano.Api.calculateMinimumUTxO`.

marlowe/src/Language/Marlowe/Client.hs
```diff
@@ -428,7 +443,7 @@ getAction :: MarloweSlotRange -> Party -> MarloweData -> PartyAction
 getAction slotRange party MarloweData{marloweContract,marloweState} = let
     env = Environment slotRange
     in case reduceContractUntilQuiescent env marloweState marloweContract of
-        ContractQuiescent _reduced _warnings _payments state contract ->
+        ContractQuiescent _reduced _warnings payments state contract ->
```
bwbush, line 446: In line 430 above, the min-ADA for a native token will need to be included along with the role token. Also, it looks like explicit min-ADA will also be necessary with lines for 279, as a work-around for the situation that the ledger constraints apparently make line 279 and 290 into separate UTxOs, without sufficient min-ADA for the role token.

marlowe/src/Language/Marlowe/Client.hs
bwbush: Here is a little more information about min-ADA:
1. [`Plutus.Ledger.Index.minAdaTxOut`](https://github.com/input-output-hk/plutus-apps/blob/a6868cde87600051a4cc808777d665d0b859853c/plutus-ledger/src/Ledger/Index.hs#L378-L379) uses 2 ADA as a generic min-ADA.
2. Given the current protocol parameters, [min-ADA](https://mantis.functionally.io/how-to/min-ada-value/) could be as high as 22.5 ADA. This worst-case scenario corresponds to the situation where there are 123 native tokens with 32-byte names on deposit in the Marlowe script.

bwbush (approved):
I reviewed the latest changes, which addressed previous comments on the PR. Everything looks great!

I also ran the tests, which all passed.

PR #19, opened 2021-11-11
Updating Formatting

jhbertra:
One massive formatting update PR:

- [x] Update the stylish-haskell config to match `plutus-apps`
- [x] Switch from purty to purs-tidy (purty causes syntax errors under certain cases)
- [x] Add prettier for formatting JavaScript, HTML, and CSS files.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


.tidyrc.json
```diff
@@ -0,0 +1,10 @@
+{
+  "importSort": "ide",
+  "importWrap": "auto",
+  "indent": 2,
+  "operatorsFile": ".tidyoperators",
+  "ribbon": 1,
+  "typeArrowPlacement": "first",
+  "unicode": "never",
+  "width": 80
```
nau, line 9: Can we do 120 as in Haskell? 80 columns is almost a mockery...

jhbertra: I don't share that preference personally... I prefer vertical spread over horizontal spread. 80 col is pretty standard, and is the default in `prettier`. It makes it nice for stacking panes horizontally, which is a workflow I use quite often. I'd rather keep it at 80 if that's alright.

jhbertra: Keep in mind - the indent size is 2 spaces and purs-tidy does not horizontally align blocks as we do for Haskell. 80 col is plenty wide, particularly if you avoid excessive nesting.

nau: If Hernan is OK, sure. I personally never use split screens, so to me it's always just a waste of space.

jhbertra: 2021-11-22-142204_2559x1164_scrot

This is what my editing layout typically looks like. I wouldn't be able to do this with 3 split panes if the document width was 120. I also use vertical monitors which favors narrower, taller source code.

palas (approved):
Looks good 👍

.tidyrc.json
hrajchert: I tend to prefer 120 over 80 but is not a strong opinion. As long as the format is made automatically Im ok with both options. @bwbush any particular opinion on this?

marlowe-dashboard-client/src/API/Lenses.purs
```diff
@@ -18,28 +18,45 @@ import Type.Proxy (Proxy(..))
...
 import Wallet.Emulator.Wallet (Wallet)
-import Wallet.Types (ContractInstanceId, EndpointDescription, _EndpointDescription)
+import Wallet.Types
+  ( ContractInstanceId
+  , EndpointDescription
+  , _EndpointDescription
+  )
 
-_cicContract :: Lens' (ContractInstanceClientState MarloweContract) ContractInstanceId
+_cicContract :: Lens' (ContractInstanceClientState MarloweContract)
```
hrajchert, line 36: yikes, that is a weird spot to make a line break, but whatever

marlowe-dashboard-client/src/API/Request.purs
```diff
@@ -18,13 +18,15 @@ import Data.Newtype (unwrap)
...
 import Servant.PureScript (AjaxError, ErrorDescription(..))
 
-doPostRequest ::
-  forall m d e.
-  MonadError AjaxError m =>
-  MonadAff m =>
-  DecodeJson d =>
-  EncodeJson e =>
-  String -> e -> m d
+doPostRequest
```
hrajchert, line 21: 👍🏻

marlowe-dashboard-client/src/Capability/Marlowe.purs
```diff
@@ -250,14 +326,27 @@ sendWsMessage msg = do
...
-  followContract walletDetails marloweParams = lift $ followContract walletDetails marloweParams
+  followContract walletDetails marloweParams = lift $ followContract
+    walletDetails
+    marloweParams
   createPendingFollowerApp = lift <<< createPendingFollowerApp
-  followContractWithPendingFollowerApp walletDetails marloweParams followAppId = lift $ followContractWithPendingFollowerApp walletDetails marloweParams followAppId
-  createContract walletDetails roles contract = lift $ createContract walletDetails roles contract
-  applyTransactionInput walletDetails marloweParams transactionInput = lift $ applyTransactionInput walletDetails marloweParams transactionInput
-  redeem walletDetails marloweParams tokenName = lift $ redeem walletDetails marloweParams tokenName
+  followContractWithPendingFollowerApp walletDetails marloweParams followAppId =
```
hrajchert, line 338: mhmhmh, these line breaks also feel a little bit arbitrary... are these made because of a char limit set somewhere? When possible I prefer the formatter to only do alignment and leave the user the choice on when to break lines. If not possible, not a biggie.

hrajchert (approved):
Just a minor note on line breaks, it looks good to me otherwise.

I'd still wait for @bwbush to cast an opinion on the line width

yveshauser (commented):
fine with 80 :+1:

yveshauser (approved):

marlowe-dashboard-client/src/Capability/Marlowe.purs
jhbertra: It is possible... but I think it's also a double-edge sword.

I like formatters to automatically chop lines that are two long, but this one does not chop _all_ function arguments, so you get kind of wonky stuff like this if it's a big long line.

I think in practice, the best way to deal with it is manually break the first argument it the formatter does this, then it will clean it up:

```
someFunc foo bar baz

// formatted
someFunc foo bar
  baz

// Chop before "foo"
someFunc
  foo bar
  baz

// Format again
someFunc
  foo
  bar
  baz
```

I personally wish it could be configured to just go to the last one directly... this is what more opinionated formatters e.g. `elm-format` and `prettier` do.

jhbertra: I am willing to pay the price of some odd line breaks for now though if it means I don't have to deal with any more 300-400 col wide lines in the code! That has always been a major annoyance in the codebase for me.

jhbertra:
@hrajchert I've submitted an issue to `purs-tidy` about line breaks. For now, I'm afraid a small amount of hand-holding will be necessary to avoid cases like the ones you pointed out. https://github.com/natefaubion/purescript-tidy/issues/76

.tidyrc.json
jhbertra: The way I think about line width is like I think about responsive web design: desktop-first designs are better on large screens, but unusable on small ones. mobile-first designs work well on all screen sizes, large and small.

Similarly, vertical scrolling is relatively easy to manage. Horizontal scrolling is a nuisance, as our editors and tools are all optimized for vertical navigation first. If you have a large editor screen, covering more horizontal space might allow you to fit more code on screen, but it makes the code much less navigable for those of us who prefer narrow windows, use split panes, etc. Narrower and taller layouts are perfectly usable by everyone. If you have large amounts of unused screen real estate as a result, you could use more space for other panels in your IDE, or tile a second window like a web browser side-by-side with your editor for example.

bwbush (approved):

PR #26, opened 2021-11-26
SCP-643 - Implement merkleization in Plutus semantics

palas:
This PR adds merkleization to the Plutus semantics and updates:
 - JSON serialisation and deserialisation (in a way that is backwards compatible)
 - Haskell code that depends on the format of inputs

And also:
 - Adds a test for merkleized excution
- Augment QuickCheck generator to include MerkleizedCase

I had to revert the deserialisation implementation to the tree version because Plutus doesn't implement matching on integers.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marlowe/src/Language/Marlowe/Semantics.hs
+++ b/marlowe/src/Language/Marlowe/Semantics.hs
@@ -428,7 +428,7 @@ data ApplyAction = AppliedAction ApplyWarning State
                  | NotAppliedAction
   deriving stock (Haskell.Show)
 
--- | Try to apply a single input contentent to a single action
+-- | Try to apply a single input content to a single action
 applyAction :: Environment -> State -> InputContent -> Action -> ApplyAction
 applyAction env state (IDeposit accId1 party1 tok1 amount) (Deposit accId2 party2 tok2 val) =
     if accId1 == accId2 && party1 == party2 && tok1 == tok2 && amount == evalValue env state val
--- a/marlowe/src/Language/Marlowe/Semantics.hs
+++ b/marlowe/src/Language/Marlowe/Semantics.hs
@@ -42,30 +42,33 @@ and actions (i.e. /Choices/) are passed as
 
 module Language.Marlowe.Semantics where
 
-import           Control.Applicative             ((<*>), (<|>))
-import qualified Data.Aeson                      as JSON
-import           Data.Aeson.Types                hiding (Error, Value)
-import qualified Data.Foldable                   as F
-import           Data.Scientific                 (Scientific)
-import           Data.Text                       (pack)
+import           Control.Applicative                       ((<*>), (<|>))
+import qualified Data.Aeson                                as JSON
+import           Data.Aeson.Types                          hiding (Error, Value)
+import qualified Data.Foldable                             as F
+import           Data.Scientific                           (Scientific)
+import           Data.Text                                 (pack)
 import           Deriving.Aeson
-import           Language.Marlowe.ParserUtil     (getInteger, withInteger)
-import           Language.Marlowe.Pretty         (Pretty (..))
-import           Language.Marlowe.SemanticsTypes (AccountId, Accounts, Action (..), Case (..), Contract (..),
-                                                  Environment (..), Input (..), IntervalError (..), IntervalResult (..),
-                                                  Money, Observation (..), Party, Payee (..), SlotInterval, State (..),
-                                                  Token (..), Value (..), ValueId, inBounds)
-import           Ledger                          (Slot (..), ValidatorHash)
-import           Ledger.Value                    (CurrencySymbol (..))
-import qualified Ledger.Value                    as Val
-import           PlutusTx                        (makeIsDataIndexed)
-import qualified PlutusTx.AssocMap               as Map
-import qualified PlutusTx.Builtins               as Builtins
-import           PlutusTx.Lift                   (makeLift)
-import           PlutusTx.Prelude                hiding (encodeUtf8, mapM, (<$>), (<*>), (<>))
-import           Prelude                         (mapM, (<$>))
-import qualified Prelude                         as Haskell
-import           Text.PrettyPrint.Leijen         (comma, hang, lbrace, line, rbrace, space, text, (<>))
+import           Language.Marlowe.ParserUtil               (getInteger, withInteger)
+import           Language.Marlowe.Pretty                   (Pretty (..))
+import           Language.Marlowe.SemanticsDeserialisation (byteStringToContract)
+import           Language.Marlowe.SemanticsTypes           (AccountId, Accounts, Action (..), Case (..), Contract (..),
+                                                            Environment (..), Input (..), InputContent (..),
+                                                            IntervalError (..), IntervalResult (..), Money,
+                                                            Observation (..), Party, Payee (..), SlotInterval,
+                                                            State (..), Token (..), Value (..), ValueId, getAction,
+                                                            getInputContent, inBounds)
+import           Ledger                                    (Slot (..), ValidatorHash)
+import           Ledger.Value                              (CurrencySymbol (..))
+import qualified Ledger.Value                              as Val
+import           PlutusTx                                  (makeIsDataIndexed)
+import qualified PlutusTx.AssocMap                         as Map
+import qualified PlutusTx.Builtins                         as Builtins
+import           PlutusTx.Lift                             (makeLift)
+import           PlutusTx.Prelude                          hiding (encodeUtf8, mapM, (<$>), (<*>), (<>))
+import           Prelude                                   (mapM, (<$>))
+import qualified Prelude                                   as Haskell
+import           Text.PrettyPrint.Leijen                   (comma, hang, lbrace, line, rbrace, space, text, (<>))
 
 {- HLINT ignore "Avoid restricted function" -}
 
@@ -81,6 +84,8 @@ import           Text.PrettyPrint.Leijen         (comma, hang, lbrace, line, rbr
 {-# INLINABLE giveMoney #-}
 {-# INLINABLE reduceContractStep #-}
 {-# INLINABLE reduceContractUntilQuiescent #-}
+{-# INLINABLE applyAction #-}
+{-# INLINABLE getContinuation #-}
 {-# INLINABLE applyCases #-}
 {-# INLINABLE applyInput #-}
 {-# INLINABLE convertReduceWarnings #-}
@@ -136,6 +141,7 @@ data ApplyWarning = ApplyNoWarning
 -- | Result of 'applyCases'
 data ApplyResult = Applied ApplyWarning State Contract
                  | ApplyNoMatchError
+                 | ApplyHashMismatch
   deriving stock (Haskell.Show)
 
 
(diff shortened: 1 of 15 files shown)


marlowe/src/Language/Marlowe/SemanticsTypes.hs
```diff
@@ -651,6 +692,11 @@ instance Eq Action where
     Notify obs1 == Notify obs2 = obs1 == obs2
     _ == _ = False
 
+instance Eq a => Eq (Case a) where
+    {-# INLINABLE (==) #-}
+    Case acl cl == Case acr cr                       = acl == acr && cl == cr
+    MerkleizedCase acl bsl == MerkleizedCase acr bsr = acl == acr && bsl == bsr
+    _ == _                                           = False
```
bwbush, line 699: What about situations where a `Case` is conceptually equal to a `MerkleizedCase` if the latter deserialized to the former's contract? Does the rest of the code depend upon the semantics of `Eq (Case a)`? Is there a danger that users of Marlowe might assume that merkleized cases could be equal to the equivalent cases?

marlowe/src/Language/Marlowe/SemanticsTypes.hs
```diff
@@ -662,16 +708,12 @@ instance Eq Contract where
     When cases1 timeout1 cont1 == When cases2 timeout2 cont2 =
         timeout1 == timeout2 && cont1 == cont2
         && length cases1 == length cases2
-        && let cases = zip cases1 cases2
-               checkCase (Case action1 cont1, Case action2 cont2) =
-                    action1 == action2 && cont1 == cont2
-           in all checkCase cases
+        && all id (zipWith (==) cases1 cases2)
```
bwbush, line 711: `all id` = `and`

marlowe/src/Language/Marlowe/SemanticsTypes.hs
palas: I am not sure I understand the full extent of your question, but I think they are different to all effects, because they take different kinds of inputs, so they are never interchangeable. The semantics originally depended on `Eq`, but not anymore, we just keep it in case it is useful

marlowe/src/Language/Marlowe/SemanticsTypes.hs
palas: Oh! Thanks. I was looking for a function that did that in Hoogle and couldn't find it for some reason

marlowe/src/Language/Marlowe/SemanticsTypes.hs
palas: We may be able to protect the users from seeing `MerkleizedCase` all together, by not including it in extended Marlowe, and keeping it limited to core Marlowe

marlowe/src/Language/Marlowe/SemanticsDeserialisation.hs
```diff
@@ -8,152 +9,468 @@ import           Language.Marlowe.SemanticsTypes  (Action (..), Bound (..), Case
...
-  where withOneObs = withOne byteStringToObservation
-        withTwoObs = withTwo byteStringToObservation
-        withChoiceId = withOne byteStringToChoiceId
-        withTwoValues = withTwo byteStringToValue
-        withOne bsToOne bs f = do (z, t) <- bsToOne bs
-                                  return (f z, t)
-        withTwo bsToOne bs f = do (y, t1) <- bsToOne bs
-                                  (z, t2) <- bsToOne t1
-                                  return (f y z, t2)
+byteStringToObservation x =
```
bwbush, line 71: Another way of dealing with pattern-matching not working against integers in Plutus on-chain code is to use guards and `PlutusTx.Eq`, for example
```haskell
  | y PlutusTx.== 0 = . . .
  | y PlutusTx.== 1 = . . .
  . . .
```
The case statements here might produce smaller compiled code, however.

marlowe/src/Language/Marlowe/SemanticsTypes.hs
```diff
@@ -184,9 +186,13 @@ data Payee = Account AccountId
     datatype Case is mutually recurvive with @Contract@
 -}
 data Case a = Case Action a
+            | MerkleizedCase Action BuiltinByteString
```
bwbush, line 189: Sometimes the practice in other Cardano packages is to create a `newtype` and a `Hash` type-family member, but I don't know that Marlowe should conform to that.
```haskell
newtype SerialisedContinuation = SerialisedContinuation {unSerialisedContinuation :: BuiltinByteString}
data Case a = Case Action a | MerkleizedCase Action (Hash SerialisedContinuation)
data Input = NormalInput InputContent | MerkleizedInput InputContent SerialisedContinuation
```
In any case, it seems safer to use types to distinguish the continuation from its hash.

marlowe/src/Language/Marlowe/Util.hs
```diff
@@ -65,7 +65,9 @@ foldMapContract fcont fcase fobs fvalue contract =
         Assert obs cont      -> fobs' obs <> go cont
   where
     go = foldMapContract fcont fcase fobs fvalue
-    fcase' cs@(Case _ cont) = fcase cs <> go cont
+    fcase' cs = fcase cs <> case cs of
+        Case _ cont        -> go cont
+        MerkleizedCase _ _ -> mempty
```
bwbush, line 70: Should the semantics of `foldMapContract` make a special provision for handling merkleization? For instance, it could include an argument `fmerk :: SerialisedContinuation -> m` so that `MerkleizedCase _ _ -> mempty` would become `MerkleizedCase _ cont -> fmerk cont`. One could always recover the behavior here by setting `fmerk = const mempty`, but more complex code could use `fmerk` to lookup and deserialise the contract corresponding to the hash.

marlowe/test/Spec/Marlowe/Common.hs
```diff
@@ -260,12 +262,15 @@ shrinkAction action = case action of
 
 
 caseRelGenSized :: Int -> Integer -> Gen (Case Contract)
-caseRelGenSized s bn = Case <$> actionGenSized s <*> contractRelGenSized s bn
-
+caseRelGenSized s bn = frequency [ (9, Case <$> actionGenSized s <*> contractRelGenSized s bn)
```
bwbush, line 265: Would it be worthwhile adding a QuickCheck test to serialise and deserialise merkelized contracts, or do the existing test cases implicitly do this already?

marlowe/src/Language/Marlowe/Semantics.hs
```diff
@@ -416,39 +424,56 @@ reduceContractUntilQuiescent env state contract = let
...
-        if choId1 == choId2 && inBounds choice bounds
-        then let
-            newState = state { choices = Map.insert choId1 choice (choices state) }
-            in Applied ApplyNoWarning newState cont
-        else applyCases env state input rest
-    (INotify, Case (Notify obs) cont : _)
-        | evalObservation env state obs -> Applied ApplyNoWarning state cont
-    (_, _ : rest) -> applyCases env state input rest
-    (_, []) -> ApplyNoMatchError
+-- | Try to apply a single input contentent to a single action
```
bwbush, line 431: typo "contentent"

bwbush (commented):
I reviewed everything except FSSemantics.hs, which I don't understand yet. Please let me know if you'd like me to invest time in that, too.

My only concerns are ~~the semantics of `Eq (Case a)` and~~ `foldMapContract`, which are awkward because of merkleization.

marlowe/src/Language/Marlowe/SemanticsTypes.hs
bwbush: I guess my concern was that a merkleization of a contract and the original normal contract would logically embody the same real-world contract, but they wouldn't be `==`. So I guess we have to accept that equality here is about equality of representation.

palas: Yes, in that sense there already are many contracts that are equivalent and we treat them as different, like `If TrueObs Close Close` and `Close`, for example.

marlowe/src/Language/Marlowe/SemanticsDeserialisation.hs
palas: Yes, I doubt that will get optimised, but I hadn't thought of that possibility. We will be replacing this when Plutus implements serialisation anyway, so it is not too bad

marlowe/test/Spec/Marlowe/Common.hs
palas: Yes, there is a test for serialising and deserialising contracts that are generated by this generator (I think)

marlowe/src/Language/Marlowe/Semantics.hs
palas: Well spotted!

palas:
> I reviewed everything except FSSemantics.hs, which I don't understand yet. Please let me know if you'd like me to invest time in that, too.
> 
> My only concerns are ~the semantics of `Eq (Case a)` and~ `foldMapContract`, which are awkward because of merkleization.

Thanks for the reviews. I wouldn't worry about `FSSemantics` for now, it is pretty dense and I think that what I changed shouldn't make any difference to existing contracts that don't use `MerkleizedCase` (which are the ones that can be submitted from the playground).

marlowe/src/Language/Marlowe/Util.hs
palas: Yes, you are right, it can be quite problematic, I think I addressed it now. I am also a bit nervous about `contractLifespanUpperBound`, but we can think what to do with it later

marlowe/src/Language/Marlowe/SemanticsTypes.hs
palas: This makes a lot of sense, but I am going to leave it out for now since we no longer have to deal with both serialised continuations and hashes

marlowe/src/Language/Marlowe/Serialisation.hs
```diff
@@ -42,12 +42,18 @@ intToByteString x =
    then positiveIntToByteString (x * 2)
    else positiveIntToByteString (0 - (x * 2 + 1))
 
+{-# INLINABLE genericLength #-}
```
nau, line 45: There is `length` in Prelude.

nau:
I think I've come up with alternative solution to Marlowe ByteString serialization.

We can include unmerkleised `Contract` (one from `MerklizedInput`) in `Tx` `Datums`. Then, the node will hash it for us (and we trust it), and we get `[(DatumHash, Datum)] `
In `MerkelizedInput`  we pass `DatumHash` of continuation.
In validator with find corresponding `Datum`, `fromBuiltinData`  it into `Contract` and pass it.
This way we remove all the serialization (>1000 bytes of validator) and don't waste CPU units on serialization and hashing.

nau:
This way, `Input` can remain the same (we don't need `MerkleisedInput`), 
`Case` would become 
```
data Case a = Case Action (Either a ByteString)
```
and we extend `Env` with `continuations :: Map ByteString Contract`

And we will be able to do multiple inputs in the same transaction, btw.

PR #27, opened 2021-12-02
SCP-3088 - Marlowe Run - Add restoreWallet endpoint

hrajchert:
This PR creates a new endpoint in the Marlowe Run Server to restore a wallet from the Mnemonic phrase and adds the code to call it from the client side (Although the usage will be added in a future PR).

Pre-submit checklist:
- Branch
- PR
    - [X] Self-reviewed the diff
    - [X] Reviewer requested


marlowe-dashboard-client/webpack.config.js
```diff
@@ -18,19 +18,25 @@ module.exports = {
         https: true,
         stats: "errors-warnings",
         proxy: {
-            "/api": {
+            "/pab/api": {
```
hrajchert, line 21: @shlevy we need to modify the Traefik configuration accordingly, not sure where is that located.

shlevy: @hrajchert In the current state everything is served from the PAB, so traefik just sends everything there. We'll only have to worry about routes once we have the marlowe-run-server in the mix.

shlevy: Ah I misread the PR, it looks like that's what's happening here.

Will it make sense to deploy this just using all of the mock stuff from the PAB, or is it only something we can deploy with a proper WBE, node, etc.?

hrajchert: Yes, the idea of this PR is to bring marlowe-run-server into the mix. I'm not too worried about the deployment as there are other things that need to happen for this to properly work, but yes, we are going to need the WBE and node in the deployment relatively soon.

marlowe-dashboard-server/src/Marlowe/Run/Webserver/Server.hs
```diff
@@ -0,0 +1,206 @@
...
+            :<|> (handleVersion
+                   :<|> restoreWallet
+                 )
+            :<|> serveDirectoryFileServer staticPath
+        )
+    where
+    liftHandler :: ReaderT Env (ExceptT ServerError IO) a -> Handler a
+    liftHandler = Handler . flip runReaderT env
+
+-- TODO: Can we get this from cabal somehow?
```
hrajchert, line 75: @shlevy how would you solve this issue? Should we put this version on a ENV variable?

shlevy: `Paths_pkgname` includes a version export: https://cabal.readthedocs.io/en/3.6/cabal-package.html#accessing-data-files-from-package-code

bwbush: You can also use [`Cardano.Config.Git.Rev.gitRev`](https://github.com/input-output-hk/cardano-config/blob/master/src/Cardano/Config/Git/Rev.hs#L16) if you want the git revision.

marlowe-dashboard-server/src/Marlowe/Run/Webserver/Server.hs
```diff
@@ -0,0 +1,208 @@
...
+decodeError
+    :: BL.ByteString
+    -> Maybe String
+decodeError bytes = do
+    obj <- Aeson.decode bytes
+    Aeson.parseMaybe (Aeson.withObject "Error" (.: "message")) obj
+
+-- NOTE: Coppied from plutus-apps/playground-common/Schema.hs with same note
+-- We could take this from the `safe` package, but I don't think it's worth the extra dependency.
+safeHead :: [a] -> Maybe a
```
bwbush, line 188: This function is called `headMay` in `Cardano.Prelude`.

marlowe-dashboard-server/src/Marlowe/Run/Webserver/Server.hs
```diff
@@ -0,0 +1,208 @@
...
+safeHead []    = Nothing
+safeHead (x:_) = Just x
+
+hush :: Either a b -> Maybe b
+hush = either (const Nothing) Just
```
bwbush, line 193: `hash` = `Cardano.Prelude.rightToMaybe`

bwbush (approved):
The code looks great. I also successfully tested this with cURL:
```bash
curl --trace-ascii curl.log \
     -X POST \
     -H "Content-type: application/json" \
     -H "Accept: application/json" \
     http://127.0.0.1:8080/api/wallet/restore \
     -d @test.wallet
```

PR #34, opened 2021-12-10
SCP-3089 - Marlowe Run Frontend - Re implement welcome page with wallet restore

hrajchert:
This PR adds a modal to restore a wallet from the centralized server using the endpoint created in [PR 27](https://github.com/input-output-hk/marlowe-cardano/pull/27). We can now advance to the Main screen but we can't create a contract yet.

Pre-submit checklist:
- Branch
- PR
    - [X] Self-reviewed the diff
    - [X] Reviewer requested

### DIFF
--- a/marlowe-dashboard-client/default.nix
+++ b/marlowe-dashboard-client/default.nix
@@ -25,10 +25,9 @@ let
   start-backend = pkgs.writeShellScriptBin "marlowe-pab-server" ''
     echo "marlowe-pab-server: for development use only"
     (trap 'kill 0' SIGINT;
-      $(nix-build ../default.nix --quiet --no-build-output -A marlowe-dashboard.marlowe-invoker)/bin/marlowe-pab --config plutus-pab.yaml all-servers &
+      $(nix-build ../default.nix --quiet --no-build-output -A marlowe-dashboard.marlowe-invoker)/bin/marlowe-pab --config plutus-pab.yaml webserver &
       $(nix-build ../default.nix -A marlowe-dashboard.marlowe-run-backend-invoker)/bin/marlowe-dashboard-server webserver -c ./config.json
     )
-
   '';
 
   cleanSrc = gitignore-nix.gitignoreSource ./.;
--- a/marlowe-dashboard-client/default.nix
+++ b/marlowe-dashboard-client/default.nix
@@ -24,7 +24,11 @@ let
 
   start-backend = pkgs.writeShellScriptBin "marlowe-pab-server" ''
     echo "marlowe-pab-server: for development use only"
-    $(nix-build ../default.nix --quiet --no-build-output -A marlowe-dashboard.marlowe-invoker)/bin/marlowe-pab --config plutus-pab.yaml all-servers
+    (trap 'kill 0' SIGINT;
+      $(nix-build ../default.nix --quiet --no-build-output -A marlowe-dashboard.marlowe-invoker)/bin/marlowe-pab --config plutus-pab.yaml all-servers &
+      $(nix-build ../default.nix -A marlowe-dashboard.marlowe-run-backend-invoker)/bin/marlowe-dashboard-server webserver -c ./config.json
+    )
+
   '';
 
   cleanSrc = gitignore-nix.gitignoreSource ./.;
--- a/marlowe-dashboard-client/src/Capability/Marlowe.purs
+++ b/marlowe-dashboard-client/src/Capability/Marlowe.purs
@@ -20,6 +20,7 @@ module Capability.Marlowe
 
 import Prologue
 import API.Lenses (_cicContract, _cicCurrentState, _cicDefinition, _cicWallet, _observableState)
+import API.Marlowe.Run.TestnetWallet (RestoreError(..))
 import API.Marlowe.Run.TestnetWallet (RestoreWalletOptions)
 import AppM (AppM)
 import Bridge (toBack, toFront)
@@ -33,7 +34,7 @@ import Component.Contacts.Lenses (_companionAppId, _marloweAppId, _pubKeyHash, _
 import Component.Contacts.Types (WalletDetails, WalletId(..), WalletInfo)
 import Control.Monad.Except (ExceptT(..), except, lift, runExceptT, withExceptT)
 import Control.Monad.Reader (asks)
-import Data.Argonaut.Decode (JsonDecodeError)
+import Data.Argonaut.Decode (JsonDecodeError, (.!=))
 import Data.Argonaut.Extra (parseDecodeJson)
 import Data.Array (filter) as Array
 import Data.Array (find)
@@ -46,7 +47,6 @@ import Data.Tuple.Nested ((/\))
 import Halogen (HalogenM, liftAff)
 import Marlowe.Client (ContractHistory)
 import Marlowe.PAB (PlutusAppId)
-import API.Marlowe.Run.TestnetWallet (RestoreError(..))
 import Marlowe.Semantics (Contract, MarloweData, MarloweParams, PubKeyHash, TokenName, TransactionInput)
 import MarloweContract (MarloweContract(..))
 import Plutus.PAB.Webserver.Types (CombinedWSStreamToServer(..), ContractInstanceClientState)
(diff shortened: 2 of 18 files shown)


marlowe-dashboard-client/default.nix
```diff
@@ -24,7 +24,10 @@ let
 
   start-backend = pkgs.writeShellScriptBin "marlowe-pab-server" ''
     echo "marlowe-pab-server: for development use only"
-    $(nix-build ../default.nix --quiet --no-build-output -A marlowe-dashboard.marlowe-invoker)/bin/marlowe-pab --config plutus-pab.yaml all-servers
+    (trap 'kill 0' SIGINT;
+      $(nix-build ../default.nix --quiet --no-build-output -A marlowe-dashboard.marlowe-invoker)/bin/marlowe-pab --config plutus-pab.yaml webserver &
+      $(nix-build ../default.nix -A marlowe-dashboard.marlowe-run-backend-invoker)/bin/marlowe-dashboard-server webserver -c ./config.json
+    )
```
hrajchert, line 30: Hi @shlevy, I made this as a placeholder to be able to run the PAB + the Marlowe Run BE at the same time, but is not very robust, if one of the process dies you can't kill the other one.

How should we launch these on development? Take in mind that we'll also need WBE and chainindex

shlevy: Will take a look next week.

marlowe-dashboard-client/src/Capability/Marlowe.purs
```diff
@@ -102,6 +107,32 @@ instance manageMarloweAppM :: ManageMarlowe AppM where
             , previousCompanionAppState: Nothing
             }
         pure $ createWalletDetails <$> ajaxCompanionAppId <*> ajaxMarloweAppId <*> ajaxAssets
+  restoreWallet options = do
+    mWalletInfo <- Wallet.restoreWallet options
+    case mWalletInfo of
+      Left err -> pure $ Left err
+      Right walletInfo -> do
+        let
+          walletId = view _walletId walletInfo
```
paluh, line 116: I'm not sure if I understand the context and the comment:

`Variant` based error handling allow us to compose errors from different independent parts of the application into a single type (no need for nesting errors).
Adding more more detailed information / semantics to errors is a bit orthogonal problem I think and can be done without _checked-exceptions_.

marlowe-dashboard-client/src/Capability/Marlowe.purs
```diff
@@ -102,6 +107,32 @@ instance manageMarloweAppM :: ManageMarlowe AppM where
...
+        -- create the WalletCompanion and MarloweApp for this wallet
+        -- TODO: We had initial discussions for adding Variants for errors probably using Checked Exceptions.
+        -- Instead of the general ClientServerError we could encapsulate the type of error inside a context
+        -- (restoring the wallet, activating a wallet companion or marlowe app).
+        -- For the end users this wouldn't change too much their UX, but for production code it can be useful
+        -- to diagnose problems using a tool like sentry SCP-3171.
+        -- https://github.com/natefaubion/purescript-checked-exceptions
+        ajaxCompanionAppId <- lmap ClientServerError <$> Contract.activateContract WalletCompanion walletId
+        ajaxMarloweAppId <- lmap ClientServerError <$> Contract.activateContract MarloweApp walletId
+        let
```
paluh, line 126: Cosmetics - do we avoid `puns` here for clarity?
 ```
createWalletDetails =
  { walletNickname: ""
  , companionIAppId: _
  , marloweAppId: _
  ...
  }
```

File: marlowe-dashboard-client/src/Page/Welcome/State.purs [hunk 4]
```diff
@@ -93,74 +117,43 @@ that wallet: a `WalletCompanion` and a `MarloweApp`.
...
-      walletLibrary <- use _walletLibrary
-      for_ (lookup walletNicknameOrId walletLibrary) (handleAction <<< OpenUseWalletCardWithDetails)
-    _ -> pure unit
+handleAction RestoreTestnetWallet = do
+  walletName <- use (_walletNicknameInput <<< _value)
+  mnemonicPhraseStr <- use (_walletMnemonicInput <<< _value)
+  let
+    mnemonicPhrase = words mnemonicPhraseStr
+  result <- restoreWallet { walletName, mnemonicPhrase, passphrase: "" }
+  case result of
```
paluh, line 142: Cosmetics: maybe some parts like `Milliseconds 1200.0` or even the whole `tell _submitButtonSlot "restore-wallet" <<< SubmitResult (Millisecond 1200.0)` could be extracted into a meaning full variable above?

marlowe-dashboard-client/src/Page/Welcome/State.purs
```diff
@@ -49,13 +52,25 @@ mkInitialState walletLibrary =
...
   }
 
+words :: String -> Array String
+words = Array.filter (notEq "") <<< split (Pattern " ")
+
+walletMnemonicError :: Maybe String -> String -> Maybe WalletMnemonicError
+walletMnemonicError invalidFromServer phrase =
+  if Just phrase == invalidFromServer then
+    Just InvalidMnemonicFromServer
+  else if 24 /= (Array.length $ words phrase) then
```
paluh, line 69: Maybe `MnemonicAmountOfWords` could be extended with like `MnemonicAmountOfWords { expected :: Int, got :: Int }` info so info for the user can be more meaningful.

marlowe-dashboard-client/src/Page/Welcome/View.purs
```diff
@@ -72,71 +72,57 @@ welcomeCard state =
...
+        [ classNames [ "pt-2", "flex", "justify-between" ] ]
+        [ a
+            [ classNames [ "flex", "font-bold" ]
+            , href "https://staging.marlowe-web.iohkdev.io"
+            ]
+            [ icon_ Icon.Previous
+            , text "Back to home page"
+            ]
+        , a
+            [ classNames [ "font-bold" ]
```
paluh, line 119: This `FIXME` is intended or should be fixed in this PR?

paluh (approved):

marlowe-dashboard-client/src/Capability/Marlowe.purs
hrajchert: Well, handling errors is tricky. 
For an end user, if we cannot restore a wallet, of the 3 possible error options that I modelled, only one gives the user something meaningful to do:

```purescript
data RestoreError
  = InvalidMnemonic
  | ClientServerError AjaxError
  | ServerError BE.RestoreError
```
If the user provides an invalid mnemonic phrase because he/she made out words, then we provide a message saying that the phrase is invalid, and the user can try again with a new phrase.

The other two is either a Client to Server ajax error that we are not catching (any non 2xx message), or the server failing somehow (maybe the wallet backend is not available). 

In any case, it's probably not worth it to provide a message more elaborated than "There was a problem, try again later or contact support". 

If that is the approach we want to take for Client/Server errors, then this could be a Variant of 3 individual errors. Invalid mnemonic would only apply to restore wallet, and the other two could be re-utilized for other endpoints (for example the getTotalFunds that I'm working on). 

Showing a message to a user is the bare minimum we can do to avoid showing a loading icon forever, but it doesn't help us diagnose or fix the error. That is where observability comes into play, and tools like getsentry are super useful. But we need to add/aggregate enough information to be able to work offline. Some of the context is manually added by the tools (browser information, list of pages visited) and we could add our own information, for example a list of Actions that were executed.
In this case in particular, we are treating Client/Server communication problems the same, and we are not even saying which step did it fail, was it on the restoreWallet endpoint? Was it when trying to activate the wallet companion? when activating the MarloweApp?

We could probably create a variant that is ClientServer communication, in which we allow to add a name of step, arbitrary JSON context and the Ajax error itself.

In general, I was imagining on making the `handleAction` to work with a `MonadError` of a Variant in which all members need to implement a `ToSentry` typeclass (not sure if it's possible to model that), and maybe a `ToToast` typeclass. That way, we can create some variants like InvalidMnemonic that needs to be cached, and for the rest, we can provide a general handler that sends the error to sentry and puts a toast message

marlowe-dashboard-client/src/Capability/Marlowe.purs
hrajchert: not sure what you mean

marlowe-dashboard-client/src/Page/Welcome/State.purs
hrajchert: yeah, more than variable, a function with clearer meaning. Something like `informLoadingSubmitError` and `informLoadingSubmitSuccess`

PR #36, opened 2021-12-15
SCP-3148 Add support for reordering parameter descriptions

paluh:
This commit changes a bit more than it should as it provides an example of `Variant` based actions + handling some generic actions over a state record (more details in code comments).

Few questions:

* I've decided to put code which we can possibly separate as libs etc. into the `Contrib` subpackage like `Contrib.Array.Builder` . Do I break some convention here?

* I'm not sure what is the policy regarding css generalization - for example there are repeated stylings for different buttons in the codebase which turn them into "disabled" buttons (color change), there is `small-btn` which is not used in `panels.css` instead we use separate button types which set a bit smaller size etc. I've tried to follow the convention from `panels.css`

* My final styling of the buttons is not really nice - should I care about this more at this stage?
2021-12-15-112157_1920x1080_scrot


paluh:
The `Contrib.Array.Builder](https://github.com/input-output-hk/marlowe-cardano/pull/36/files#diff-fee8411371d890a0f3e1ccf2c7ccf1242b1725fa370c6923dcadfc690d54025c) implementation was proposed by me in a [separate context](https://github.com/purescript/purescript-arrays/issues/115). I should publish this as a small, independent lib but I'm including it for now here so we can use it - usability is nearly the same as `Array` because we use `<>`, but the performance is better and sometimes code formatting can be less cumbersome (or more it depends ;-)).

hrajchert:
I'll review in more detail but in the meantime:

1. We don't have a convention for this, the packages are normally called what they would be called if they were independent and we use comments to indicate if a package could be promoted to a library at some point. I don't really have an opinion of this subject, you can create a github discussion and we can post it  #frontend-dev.
2. We are using [tailwind.css](https://tailwindcss.com/) for the most part, for example in Marlowe Run or the Marlowe Website. But this project was created before that decision and it mixed general classes (`.simulator-something`) with made up utility classes (`a-horizontally`). A little [more context](https://github.com/input-output-hk/plutus/pull/2551#discussion_r548189966) on why and when this was added.
3. All this should be redesigned, so I wouldn't spend too much time on this.

marlowe-playground-client/src/Component/MetadataTab/View.purs
```diff
@@ -209,38 +227,57 @@ metadataList metadataAction metadataMap hintSet metadataRenderer typeNameTitle t
...
-                              , onClick $ const $ metadataAction (setEmptyMetadata key)
-                              ]
-                              [ text "+" ]
-                          ]
-                      ]
-                )
-            )
-            $ OMap.toUnfoldable combinedMap
-        )
+    AB.build
```
jhbertra, line 248: `ArrayBuilder` - nice, I'm guesing this is like a difference list that defers all updates until the end then "fuses" them together mutably in JS? That's really handy.

marlowe-playground-client/src/Contrib/Data/Array/Builder.js
```diff
@@ -0,0 +1,27 @@
...
+    return arr;
+  }
+}
+exports.unsafeSnoc = function(a) {
+  return function(arr) {
+    arr.push(a);
+    return arr;
+  };
+};
+
```
jhbertra, line 27: This is great, I think we should open-source this.

marlowe-playground-client/src/Contrib/Data/Array/Builder.purs
```diff
@@ -0,0 +1,40 @@
...
+newtype Builder a
+  = Builder (Array a → Array a)
+
+instance semigroupBuilder ∷ Semigroup (Builder a) where
+  append (Builder b1) (Builder b2) = Builder (b1 <<< b2)
+
+instance monoidBuilder ∷ Monoid (Builder a) where
+  mempty = Builder identity
+
+foreign import unsafeAppend ∷ ∀ a. Array a → Array a → Array a
```
jhbertra, line 19: Just an FYI that we will be enforcing ascii syntax with the switch to `purs-tidy` (although I could be convinced to make it do unicode instead - it is more concise for sure)

marlowe-playground-client/src/Contrib/Halogen/State/Record.purs
```diff
@@ -0,0 +1,74 @@
...
+-- | accepts a `Variant` value (an "action") and dispatches it
+-- | using its label in a type safe manner over our state value.
+data SettersDispatchStep
+  = SettersDispatchStep
+
+-- | We use `FoldingWithIndex` here and not
+-- | `MappingWithIndex` because mapping requires
+-- | a value to operate on and folding allows us to build up
+-- | a value just from the `Proxy (RowList k)`.
+instance purtyProblem0 ::
```
jhbertra, line 26: what is this instance name abount?

marlowe-playground-client/src/Contrib/Halogen/State/Record.purs
```diff
@@ -0,0 +1,74 @@
...
+  ( IsSymbol l
+  , Symbol.Transform.ToUpperFirst l l'
+  , Symbol.Append "set" l' l''
+  , IsSymbol l''
```
jhbertra, line 30: This is awesome

marlowe-playground-client/src/Contrib/Halogen/State/Record.purs
```diff
@@ -0,0 +1,74 @@
+module Contrib.Halogen.State.Record where
```
jhbertra, line 1: As mentioned before, I think that it would be good in the future to generalize this to a set of tools for creating field lenses and prisms from records and variants. Also, I think it doesn't make a lot of sense for this to be namespaced with `Halogen`, as you say below, it isn't really related to halogen... I'd call it a "peer-dependency" :)

marlowe-playground-client/src/Contrib/Type/Proxy/Generic.purs
```diff
@@ -0,0 +1,40 @@
+module Contrib.Type.Proxy.Generic where
```
jhbertra, line 1: The current convention in our codebase is something like

```
module Type.Proxy.Extra
```

Though again, I could be convinced to switch to `Contrib.`

jhbertra (approved):
Lots of interesting stuff in here. I've got no major problems with anything here but I do think that some of the work here should be extracted as "library" code

marlowe-playground-client/src/Component/MetadataTab/Types.purs
```diff
@@ -1,40 +1,111 @@
...
+-- | handle them in our final state handler together with generic handlers.
+-- |
+-- |
+-- | `Variant` usage notes:
+-- |
+-- | * We close the row here and wrap the type in a `newtype` so actually we
+-- | are building a closed sum type here. We do this because there is no
+-- | obvious benefit of keeping the row polymorphic. In theory if we would have
+-- | multiple components with shared subset of actions then possibly we could
+-- | reuse some templates / handlers etc. This scenario seems a bit unrealistic.
```
hrajchert, line 43: Well, actually in MarloweRun I've seen some possible uses of this scenario. At some point we have two sets of data in the same place, contract step information and tabs UI information. They are together because we wanted to persist tabs state for different steps, but that made the logic to be tied to a particular type instead of being able to say something that has tab information and something else.

marlowe-playground-client/src/Component/MetadataTab/State.purs
```diff
@@ -1,49 +1,69 @@
...
+  -- | Let's cache generic set actions handlers
+  let
+    genericDispatcher =
+      Record.merge
+        (mkSettersDispatcher (Proxy :: Proxy MetaData))
+        (mkUpdatersDispatcher (Proxy :: Proxy MetaData))
+  \(MetadataAction action) -> do
+    let
+      go =
+        ( (Variant.match genericDispatcher <<< Variant.expand)
```
hrajchert, line 38: I'm not really sure what is going on here... In the initial code that I saw, I thought all of the handlers were generic, in here I thiiink some are being manually dispatched and some are being genericly dispatched? Some comments to clarify this would be useful. Also, the fact that is not immediately obvious is something to consider 🤔

marlowe-playground-client/src/Contrib/Type/Data/Symbol/Transform.purs
```diff
@@ -0,0 +1,61 @@
...
+
+instance purtyProblemX :: ToUpper "x" "X"
+
+instance purtyProblemY :: ToUpper "y" "Y"
+
+instance purtyProblemZ :: ToUpper "z" "Z"
+
+class ToUpperFirst (s :: Symbol) (s' :: Symbol) | s -> s'
+
+instance purtyProblem :: (Symbol.Cons h t s, ToUpper h h', Symbol.Cons h' t s') => ToUpperFirst s s'
```
hrajchert, line 61: `Capitalize`?

marlowe-playground-client/src/Halogen/Classes.purs
```diff
@@ -169,6 +169,12 @@ plusBtn = ClassName "plus-btn"
 minusBtn :: ClassName
 minusBtn = ClassName "minus-btn"
 
+upBtn :: ClassName
```
hrajchert, line 172: we are trying to get away from this module and just use tailwind classes as string.

marlowe-playground-client/src/Component/MetadataTab/View.purs
paluh: Yeah. It closely follows `Record.Builder` internals.

marlowe-playground-client/src/Contrib/Data/Array/Builder.purs
paluh: This is an accident. I thought that purty is going to drop this. This is going to be removed by `purs-tidy` for sure. Should I fix it before merge?

marlowe-playground-client/src/Contrib/Halogen/State/Record.purs
paluh: I'm leaving instance names in the code because `purty` enforces it. We can drop this stuff naming instances boilerplate when we move to `purs-tidy` because instance names are not required any more.

marlowe-playground-client/src/Contrib/Type/Data/Symbol/Transform.purs
paluh: I wasn't sure... css `capitlize` transforms every word in a sentence. On the other hand we should rather not expect `Symbol` with... white characters in it. I can rename it as a part of next refactoring PR.

marlowe-playground-client/src/Component/MetadataTab/State.purs
paluh: The motivation behind this complexity here is motivated by the large comment in the `MetadataTab/Types.purs` above - I've tried to provide "an example" of composition. I'm not sure if I should leave it as it is or simplify this code altogether (into a fully generic handler) and make example composition of handlers a part of the docs.

marlowe-playground-client/src/Component/MetadataTab/Types.purs
paluh: I would like to discuss this in more details. It would be nice to check if `Variant` is going simplify the code in this other case or make it more complex :-P

marlowe-playground-client/src/Contrib/Type/Proxy/Generic.purs
paluh: I've just missed `Extra` - I can turn this into `Extra` to keep consistency in the codebase.

It is probably a matter of a habit but maybe it is a bit easier to organize namespaces (split extras) when `Contrib` namespace is on the top.

I don't have a really strong preference in this regards.

paluh: I'm leaving `Contrib` namespace for now - we can discuss pros and cons in the near future.

marlowe-playground-client/src/Contrib/Halogen/State/Record.purs
paluh: :-)

marlowe-playground-client/src/Contrib/Halogen/State/Record.purs
paluh: Yeah, I've got some working prototype here:

https://try.purescript.org/?gist=b7b3e768886639004c355250b7a6ea07

I want to provide a nicer way to specify signatures because signatures generated by compiler are awful (the `Lens` related alises are expanded etc.).

marlowe-playground-client/src/Halogen/Classes.purs
paluh: I'm leaving this for now as it is - ok?

paluh:
I'm closing this in favor of https://github.com/input-output-hk/marlowe-cardano/pull/64 . I've moved `Array.Builder` there but I didn't introduced generic `Variant` parts there.

@jhbertra, @hrajchert ,@palas Some of these generic utils are possibly useful and can reduce boilerplate in simple components - I can try to reintroduce them at some point but maybe in some simpler context when they would be clearly the boilerplate cutters. Does it sounds ok?

PR #38, opened 2021-12-20
Drop unnecessary Proxy from Marlowe.Hole constructor

paluh:
The `Proxy a` in `Hole a` carries no information as `a` type is already in the `Hole` type itself. I've simplified the implementation and provided the `Proxy a` in a place where it is required based on the `a` parameter from the `Hole a` type.

This work was done alongside SCP-3164 which works with `Marlowe.Holes`.


hrajchert:
Code looks good and the Marlowe editor provides the correct fillers.

LGTM 👍🏻

hrajchert (approved):

PR #39, opened 2021-12-23
SCP-2866 - Add halogen store to Marlowe Run

hrajchert:
This PR is part of the Marlowe Run technical debt epic (SCP-2830) and it's a requirement for removing the sub-component architecture and use regular Halogen components. The main reasons for doing that change are:

- Sub-components are not standard practice, so they are not maintained
- They are very susceptible to performance issues (every time you update the state of a component the whole page is re-rendered)
- They add a lot of boilerplate and manual passing of data
- It's harder to separate logic concerns (there is functionality that needs to be split between the Contract page, Dashboard and sometimes MainFrame).
- It complicates the lifecycle events.

Having said that, it is worth noting that at the time that sub-components were introduced there wasn't an established way of having a shared state. Now Halogen State has taken that role in the community and this is the first step in that direction.

The PR doesn't do much itself, it adds the library, changes the ReaderT implementation of the `Env`ironment and sets the first mutable state `currentSlot`. That variable is not being read by the different components and we still have to manually pass it through the view and state until we can replace the sub-components.

Pre-submit checklist:
- Branch
- PR
    - [X] Self-reviewed the diff
    - [X] Reviewer requested


paluh (approved):
I'm not sure why `StoreT` provides `MonadAsk` implementation (in the newest version it provides this instance [directly](https://github.com/thomashoneyman/purescript-halogen-store/blob/v0.3.0/src/Halogen/Store/Monad.purs#L59)) because it in some sens violates immutability of the `ReaderT`.... (for example `StateT` is not a provider of `MonadAsk` but relies on the reader from the transformers stack).
Because it is a part of the official API for this type I won't object to use it here - I also understand that keeping `Env e` polymorphic should in some sens protect us against such misunderstandings ;-)

marlowe-dashboard-client/src/AppM.purs
```diff
@@ -1,19 +1,24 @@
...
-import Env (Env)
-import Type.Equality (class TypeEquals, from)
+import Halogen (Component)
+import Halogen.Store.Monad (class MonadStore, StoreT, getStore, runStoreT)
+import Safe.Coerce (coerce)
+import Store as Store
 
 newtype AppM a
-  = AppM (ReaderT Env Aff a)
+  = AppM (StoreT Store.Action Store.Store Aff a)
```
jhbertra, line 18: I agree with @paluh, I don't think it is a good idea to merge `Env` and `Store`. `Store` should be global mutable state, and `Env` should be global readonly state. I'd prefer it if this was 

```suggestion
  = AppM (ReaderT Env (StoreT Store.Action Store.Store Aff) a)
```

jhbertra (changes requested):

marlowe-dashboard-client/src/AppM.purs
jhbertra: Ah, I see there aren't instances for `MonadStore` `ReaderT`. Hence, `ReaderT Env (StoreT S m) a` won't have a `MonadStore` instance... What the heck is the matter with this library?! This is not at all the way monad transformers are supposed to work... 😞

jhbertra (approved):

PR #41, opened 2021-12-27
SCP-3118: Add real node, WBE, and chain-index to deployment and shell

shlevy:


shlevy:
Note that the last 3 commits create `launch-node`, `launch-chain-index`, and `launch-wbe` scripts to launch instances locally that point to our private testnet.

marlowe-dashboard-client/default.nix
```diff
@@ -22,11 +22,18 @@ let
...
+    cat > marlowe-run.json <<EOF
+    {
+      "getWbeConfig": { "_wbeHost": "localhost", "_wbePort": $NOMAD_PORT_wbe },
+      "getStaticPath": "/var/empty"
+    }
+    EOF
     (trap 'kill 0' SIGINT;
       $(nix-build ../default.nix --quiet --no-build-output -A marlowe-dashboard.marlowe-invoker)/bin/marlowe-pab --config plutus-pab.yaml webserver &
-      $(nix-build ../default.nix -A marlowe-dashboard.marlowe-run-backend-invoker)/bin/marlowe-dashboard-server webserver -c ./config.json
+      $(nix-build ../default.nix -A marlowe-dashboard.marlowe-run-backend-invoker)/bin/marlowe-dashboard-server webserver -c ./marlowe-run.json
```
hrajchert, line 36: I added this a couple of weeks ago, now I updated it with the changes you made. Is there a better way to run multiple programs at once in bash/nix? Should we do something similar to launch the WBE, node and chainlink?

hrajchert (approved):
LGTM. I tested the centralized server deploying to the `deploy-hernan` branch and I tested it locally as well 👍🏻

marlowe-dashboard-client/default.nix
shlevy: @hrajchert I don't think there's a great answer to the problem of local test services currently, but one thing I like about the scripts I wrote in shell.nix is that they reuse all of the code that's used in the real deployment. So to start the backend here you'd just have to export the port env var, not create the config file yourself or do the nix-build yourself etc.

What I'd do is create a bunch of individual scripts like I did in the shell, then create an umbrella script that starts each in the background then waits for any to fail (using the `wait` builtin) and kills them all if they do.

bitte/pab.nix
```diff
@@ -79,31 +70,33 @@ let
...
+      -e "s|@NOMAD_ADDR_index@|$NOMAD_ADDR_index|g" \
+      ${pabYamlIn} > pab.yaml
 
+  wait-for-socket "$NOMAD_ALLOC_DIR/node.sock"
 
   ${pab-init-cmd}/bin/pab-init-cmd
 
   # Ugly ugly hack to kill the PAB at midnight UTC
-  ${pabExe} --config=${pabYaml} all-servers&
+  ${pabExe} --config=pab.yaml webserver &
```
bwbush, line 99: Is midnight UTC the best time for this? I'm wondering what the UTC time is when this would least impact users?

flake.lock
```diff
@@ -280,11 +280,11 @@
...
       "flake": false,
       "locked": {
-        "lastModified": 1636122782,
-        "narHash": "sha256-+T9TGzHEzyfixBysxLwy5VWVrL5xqKF5pcbRlHQr+wI=",
+        "lastModified": 1640364395,
+        "narHash": "sha256-q+neKzIYSzccmYl/u+hj6rsuKWTEEVGu8+mIqw8Rf7M=",
         "owner": "input-output-hk",
         "repo": "plutus-apps",
-        "rev": "404af7ac3e27ebcb218c05f79d9a70ca966407c9",
+        "rev": "57d10773c793fbdb7dc623c5923ef0c07a70c957",
```
bwbush, line 287: This version of plutus apps is much more recent than what @hrajchert is currently using.

shlevy: @bwbush It's only two commits, both touching deployment code: https://github.com/input-output-hk/plutus-apps/compare/404af7ac3e27ebcb218c05f79d9a70ca966407c9...57d10773c793fbdb7dc623c5923ef0c07a70c957

bitte/pab.nix
shlevy: This is what was decided for the summit earlier this year. I don't know currently how our users our distributed... Ideally we'd just fix the bug causing memory leak over time though.

PR #37, opened 2021-12-16
SCP-3132 Get total funds through WBE (part 1 - backend API)

hrajchert:
This PR refactors the current backend to modularize endpoints.
It also adds a new endpoint for get the total funds from the WBE.

Note that the `restoreWallet` is placed under a `CentralizedTestnet` module, indicating that this endpoint is only useful in that architecture, but the getTotalFunds is placed in plain Wallet, as its going to be useful for this architecture and also the full node one.

The previous `restoreWallet` endpoint is now under
```
POST /api/wallet/centralized-testnet/restore
  {  "getMnemonicPhrase" : ["word1", "word2", "...", "word24"]
  ,  "getPassphrase": "something"
  , "getWalletName": "my-wallet"
  }
```
The new `getTotalFunds is under

```
GET /api/wallet/<wallet-id>/get-total-funds
```

Pre-submit checklist:
- Branch
    - [ ] Tests are provided (if possible)
    - [X] Commit sequence broadly makes sense
    - [X] Key commits have useful messages
    - [X] Relevant tickets are mentioned in commit messages
    - [X] Formatting, materialized Nix files, PNG optimization, etc. are updated
- PR
    - [X] Self-reviewed the diff
    - [X] Useful pull request description
    - [X] Reviewer requested

PR #42, opened 2021-12-27
SCP-3195 - Re implement get total funds through WBE (part 2 frontend)

hrajchert:
This is part 2 of [this PR](https://github.com/input-output-hk/marlowe-cardano/pull/37). It gets both the ada value and the custom tokens (including role tokens) from the WalletBackEnd.

There is an important warning under the `Note [polling updateTotalFunds]` which may cause some bugs while the wallet is syncing. I created two issues to solve that SCP-3211 (add global polling to getTotalFunds) and SCP-3212 (Show wallet sync status in the UI) 

Pre-submit checklist:
- Branch
- PR
    - [X] Self-reviewed the diff
    - [X] Reviewer requested

### DIFF
--- a/marlowe-dashboard-client/src/Capability/Marlowe.purs
+++ b/marlowe-dashboard-client/src/Capability/Marlowe.purs
@@ -89,24 +89,16 @@ instance manageMarloweAppM :: ManageMarlowe AppM where
         -- create the WalletCompanion and MarloweApp for this wallet
         ajaxCompanionAppId <- Contract.activateContract WalletCompanion walletId
         ajaxMarloweAppId <- Contract.activateContract MarloweApp walletId
-        -- get the wallet's current funds
-        -- Note that, because it can take a moment for the initial demo funds to be added, at
-        -- this point the funds might be zero. It doesn't matter though - if we connect this
-        -- wallet, we'll get a WebSocket notification when the funds are added (and if we don't
-        -- connect it, we don't need to know what they are.)
-        -- TODO(?): Because of that, we could potentially forget about this call and just set
-        -- assets to `mempty`.
-        ajaxAssets <- Wallet.getWalletTotalFunds walletId
         let
-          createWalletDetails companionAppId marloweAppId assets =
+          createWalletDetails companionAppId marloweAppId =
             { walletNickname: ""
             , companionAppId
             , marloweAppId
             , walletInfo
-            , assets
+            , assets: mempty
             , previousCompanionAppState: Nothing
             }
-        pure $ createWalletDetails <$> ajaxCompanionAppId <*> ajaxMarloweAppId <*> ajaxAssets
+        pure $ createWalletDetails <$> ajaxCompanionAppId <*> ajaxMarloweAppId
   restoreWallet options = do
     mWalletInfo <- Wallet.restoreWallet options
     case mWalletInfo of
--- a/marlowe-dashboard-client/src/Capability/Marlowe.purs
+++ b/marlowe-dashboard-client/src/Capability/Marlowe.purs
@@ -205,7 +205,6 @@ instance manageMarloweAppM :: ManageMarlowe AppM where
             wallet = toFront $ view _cicWallet clientState
           walletContracts <- withExceptT Just $ ExceptT $ Contract.getWalletContractInstances wallet
           walletInfo <- withExceptT Just $ ExceptT $ Wallet.getWalletInfo wallet
-          assets <- withExceptT Just $ ExceptT $ Wallet.getWalletTotalFunds wallet
           case find (\state -> view _cicDefinition state == MarloweApp) walletContracts of
             Just marloweApp ->
               pure
@@ -213,7 +212,7 @@ instance manageMarloweAppM :: ManageMarlowe AppM where
                 , companionAppId
                 , marloweAppId: toFront $ view _cicContract marloweApp
                 , walletInfo
-                , assets
+                , assets: mempty
                 , previousCompanionAppState: Nothing
                 }
             Nothing -> except $ Left Nothing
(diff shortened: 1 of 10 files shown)


marlowe-dashboard-client/src/Page/Dashboard/State.purs
```diff
@@ -384,6 +395,29 @@ handleAction input@{ currentSlot, tzOffset } (ContractAction followerAppId contr
...
+--       Refactor this code and do a global polling mechanism.
+updateTotalFunds ::
+  forall m slots msg.
+  MonadAff m =>
+  ManageWallet m =>
+  HalogenM State Action slots msg m Unit
+updateTotalFunds = do
+  walletId <- use (_walletDetails <<< _walletInfo <<< _walletId)
+  response <- getWalletTotalFunds walletId
+  for_ response \(GetTotalFunds { assets }) ->
```
paluh, line 417: We can shorten this to [`Lens.Setter.assign`](https://github.com/purescript-contrib/purescript-profunctor-lenses/blob/v7.0.1/src/Data/Lens/Setter.purs#L96) like above or even just [`.=`](https://github.com/purescript-contrib/purescript-profunctor-lenses/blob/v7.0.1/src/Data/Lens/Setter.purs#L45)

marlowe-dashboard-client/src/Capability/Marlowe.purs
```diff
@@ -89,24 +89,16 @@ instance manageMarloweAppM :: ManageMarlowe AppM where
         -- create the WalletCompanion and MarloweApp for this wallet
         ajaxCompanionAppId <- Contract.activateContract WalletCompanion walletId
         ajaxMarloweAppId <- Contract.activateContract MarloweApp walletId
```
paluh, line 91: Understanding check: Do we drop this because we don't want to event checkt total funds against freshly created wallet?

marlowe-dashboard-client/src/Capability/Marlowe.purs
```diff
@@ -213,15 +205,14 @@ instance manageMarloweAppM :: ManageMarlowe AppM where
             wallet = toFront $ view _cicWallet clientState
           walletContracts <- withExceptT Just $ ExceptT $ Contract.getWalletContractInstances wallet
           walletInfo <- withExceptT Just $ ExceptT $ Wallet.getWalletInfo wallet
```
paluh, line 207: Understanding check: Do we drop this because we want to introduce more general solution for checks (per specific view sync or maybe based on some "app routing" logic etc.)?

marlowe-dashboard-client/src/Component/Contacts/State.purs
```diff
@@ -172,6 +167,24 @@ handleAction (ClipboardAction clipboardAction) = do
   mapAction ClipboardAction $ Clipboard.handleAction clipboardAction
   addToast $ successToast "Copied to clipboard"
```
paluh, line 169: Understanding check: This is just a plain refactoring - replacing an action which was exclusively used by other handler.

marlowe-dashboard-client/src/Component/Contacts/State.purs
```diff
@@ -188,11 +201,24 @@ toWalletIdInput ::
 toWalletIdInput = mapSubmodule _walletIdInput WalletIdInputAction
 
 ------------------------------------------------------------
+-- The cardano Blockchain has Multi-Asset support, each monetary policy is identified
+-- by a policyId which is also known as the CurrencySymbol. The ADA token is a special
+-- case that has the empty string as both the token name and the currency symbol
+-- https://docs.cardano.org/native-tokens/learn
+-- https://github.com/input-output-hk/plutus/blob/1f31e640e8a258185db01fa899da63f9018c0e85/plutus-ledger-api/src/Plutus/V1/Ledger/Ada.hs#L45
```
paluh, line 208: There is already: https://github.com/input-output-hk/purescript-web-common/tree/main/src/Plutus/V1/Ledger

paluh (commented):
I've included "self-check" comments with prefix "Understanding check" - these are only checks if I get the idea behind the code.

marlowe-dashboard-client/src/Capability/Marlowe.purs
hrajchert: As you mentioned yesterday, for a real wallet (testnet or even mainnet), it wouldn't make sense

marlowe-dashboard-client/src/Capability/Marlowe.purs
hrajchert: Yeah, as I mentioned in the general description, as part of SCP-3211 I want to have a more general way of getting this information. Probably setting it to the global store and having a heuristic for doing a global poll (either a fixed interval of X seconds or every time there is Y type of event, or similar). Also, we could later think if it makes sense for us to use the chainlink to listen for changes to a particular address.

marlowe-dashboard-client/src/Component/Contacts/State.purs
hrajchert: Yes, it didn't make sense to have it as an Action if it was only used internally

marlowe-dashboard-client/src/Component/Contacts/State.purs
hrajchert: Cool, we can put this next to that

marlowe-dashboard-client/src/Page/Dashboard/State.purs
hrajchert: Yeah, the only caveat I see is that if we want to modify 2 properties we need to change the "structure" of how we set things. I mean for a single property use assign, for multiple using modify_ + set, why not just use the latter.

paluh: I'm not sure I get it... you can always:

```purescript
do
  _lens1 := x
  _lens2 := y
```

hrajchert: yes, but I'm almost sure that doing 

```purescript
do
  _lens1 := x 
  _lens2 := y
  _lens3 := z
```

causes 3 renders in Halogen, while doing

```purescript
modify_ $ 
   set _lens1 x 
   <<< set _lens2 y 
   <<< set _lens3 z
```

only triggers 1 render.

PR #46, opened 2022-01-04
Add halogen-store to marlowe-playground

paluh:


paluh:
This is nearly a copy of @hrajchert change for `marlowe-run`.

It can seems strange that we provide `MonadAsk` instance for our `AppM` which exposes part of our mutable global state but that is a part of the [official `StoreT` API too](https://github.com/thomashoneyman/purescript-halogen-store/blob/v0.3.0/src/Halogen/Store/Monad.purs#L59). I'm still not sure about this approach and my first reaction was also [sceptical](https://github.com/input-output-hk/marlowe-cardano/pull/39#pullrequestreview-843788028) ...

marlowe-playground-client/src/Store.purs
```diff
@@ -0,0 +1,31 @@
...
+type Store
+  = { | StateRow (EnvRow ()) }
+
+instance hasSPSettingsEnv :: HasSPSettings Env where
+  spSettings (Env { ajaxSettings }) = ajaxSettings
+
+toEnv :: Store -> Env
+toEnv { ajaxSettings } = Env { ajaxSettings }
+
+data Action
```
jhbertra, line 28: For now, what do you think about the simple Action type proposed in `HalogenStore`?

```
type Action = Store -> Store
```

marlowe-playground-client/src/Store.purs
```diff
@@ -0,0 +1,31 @@
...
+import Marlowe (class HasSPSettings, SPSettings_)
+import Type.Row (type (+))
+
+type EnvRow :: Row Type -> Row Type
+type EnvRow r
+  = ( ajaxSettings :: SPSettings_ | r )
+
+-- | Immutable part of the global state. We wrap it in a `newtype`
+-- | because we need an instance of `HasSPSettings`.
+newtype Env
```
jhbertra, line 12: I agree with your comment in the PR description - I don't think this is a good idea. I'm not a fan of putting the "read-only" environment in the state like this.  I propose we have `newtype AppM = AppM (ReaderT Env (StoreT Store Aff) a)` instead. That way, the immutable environment cannot be accidentally modified.

jhbertra (changes requested):
I agree with your PR comment, I think we still want to keep the `Env` and the `Store` separate via `ReaderT Env (StoreT Store Aff) a`

paluh:
As discussed earlier I'm closing this because I ended up with empty store.

PR #54, opened 2022-01-10
Turn `Component.MetadataTab` into a proper halogen component

paluh:
This change turns `Component.MetadataTab` tab into a halogen component (using hooks + output actions). I've dropped the idea of moving `MetaData` into the global `Store` because the impact of such a change was really broad and I've found a way to transition into a component by just reusing `output` of the component and already existing `MetadataAction` type.

Pre-submit checklist:
- Branch
    - [ ] Tests are provided (if possible)
    - [x] Commit sequence broadly makes sense
    - [x] Key commits have useful messages
    - [ ] Relevant tickets are mentioned in commit messages
    - [x] Formatting, materialized Nix files, PNG optimization, etc. are updated
- PR
    - [x] Self-reviewed the diff
    - [x] Useful pull request description
    - [x] Reviewer requested

PR #64, opened 2022-01-13
SCP-3188 Metadata reordering using drag and drop

paluh:
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

This PR builds upon the https://github.com/input-output-hk/marlowe-cardano/pull/54 so the first two commits are not strictly part of this change.

I'm introducing here a reusable halogen "renderless component" and a hook. A minimal usage example could be this snippet which shows what is required to add d&d sorting to a static list of elements:

```diff
 type State = Array String

 hookComponent :: forall t5 t6 t8 t9. MonadEffect t9 => H.Component t5 t6 t8 t9
 hookComponent = Hooks.component \_ _ -> Hooks.do
   items /\ modifyState <- useModifyState_ ["one", "two", "three", "four"]
+  let
+    reorder move st = fromMaybe st (Unfoldable.move move st)
+  { genDragHandlers } <- useApplySortable (modifyState <<< reorder)
+
   Hooks.pure $ HH.div_
     [ HH.ul_
-    $ items `flip Array.mapWithIndex` \idx  label ->
+    $ Infinite.Finalize.zip genDragHandlers items `flip Array.mapWithIndex` \idx (Sortable.DragHandlers eventHandlers /\ label) ->
        HH.li
-        []
+        [ eventHandlers.onDragEnter, eventHandlers.onDragEnd, eventHandlers.onDragStart, HP.draggable true ]
         [ HH.text $ label <> " (" <> show idx <> ")"]
     , HH.button
       [ HE.onClick \event -> do
           liftEffect $ preventDefault (MouseEvent.toEvent event)
           let
             label = show $ Array.length items
           modifyState (flip Array.snoc label)
       ]
       [ HH.text "add" ]
     ]
 
 main :: Effect Unit
 main =
   HA.runHalogenAff do
     body <- HA.awaitBody
     void $ runUI hookComponent Nothing body
```

I could separate this into a package and provide an example / docs like the above in the future but now I'm not sure where to put such a doc snippet (`MetadataTab` is not good example of this hook usage because it has complicated rendering function).


palas:
There is something odd on going on with the CSS:
css

And I think it would be more clear to the user to have something like an "=" to the left, for indicating that it can be dragged. Of course, we would need proper UI design for this, but still I think it would be a bit nicer.

I will have a more detailed look at the code later today

paluh:
> There is something odd on going on with the CSS:

We are not able to use one "global css grid" anymore. We need to wrap rows so we are able to drag them. In those parts of the form I've used quite simple flexbox style with predetermined width for some pieces so it still layouts itself aligned. I didn't invest a lot of time into it because it has to be properly designed - should I create a separate issue for this redesign?


> And I think it would be more clear to the user to have something like an "=" to the left, for indicating that it can be dragged.

Shoul I move the current button from the last column to the first or should I wait for the proper redesign of this part?


> Of course, we would need proper UI design for this, but still I think it would be a bit nicer.

Totally agree. Should I create a ticket for this redesign?

palas:
> Totally agree. Should I create a ticket for this redesign?

Yes, I think that would be good

marlowe-playground-client/src/Contrib/Halogen/Components/Sortable/Hook.purs
```diff
@@ -0,0 +1,144 @@
...
+              Sortable.DragHandlers
+                { onDragStart:
+                    Events.onDragStart \event -> do
+                      { dragged, orderingVersion } <- getState
+                      if
+                        ( orderingVersion == state.orderingVersion && dragged /=
+                            Just idx
+                        ) then
+                        modifyState _ { dragged = Just idx }
+                      else do
```
palas, line 76: Also here:
```suggestion
                      if orderingVersion == state.orderingVersion
                         && dragged /= Just idx
                      then
                        modifyState _ { dragged = Just idx }
                      else
```

marlowe-playground-client/src/Contrib/Halogen/Components/Sortable/Hook.purs
```diff
@@ -0,0 +1,144 @@
...
+                        liftEffect $ Event.preventDefault
+                          (DragEvent.toEvent event)
+                , onDragEnd:
+                    Events.onDragEnd
+                      $ \_ -> do
+                          { dragged } <- getState
+                          when (isJust dragged) do
+                            modifyState _ { dragged = Nothing }
```
palas, line 84:
```suggestion
                          when (isJust dragged)
                            $ modifyState _ { dragged = Nothing }
```

marlowe-playground-client/src/Contrib/Halogen/Components/Sortable/Hook.purs
```diff
@@ -0,0 +1,144 @@
...
+                                orderingVersion
+
+                              move = Just { from: dragged, to: idx }
+
+                              state' =
+                                { dragged: Just idx
+                                , orderingVersion: orderingVersion'
+                                , move
+                                }
+                            modifyState $ const state'
```
palas, line 101:
```suggestion
                            dragged /= idx ->
                              modifyState $ const
                                { dragged: Just idx
                                , orderingVersion: Sortable.nextVersion orderingVersion
                                , move: Just { from: dragged,
                                             , to: idx }
                                }
```

palas (changes requested):
Code looks good :)
I wrote a few lints and typo fixes that I found.
TBH I didn't completely understand some bits, but I did some testing too and couldn't find any problems

marlowe-playground-client/src/Contrib/Halogen/Components/Sortable.purs
```diff
@@ -0,0 +1,175 @@
...
+              -- | idices are different etc.
+              MoveTo orderingVersion idx
+    Tuple
+      ( DragHandlers
+          { onDragStart
+          , onDragEnd
+          , onDragEnter
+          }
+      )
+      (idx + 1)
```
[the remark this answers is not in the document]
paluh: I've prefer `do` over `let in` even in pure functions - I hope it is not a problem ;-)

paluh:
Thanks a lot @palas for detailed review. I've incorporated most of the changes.
Unfortunately moving the "drag" button to the front of the row make the styling even more broken so I've decided to leave it at the end and open an appropriate ticket for it.

paluh:
@palas So it took me a while to do this restyling but finally I reverted to a really simple approach - I use fixed (30%) layout for the label row and I use grids also on the row level with nearly identical styling as the form one. 
I also moved the sorting "icon" to the front of the row.

The final layout looks like this:

sortable

P.S. I want to wait with the merge until we merge the `marlowe-run-development` branch - I want to rebase on top of some changes from there.

PR #75, opened 2022-01-26
Introduce smart types for pub key hashes and wallet IDs

jhbertra:
Making a PR for this because I want to make sure the ideas in here are actually good ones.


bwbush (approved):
I think this does a good job of better organizing the wallet types. In the longer term, I think it could be feasible to (nearly) eliminate the use of public key hashes, which will eventually need to be addresses.

marlowe-dashboard-client/src/Capability/Marlowe.purs
```diff
@@ -295,22 +296,10 @@ instance manageMarloweAppM :: ManageMarlowe AppM where
...
   subscribeToPlutusApp = toBack >>> Left >>> Subscribe >>> sendWsMessage
   subscribeToWallet =
-    toBack
-      >>> un Back.Wallet
-      >>> _.getWalletId
-      >>> (\x -> Back.PubKeyHash { getPubKeyHash: x })
-      >>> Right
-      >>> Subscribe
-      >>> sendWsMessage
+    sendWsMessage <<< Subscribe <<< Right <<< WI.toPubKeyHash
```
hrajchert, line 299: This uses the toPubKeyHash which is incorrect, but I noticed that `sendWsMessage` is using `CombinedWSStreamToServer` which is defined in the PAB and exported as 

```purescript
data CombinedWSStreamToServer
  = Subscribe (Either ContractInstanceId PubKeyHash)
  | Unsubscribe (Either ContractInstanceId PubKeyHash)
```
I think that type is wrong and the PubKeyHash should be changed to WalletId. But that is out of the scope of this PR.

marlowe-dashboard-server/src/Marlowe/Run/Wallet/V1/Types.hs
```diff
@@ -0,0 +1,50 @@
...
+
+import qualified Cardano.Wallet.Primitive.Types as Cardano
+import qualified Data.Aeson as Aeson
+import Data.Aeson.Types (parseFail)
+import qualified Data.Text as T
+import Data.Text.Class (FromText (..), TextDecodingError (..), ToText (..))
+import Ledger.Address (PaymentPubKeyHash)
+import Servant (FromHttpApiData (..))
+
+newtype WalletId = WalletId { unWalletId :: Cardano.WalletId }
```
hrajchert, line 16: Why do we define WalletId here? By doing this aren't we making a difference between the WalletId needed by the PAB and the one used by the dashboard-server?

hrajchert (approved):
LGTM... minor notes

marlowe-dashboard-server/src/Marlowe/Run/Wallet/V1/Types.hs
jhbertra: For the instances (the wrapped type does not provide them)

PR #84, opened 2022-02-09
add config updates for current sprint env

ladamesny:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -7,6 +7,7 @@ on:
       - deploy-hernan
       - deploy-pablo
       - shlevy
+      - sprint-*
 jobs:
   deploy:
     runs-on: ubuntu-latest
--- a/scripts/deploy-bitte
+++ b/scripts/deploy-bitte
@@ -17,7 +17,7 @@ then
 fi
 
 ref="${GITHUB_REF_SLUG:-}"
-ref="${ref}sprint-${[0-9]} | sed -e 's/[0-9]//g'"
+ref="${ref//[0-9]/}"
 
 if [ -z "${ref}" ]
 then
--- a/scripts/deploy-bitte
+++ b/scripts/deploy-bitte
@@ -26,6 +26,7 @@ fi
 declare -A ref_env_mapping=(
     [production]=production
     [main]=staging
+    [sprint-*]=currentSprintMarlowe
     [deploy-hernan]=hernan
     [deploy-pablo]=pablo
     [shlevy]=shlevy
--- a/scripts/deploy-bitte
+++ b/scripts/deploy-bitte
@@ -16,7 +16,8 @@ then
     error "GITHUB_SHA environment variable not set!"
 fi
 
-readonly ref="${GITHUB_REF_SLUG:-}"
+ref="${GITHUB_REF_SLUG:-}"
+ref="${ref/v[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]/vYYYY-MM-DD}"
 
 if [ -z "${ref}" ]
 then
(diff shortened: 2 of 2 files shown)


scripts/deploy-bitte
```diff
@@ -26,6 +26,7 @@ fi
 declare -A ref_env_mapping=(
     [production]=production
     [main]=staging
+    [sprint-*]=currentSprintMarlowe
```
shlevy, line 29: See https://github.com/input-output-hk/marlowe-cardano/pull/84/files#diff-c69d8b18d9f3cd934d935bf02255f1d1746031d582d2ef8d4cf799e927ddee1aL34 , we currently just do a direct hash lookup which will fail here. We can probably do something similar to https://github.com/input-output-hk/plutus-apps/blob/main/scripts/deploy-bitte#L19-L21 to transform `ref` prior to doing the lookup

shlevy (changes requested):
See above comment

scripts/deploy-bitte
ladamesny: Thanks @shlevy 

I copied over that ref regex from the bitte file in plutus apps, however I'm not sure if understanding this file correctly.

Is `ref` supposed to reference a branch name from the marlowe-cardano repo? Then then we use it as a key to find a matching environment in the ref_env_mapping?

shlevy: The idea is that we look up `ref` in `ref_env_mapping`. So we don't want to copy the regex exactly from plutus-apps, there they want to match on a tag of the form `vYYYY-MM-DD` so the regex transforms any tag matching that format into the literal string "vYYYY-MM-DD", which string is then stored in ref_env_mapping.

So you'd want a regex that replaces sprint-12 with "sprint-N" or something and then store "sprint-N" in the mapping.

ladamesny: @shlevy ok thank you - that's clearer for me. Ok so I updated the regex. I'm basically replacing `sprint-12` to `sprint-` - it was just an easier regex to write. 

So I updated the command - I feel good about `| sed -e 's/[0-9]//g` part as I tested it on the terminal, but I'm not sure about the portion before it. What's the best way to test that `ref="${ref}sprint-${[0-9]}"` would match `sprint-12`?

scripts/deploy-bitte
```diff
@@ -16,7 +16,8 @@ then
     error "GITHUB_SHA environment variable not set!"
 fi
 
-readonly ref="${GITHUB_REF_SLUG:-}"
+ref="${GITHUB_REF_SLUG:-}"
+ref="${ref}sprint-${[0-9]} | sed -e 's/[0-9]//g'"
```
shlevy, line 20: This isn't right. You can test this out locally by doing:

```
$ ref=sprint-23
$ ref=<expression you want to test>
$ echo $ref
```

I recommend just using the bash replacement style like I did in plutus-apps. See https://tldp.org/LDP/abs/html/string-manipulation.html#SUBSTRREPL00

shlevy (changes requested):

scripts/deploy-bitte
ladamesny: ok I think I got it now - the new regex works locally. Thanks for your help on this @shlevy

.github/workflows/deploy.yml
```diff
@@ -7,6 +7,7 @@ on:
       - deploy-hernan
       - deploy-pablo
       - shlevy
+      - sprint-*
```
ladamesny, line 10: oops, I guess I don't need this star anymore - I'm going to remove

ladamesny: or should I @shlevy ? I think I need this. Going to leave this alone till I hear from you.

shlevy (approved):

PR #85, opened 2022-02-09
SCP-3044: Marlowe example contracts

yveshauser:
The PR proposes example Marlowe contracts to be added to a newly created sub-project: _marlowe-contracts_

The sub-project has been introduced in order to start collecting Marlowe contracts at a single location and build up a library of contracts.

To start with, I added contracts for swaps, zero-coupon bonds and more complex contracts like options, futures and structured products. Some contracts are combinations of others. In order to concurrently combine two contracts, the `both` helper function was introduced (contracts built like this tend to be big though).

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


jhbertra (changes requested):
Please update the base branch from `main` to `sprint-49` and rebase

marlowe-contracts/test/Spec/Marlowe/Analysis.hs
```diff
@@ -0,0 +1,81 @@
...
+  assertBool "Analysis ok" $ isRight result && either (const False) f result
+  where
+    f | b = isJust
+      | otherwise = isNothing
+
+testNoWarnings :: Contract -> IO ()
+testNoWarnings = testWarnings False
+
+testExpectedWarnings :: Contract -> IO ()
+testExpectedWarnings = testWarnings True
```
palas, line 42: This one was trickier than I thought. I think it is easier to understand like this:
```suggestion
hasWarnings :: Contract -> IO Bool
hasWarnings contract = do
  result <- warningsTrace contract
  case result of
    Left errDesc -> assertFailure ("Analysis failed: " ++ show errDesc)
    Right counterExample -> return $ isJust counterExample

testNoWarnings :: Contract -> Assertion
testNoWarnings contract =
  assertBool "Has no warnings" =<< not <$> hasWarnings contract

testExpectedWarnings :: Contract -> Assertion
testExpectedWarnings contract =
  assertBool "Has warnings" =<< hasWarnings contract
```
And also output in case of error should be clearer. Hopefully I didn't make any mistake

marlowe-contracts/test/Spec/Marlowe/Contracts.hs
```diff
@@ -0,0 +1,395 @@
...
+          w2Pk
+          (Slot 100)
+          (Slot 200)
+          (Constant 75_000_000)
+          (Constant 90_000_000)
+          ada
+          Close
+   in either left right $
+        step (TransactionInput (0, 0)     [NormalInput $ IDeposit w1Pk w1Pk ada 75_000_000]) ([], [], emptyState 0, zcb)
+    >>= step (TransactionInput (100, 110) [NormalInput $ IDeposit w1Pk w2Pk ada 90_000_000])
```
palas, line 83: There is a `playTrace` function defined in https://github.com/input-output-hk/marlowe/blob/master/src/Language/Marlowe/Semantics.hs#L417, I didn't realise it wasn't in this repo. Would you mind copying it to `Language.Marlowe.Semantics` and using that instead of `Either`? `playTrace` is the one we verified several properties for, and will make all this code much simpler cause it just becomes pattern matching on `TransactionOutput`

marlowe-contracts/test/Spec/Marlowe/Contracts.hs
```diff
@@ -0,0 +1,395 @@
...
+swapContractTest =
+  let backSwap =
+        swap w1Pk ada (Constant 10_000_000) w2Pk tok (Constant 30) (Slot 100) $
+        swap w2Pk ada (Constant 10_000_000) w1Pk tok (Constant 30) (Slot 200) Close
+   in either left right $
+        step (TransactionInput (0, 0)
+              [ NormalInput $ IDeposit w1Pk w1Pk ada 10_000_000
+              , NormalInput $ IDeposit w2Pk w2Pk tok 30
+              , NormalInput $ IDeposit w2Pk w2Pk ada 10_000_000
+              , NormalInput $ IDeposit w1Pk w1Pk tok 30
```
palas, line 102: Yes, that way we could do this in separate transactions too, for example, if we wanted

palas (changes requested):
Looks good, I've added some ideas to simplify the code

bwbush (approved):
Looks good.

FYI, these examples are compatible with the way that `marlowe-cli` provides templates for contracts, so we can add support for these there, too.

palas (approved):

yveshauser:
> Please update the base branch from `main` to `sprint-49` and rebase

Changed base to `sprint-50`

PR #89, opened 2022-02-16
Jhbertra/addresses

jhbertra:
Updates the create and redeem endpoints to accept Shelley Addresses. This is a bit of a hack, and support for addresses should be properly pushed upstream into Plutus.


marlowe/src/Language/Marlowe/Client.hs
```diff
@@ -128,7 +133,7 @@ data PartyAction
              | CloseContract
   deriving (Show)
 
-type RoleOwners = AssocMap.Map Val.TokenName PubKeyHash
+type RoleOwners = AssocMap.Map Val.TokenName (AddressInEra ShelleyEra)
```
bwbush, line 136: `AddressInEra ShelleyEra` should work. We may want to adjust this eventually, when for instance we allow scripts to hold role tokens.

marlowe/src/Language/Marlowe/Client.hs
```diff
@@ -465,15 +472,50 @@ setupMarloweParams owners contract = mapError (review _MarloweError) $ do
...
+              mustPayToShelleyAddress addr (Val.singleton rolesSymbol role 1 <> minAdaTxOut)
+        distributeRoleTokens <- foldMapM giveToParty $ AssocMap.toList owners
         let params = marloweParams rolesSymbol
         pure (params, mintTx <> distributeRoleTokens, lookups)
     else do
         let missingRoles = roles `Set.difference` Set.fromList (AssocMap.keys owners)
         let message = T.pack $ "You didn't specify owners of these roles: " <> show missingRoles
         throwing _ContractError $ OtherError message
 
+ownShelleyAddress
```
bwbush, line 485: We might consider a PR to move these into `plutus-apps`, but the change would involve adding a constructor to the constraint type there.

marlowe/src/Language/Marlowe/Client.hs
```diff
@@ -465,15 +472,50 @@ setupMarloweParams owners contract = mapError (review _MarloweError) $ do
...
+      in
+        case Shelley.fromShelleyStakeReference stakeRef of
+          StakeAddressByValue (StakeCredentialByScript _) ->
+            throwError $ OtherContractError $ OtherError "Script stake addresses not supported"
+          StakeAddressByPointer _ ->
+            throwError $ OtherContractError $ OtherError "Pointer stake addresses not supported"
+          NoStakeAddress -> pure (ppkh, Nothing)
+          StakeAddressByValue (StakeCredentialByKey stakeHash) ->
+            pure (ppkh,  Just . StakePubKeyHash . PubKeyHash . toBuiltin $ serialiseToRawBytes stakeHash)
+shelleyAddressToKeys _ = throwError $ OtherContractError $ OtherError "Byron Addresses not supported"
```
bwbush, line 518: It actually wouldn't be difficult to support Byron addresses, the `Cardano.Api.Byron` module has address functions similar to `Cardano.Api.Shelley`.

bwbush (approved):
Looks good!

I'll need to adjust some of the still non-automated tests for the PAB, but that will be trivial, and I can do that after this is merged.

marlowe/src/Language/Marlowe/Client.hs
jhbertra: `AddressInEra` has a `FromJSON` instance, `AddressAny` doesn't

marlowe/src/Language/Marlowe/Client.hs
jhbertra: I'll merge this PR and we can add it if/when we need it

PR #97, opened 2022-02-28
Scp 3479

paluh:
Let's discuss metadata server spec under this PR. It is an ongoing draft and I have plans to improve it but I wan to work on it in more public setup and get some feedback ;-)

### DIFF
--- a/doc/metadata-server/README.md
+++ b/doc/metadata-server/README.md
@@ -1,0 +1,321 @@
+## Abstract
+
+This document summarizes requirements and proposes a draft of a protocol for metadata information exchange for the on-chain transactions (`Tx`) which execute Marlowe contracts. The off-chain metadata storage mechanism does not rely on a centralized registry but provides schema which can be implemented by independent metadata providers.
+
+## Motivation
+
+Marlowe is as a high level level language with a set of properties which are crucial for writing reliable financial contracts. A possibility to audit the code of the contract which was already initiated on-chain seems to be critical to any participant "invited" to it. Marlowe (and its current implementation on chain) gives us access to the source code of the contract, ability to do static analysis over it and "step by step" simulation so it provides the fundamental layer of trust for all users.
+To build a contract exchange platform and execution environment we want also to be able to associate and provide contract information suitable for human consumption (like title, description, statistics and certification etc.). We want to verify that provided metadata informations about the contract on chain are consistent with the contract content itself. Additionally we should allow the contract initiatior to attach additional set of data (like documents) to the particular contract instance and present it as well to all other participants. We should provide a way to verify that given custom metadata are also dedicated for the running on-chain contract.
+
+We want to implement this in an efficient manner avoiding huge payloads on the blockchain and by use of existing or emerging Cardano standards. Additionally we need to keep the protocol decentralized so anybody can take a role of a metadata provider.
+
+### Scope limits
+
+We can identify additional sources of metadata informations which can be associated with contract transactions (`Tx`) running on the blockchain. There are also parts of proposed metadata format which we should extend in the future. These parts/problems we want to tackle in a separate proposals:
+
+- Information associated with role tokens (icon, currency description etc.) which are presented in wallets and blockchain explorers.
+
+- Identity verification of contract participants and exchange of information related with these identities (nicknames, credibility, certificates etc.).
+
+- Metadata format and enhancement:
+
+  - Localization,
+
+  - Roles / choices identifiers in the metadata layer so we can complately separate presentation from the contract code,
+
+- Encryption layer. We assume now that exchanged data are public or that the access to the information is restricted by a separate layer of the metadata server. In the future we want to probably propose some encryption schema as a part of the protocol.
+
+## Verifing Marlowe Contract Metadata on the chain
+
+### Terminology
+
+- Marlowe contract (`Contract`) - a particular contract expression which can be executed on chain which have all particiapants and values (like amounts, timeouts etc.) set.
+
+- Marlowe contract transaction (`MarloweTx`)- transaction on the block chain which is a part of the ongoing Marlowe contract.
+
+- Marlowe contract metadata (`ContractMetadata`) - a particular set of information which we want to attach to a particular Marlowe contract (description, documents / agreements etc.)
+
+- Transaction metadata (`TxMetadata`) - on chain information (`CBOR` encoded- [spec](https://github.com/input-output-hk/cardano-node/blob/master/doc/reference/tx-metadata.md)) which is attached to the transaction . It is inaccesible to the transaction validators and minting policy scripts.
+
+- Marlowe contract transaction metadata (`MarloweTxMetadata`) - metadata attached to the on chain transaction executing Marlowe Contract (`MarloweTx`).
+
+### `Contract` extraction
+
+When initiator submits the `Contract` to the blockchain it builds up a trasaction which outputs define role tokens distribution alongside with the "contract continuation" output which is guarded by `Marlowe` interpreter script. Role tokens outputs are assigned to particular addresses chosen by the initiator - we can interpret them as invitations to the contract or permissions to be a contract party (this tokens shows up in the user wallet).
+Because continuation output from the initial `Tx` contains the full code (or `Merkleized` version of it) of the `Contract` we are able to [extract it](https://github.com/input-output-hk/marlowe-cardano/blob/f452d98942a14d34c10f2963ec4b84147da5621c/marlowe/src/Language/Marlowe/Client/History.hs). In other words when someone is "invited" to the already initiated on chain contract we are able to extract the Marlowe code from that transaction and present it to the possible participant for an audit, simulation etc. We can do this at any stage of the contract because we can track back to the initial transaction and perform the extraction.
+
+<!-- [ IMAGE -> Extraction of the contract -> Simulation / Audit ] -->
+
+### `MarloweTxMetadata`
+
+The submitter of the transaction have an ability to attach additional (`CBOR` encoded) [data to it](https://github.com/input-output-hk/cardano-node/blob/master/doc/reference/tx-metadata.md). This data are not accessible to the transaction validators so they can be arbirary informations not necessarly connected and valid to the `Tx` real content. To simplify the disussion for now and assume that there are no size limits and costs associated with `TxMetadata` and we use this layer to store all contract metadata.
+
+FIXME:
+How do we detect which transactions / tokens are the marlowe ones.
+
+FIXME (move / fix this):
+It is not trivial currently to identify which transactions are Marlowe contracts currently but we can use `TxMetadata` layer for this purpose too.
+The real limitiations and possible usage scenarios are discussed in the [Implemeantation](# implementation) part.
+
+### `MarloweTxMetadata` verification
+
+We can imagine that we want to endorse a particular contract with a set of certified documents (agreements, audits etc.). We don't want to necessarily trust the user who creates the `MarloweTx` so we should assume that the contract code can be incosistent with `TxMetatdata` which provide the documents data. The simple protocol could require metadata to be signed together with the Marlowe contract so the connection between those pieces is clear.
+
+<!-- IMAGE -> Extraction of the contract -> Extraction of the metadata -->
+
+Because `Marlowe` abstracts over addresses using roles the full meaning of a particular instance of a contract is also dependent "role distribution". It seems that semantic meaning of a `Contract` on chain is fully defined by triple:
+(`Contract`, role distribution, `MarloweValidator`). Using a pseudo code and removing a lot of implementation details we can provide trivial validation of `MarloweTxMetadata`:
+
+```purescript
+type RoleDistribution = Map Role PubKeyHash
+
+type MarloweTxMetadata =
+  { signature :: Signature
+  , payload ::
+    { metadata :: MarloweContractMetatada
+    , contract :: Contract
+    , roleDistribution :: RoleDistribution
+    }
+  }
+
+extractMetadata :: OnChainTx -> Maybe Metadata
+extractMetadata tx = do
+  marloweTxMetadata <- decode <$> extractMetadata tx
+  txContract <- extractMarloweContract tx
+  txValidator <- extractValidator tx
+  txRoleDistribution <- extractRoleDistribution tx
+
+  if txMetadata.payload.contract == txContract
+    && txMetadata.payload.roleDistribution == txRoleDistribution
+    && checkValidator txValidator
+    && checkSignature txMetadata.payload pubKey MarloweTxMetadata.signature
+      then
+        Just MarloweTxMetadata.metadata
+      else
+        Nothing
+```
+
+> In the current implemenatation we need also `MarloweParams` to derive the final validator but we can treat this pieces as an implementation detail. It the context of [CIP-33](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0033) it seems beneficial to move `MarloweParams` to datum so the validator script could be shared .
+
+It may seem a bit articial or unrealistic that we want to delegate submition process to some third parties and not rely on the submitter of the `MarloweTx` as the author of metadata attachments. On the other hand there are some cases where this separation seems to be useful:
+
+- Marlowe Marketplace can possibly perform submitions of paid and certified contracts on behalve of the user. We may want to create this transactions ourselves so we can confirm to others that we really give a credit to a given transaction and we associate it with certified contract template. User should pass to us contract parameters together with signed `MarloweTxMetadata` (which contains the required triple) so we can still proove to others that a given `MarloweTx` was created and described by a particular party.
+  Of course we are not able to prevent users to use certified / paid templates on their own (without paying fee) but in such a case we are not going to endorse these transactions on chain.
+
+- In multiparty agreement it can be not so important who is the owner of the initial transaction (contract participant, some third party auditor, layer etc.). On the other hand it is rather important that the attachment for a transaction was really dedicated for its semantic meaning.
+
+- Using these schema we can require distribution of role tokens to a specific, generic utility scripts with constant addresses (like "deposit scripts") and provide the same audited Marlowe contract with the same role distribution which can be safely reused and certified. By using this schema we can push a possible role distribution logic outside of our Marlowe interpreter / validator (we don't have to change our implementation) to role scripts (with single and simpler to audity responsiblity) and distribute reusable instances of Marlowe contracts together with verified set of metadata (authorized for this Conract and hard coded role to scripts distribution).
+
+## Extended Marlowe
+
+### Terminology
+
+This is analogus terminology which we have in our codebase:
+
+- Extended Marlowe contract (`ExtContract`) - marlowe contract which contains possible placeholders for values. We use it to parametrize amounts but in these parameters can use all value expressions. Marlowe playground allows us to build these kind of contracts.
+
+- Extended contract instance (`Instance`) - extended contract which has all parameters specified / filled in. In essence it is just a proper Marlowe contract but we can associate it back with some contract template - we can point to its "source" and possibly authors, descriptions, parameters etc.
+
+- Extended contract metadata (`ExtMetadata`) - metadata suitable for human consumtion about a particular extended contract (name, description, choice, role information etc.). We don't have an official schema definion for it but only a reference implementation. The schema / spec would be probably required if we want to allow other implementations / instances of metadata servers.
+
+- Contract template (`Template`) - `ExtContract` together with its `ExtMetadata`. This would be probably the main type of artifact which we are going to distribute through Marketplace and which we want to certify so code and its human readble description are verified together.
+
+- Instance metadata (`InsMetadata`) - part of the metadata dedicated related only to the `Instance`. For example we can imagine that the user picks a `ExtContract` from the `Marketplace` and in `Marlowe Run` he is be able to add its own information set to this particular instance (add own description, describe role differently, change title etc.). Parts of this metadata set could be probably complately service dependent (for example in different instances / deployments of `Marlowe Run` we can use different schema here and allow attachment of extra documents).
+
+- Marlowe contract metadata (`ContractMetadata`) - the same as in the previous section but now we can look at it as a product: `(InsMetadata, ExtMetadata)`.
+
+# Verification of `ContractMetadata`
+
+In the current scenario we want to verify validity of two parts of `ContractMetadata` (`(InsMetadata, ExtMetadata)`) to do this we can slightly modify the previous verification schema. `InsMetadata` can be verified using the same strategy as before because it is information dedicated for a specific Contract so we need to prove the connection of the metadata with onchain triple: roles distribution, `Validator` and `Contract`.
+
+When we want to verify the `ExtContract` and `ExtMetadata` are really associated with the data on chain we should prove that there is a set of parameters which when applied to `ExtContract` produced the same `Contract` value as we have in this particular `Tx`. The validation of `ExtContract` and `ExtMetadata` association should be provided by a seperate certificate.
+
+<!-- IMAGE Extended contract comparision flow. -->
+
+```purescript
+type RoleDistribution = Map Role PubKeyHash
+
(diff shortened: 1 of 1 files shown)


doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
+## Abstract
+
+This document summarizes requirements and proposes a draft of a protocol for metadata information exchange for the on-chain transactions (`Tx`) which executes Marlowe contracts. The off-chain metadata storage mechanism does not rely on a centralized registry but provides schema which can be implemented by independent metadata providers.
+
+## Motivation
+
+Marlowe is a high-level language with a set of properties which are crucial for writing reliable financial contracts. A possibility to audit the code of the contract which was already initiated on-chain seems to be critical to any participant "invited" to it. Marlowe and its current implementation on Cardano chain gives us access to the source code of the contract, ability to do static analysis over it and "step by step" simulation so it provides the fundamental layer of trust for the language users.
```
bwbush, line 7: I suggest changing "code" to "source code" just to clarify.

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+We want to implement this in an efficient manner avoiding huge payloads on the blockchain and by use of existing or emerging Cardano standards. Additionally we need to keep the protocol decentralized so anybody can take the role of a metadata provider.
+
+### Scope limits
+
+We can identify additional sources of metadata information which can be associated with contract transactions running on the blockchain. There are also parts of the proposed metadata format which we should extend in the future. These parts/problems we want to tackle in a separate proposals:
+
+- Role token metadata (icon, currency description etc.) handling. They are presented in Cardano wallets and blockchain explorers.
+
+- Localization (`l10n`) layer for the metadata format.
```
bwbush, line 19: Good!

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+## Verifying Marlowe Contract Metadata on the chain
+
+### Terminology
+
+- Marlowe contract (`Contract`) - a particular contract expression which can be executed on-chain which has all participants and values (like amounts, timeouts etc.) set.
+
+- Marlowe contract transaction (`MarloweTx`)- transaction on the block chain which is a part of the ongoing Marlowe contract.
+
+- Marlowe contract metadata (`ContractMetadata`) - a particular set of information which we want to attach to a Marlowe contract (description, documents / agreements etc.)
```
bwbush, line 35: We may want to clarify that this metadata may be on chain or off chain.

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+- Role distribution (`RoleDistribution`) - when contract uses roles as `Party` values we use NFTs to represent roles on-chain. The initial transaction pay these NFTs to some addresses so it defines this initial distribution.
+
+- Transaction metadata (`TxMetadata`) - on-chain information ([`CBOR` encoded](https://github.com/input-output-hk/cardano-node/blob/master/doc/reference/tx-metadata.md)) which is attached to the transaction . It is inaccessible to the transaction validators and minting policy scripts.
+
+- Marlowe contract transaction metadata (`MarloweTxMetadata`) - metadata attached to the on-chain transaction executing Marlowe Contract (`MarloweTx`) related to the `Contract`.
+
+
+### `Contract` extraction
+
+When initiator submits a `Contract` to the blockchain it builds up a transaction which outputs define role tokens distribution. Additionally transaction contains "continuation" output which is "guarded" by `Marlowe` interpreter script. Role tokens outputs are assigned to a set of addresses chosen by the `Tx` author. We can interpret them as invitations to the contract or permissions to be the contract party (this tokens shows up in the user wallets).
```
bwbush, line 46: Change "outputs" to "mints".

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+Because continuation output from the initial `Tx` contains the full code ("modulo" `Merkleization`) of the `Contract` we are able to [extract it](https://github.com/input-output-hk/marlowe-cardano/blob/f452d98942a14d34c10f2963ec4b84147da5621c/marlowe/src/Language/Marlowe/Client/History.hs). In other words when someone is "invited" to the already initiated on-chain contract we are able to extract the Marlowe code from that transaction and present it to the possible participant for an audit, simulation etc. before he / she commits to it. We can do this at any stage of the contract progression because we can track the chain back to the initial transaction and perform the extraction.
```
bwbush, line 48: Add "Merkleization" to the terminology list.

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+<!-- [ IMAGE -> Extraction of the contract -> Simulation / Audit ] -->
+
+### `MarloweTxMetadata`
+
+The submitter of the transaction have an ability to attach additional [`CBOR` encoded data to it](https://github.com/input-output-hk/cardano-node/blob/master/doc/reference/tx-metadata.md). This data are not accessible to the transaction validators so they can be arbitrary information not necessarily valid and related to the `Tx` content. To simplify the discussion for now and assume that there are no size limits and costs associated with `TxMetadata` and we use this layer to store all contract metadata.
```
bwbush, line 54: - Change "have an ability" to "has the ability".
- Change "This data are" to "These data are" or "This data is".

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+### `MarloweTxMetadata` verification
+
+We can imagine that we want to endorse a particular contract with a set of certified documents (agreements, audits etc.). We don't want to necessarily trust the user who creates the `MarloweTx` so we should assume that the contract code can be incosistent with `TxMetatdata` which provides the documents data. The simple protocol could require metadata to be signed together with the Marlowe contract so the connection between those pieces is clear.
+
+<!-- IMAGE -> Extraction of the contract -> Extraction of the metadata -->
+
+Because `Marlowe` abstracts over addresses using roles the full meaning of a particular instance of a contract is also dependent on role distribution. It seems that semantic meaning of a `Contract` on-chain is fully defined by the triple: `(Contract, RoleDistribution, MarloweValidator)`. Using a pseudo code we can provide this basic validation flow for `MarloweTxMetadata`:
```
bwbush, line 62: Technically, the tuple should include the initial state of the contract, too.

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+> It is worth to mention that in the context of [CIP-33](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0033) it seems beneficial to move `MarloweParams` to the `datum` so the validator script could be shared and the transaction size could be minimized.
+
+It may seem a bit artificial or unrealistic that we want to delegate submission process to some third parties and not rely on the submitter of the `MarloweTx` as the author of metadata attachments. On the other hand there are some cases where this separation seems to be useful:
+
+- Marlowe Marketplace can possibly perform submissions of paid and certified contracts on behalve of the user. We may want to perform creation of these transactions by the platform so we can confirm to others that we really give a credit to a given transaction and we associate it with certified contract template. User should pass to us contract parameters together with signed `MarloweTxMetadata` (which contains the required triple) so we can still proove to the others that a given `MarloweTx` was created and described by a particular party.
```
bwbush, line 98: "Marlowe Marketplace" -> "Marlowe Marketplace or other service providers"

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+  Of course we are not able to prevent users from using certified / paid templates on their own (without paying fee) because in some sens all Marlowe contracts are open sourced (at least through inspection of the `Tx`). In such a case we are not going to endorse these transactions on-chain and on the platform.
```
bwbush, line 100: "sens" -> "sense"

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+- In multiparty agreement it can be not so important who is the owner of the initial transaction (contract participant, some third party auditor, layer etc.). On the other hand it is rather important that the attachment for a transaction was really dedicated for its semantic meaning.
+
+- Using these schema we can require distribution of role tokens to a specific, generic utility scripts with constant addresses (like "deposit scripts") and provide the same audited Marlowe contract with the same role distribution which can be safely reused and certified. By using this schema we can push a possible role distribution logic outside of our Marlowe interpreter / validator (we don't have to change our implementation) to role scripts (with single and simpler to audit responsiblity) and distribute reusable instances of Marlowe contracts together with verified set of metadata (authorized for this Contract and hard coded role to scripts distribution).
```
bwbush, line 104: Nice. I hadn't thought of this. Sometime we should document use cases for role token workflows.

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+## Extended Marlowe
+
+### Terminology
+
+This is analogus terminology which we have in our codebase:
+
+- Extended Marlowe contract (`ExtContract`) - marlowe contract which contains possible placeholders for values. We use it to parametrize amounts but in these parameters can use all value expressions. Marlowe playground allows us to build these kind of contracts.
```
bwbush, line 112: I believe that @hrajchert raised the possibility of phasing out extended contracts.

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+- Extended contract instance (`Instance`) - extended contract which has all parameters specified / filled in. In essence it is just a proper Marlowe contract but we can associate it back with some contract template - we can point to its "source" and possibly authors, descriptions, parameters etc.
+
+- Extended contract metadata (`ExtMetadata`) - metadata suitable for human consumtion about a particular extended contract (name, description, choice, role information etc.). We don't have an official schema definion for it but only a reference implementation. The schema / spec would be probably required if we want to allow other implementations / instances of metadata servers.
```
bwbush, line 116: "consumtion" -> "consumption"

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+We should expand this in the next iterations into a [detailed schema](https://github.com/input-output-hk/cardano-node/blob/master/doc/reference/tx-metadata.md#detailed-schema) which is translated into a precise CBOR representation (both `cardano-cli` and [`node-wallet`](https://input-output-hk.github.io/cardano-wallet/user-guide/common-use-cases/TxMetadata) support it). We want to use precise CBOR encoding because this exchange protocol should be tool and language agnostic.
+
+#### Storage
+
+Using blockchain as the only source of data has many advantages - we don't have to rely on things like hosting, domain registration, maintenance of the data storage systems etc. On the other hand, storing data on-chain has its cost because it increases transaction fees. `TxMetadata` size contributes to the overall transaction size (without any additional fees). Additionally transactions have currently [64KB](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0009/README.md#updatable-protocol-parameters) size limit which is an additional "risky" factor.
```
bwbush, line 207: Actually, it might increase the transaction fee. We should check experimentally.

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+We try to support both approaches of metadata distribution i.e. on-chain and off-chain by using optional keys in the schema. Of course on-chain metadata are immutable and we don't provide update mechanismfor them (i.e. [NFT metadata in theory supports updates](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0025#update-metadata-link-for-a-specific-token)). Instead we allow usage of both versions and off-chain metadata source (if responsive) can be used for metadata updates.
```
bwbush, line 209: Off-chain metadata can be immutable if the hash of the off-chain data is included on the chain.

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+### Draft of the metadata schema
+
+Following [CIP-0026 convention](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0026):
+
+- We use [_Blake2b-256_](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0026#hash-function),
+
+- We use [this encoding](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0026#attestation-signatures) to construct messages for signing.
```
bwbush, line 217: Could we generalize this to multiple signature and hash methods? Perhaps there could be a "type" field that specifies which CIP the signature conforms to, or whether it is a W3C DID?

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+```js
+contract = serialize(Contract)
+
+roleDistribution = [{ "name": <string>, "pubKeyHash": <string> }]
+
+insId = hash(
+  hash(CBOR(contract)) +
+  hash(CBOR(roleDistribution)) +
+  validatorHash
```
bwbush, line 231: We probably also need the hash of the initial state of the contract.

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+#### Tx metadata schema
+
+```js
+contractMetadata = {"instance": <insMetadata>, "extended": <extMetadata>}
+
+txMetadata = {
+  "1564": {
+      "on": <contractMetadata>
+      "off": {
+        url: <string>
```
bwbush, line 303: Maybe add "hash" here? However, NFT metadata usually just uses the "ipfs://" prefix along with the hash.

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+## Implementation
+
+### Prototyping metadata server
+
+For an initial protyping phase of the metadata server we can use libraries from (`input-output-hk/offchain-metadata-tools`)[`https://github.com/input-output-hk/offchain-metadata-tools`]. This project provides a basic server and simple database layer (based on `postgresql`) for key-value service which implements [CIP-0026 - Cardano Off-Chain Metadata](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0026).
+In the case of Marketplace we want to probably expose only metadata subjects for the instances of contracts (like `/metadata/{ insId }/`) and expose templates and their metadata using other routing patterns (like `/extended/{ extContractId }/`) and use redirects for properties (`/metadata/{ insId }/property/extended`). Of course if we want to flatten both resources into the metadata subjects we can do this by using some encoding for inheritance on the `SQL` level but this comes with its own overhead and makes typing and schema of the API more complicated.
```
bwbush, line 322: I've been disappointed by the lack of Cardano's adoption of RDF (Resource Description Framework) metadata in general and Dublin Core Metadata in particular. Those are well thought out and extensible standards that align with decentralization. Should this document mention that semantic web metadata standards are available and are consistent with this proposal?

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+### `TxMetadata`
+
+In the current implementation we use `PAB` to submit a transaction to the blockchain. It seems like support for the metadatadata handling is not implemented [by this layer](https://github.com/input-output-hk/plutus-apps/issues/93#issuecomment-1055562155).
```
bwbush, line 326: Actually, within the `create`, `apply-inputs`, and `redeem` endpoints we should be able to append metadata to the transaction before the transaction is sent for balancing.

bwbush (changes requested):
I think this is close to what we need at the stage where we want to accelerate discussion of a metadata standard for Marlowe.

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+To build a contract exchange platform (Marlowe Marketplace) and execution environment (Marlowe Run) we want to be able to also associate and provide contract template information suitable for human consumption (like title, description, statistics and certification etc.). We want to verify that provided metadata information about the transaction template is consistent with the Marlowe contract embedded in the transaction on-chain. Additionally we should allow the contract initiator to attach a specific set of data (like documents, agreements) to the particular contract instance. We want to be able to provide a validity proof for these data and present it to all other participants as well if necessary.
+
+We want to implement this in an efficient manner avoiding huge payloads on the blockchain and by use of existing or emerging Cardano standards. Additionally we need to keep the protocol decentralized so anybody can take the role of a metadata provider.
```
palas, line 11: I think this could be much more to the point. Focusing on explaining the reader why we need this. For example:
> When participating on a Marlowe contract only a small amount of information is stored on-chain. Most of the time we can obtain the full code of the contract by looking on-chain, but this is not even true if the contract is merkleized.
> 
> In addition to the actual contract, there are several other pieces of information that would be useful for the participants to know, and without a metadata server there is no standard way for them to get such information.
> 
> Useful information that participants would like to be able to obtain from running contracts include:
> - Role descriptions
> - The template that produced the contract
> - Audit information (i.e: a signed certification from an auditor)
> - etc
> 
Then the issues we face (you did write them, but would be good to add justification):
> A naïve solution would be just add all the information to the chain, but that would be very expensive, bloat the chain, and still it would not be safe, cause someone could forge it. We could just store it in a centralised server, but we want the system to be decentralised to match the nature of the blockchain and to prevent a single point of failure. And it would be desirable to follow existing standards to promote interoperability.

And then an overview of the plan would be good before going into the detail too:

> In order to solve all these problems, this document:
> - Proposes using servers that follow the standard CIP-26
> - Defines a schema for using in conjunction with the standard CIP-26
> - Describe a system to link entries in the schema to contracts on-chain, and a way to verify those links so that entries cannot be forged by bad actors
> - etc

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+### Scope limits
```
palas, line 13: I would move this section to the end of the document (cause it is not necessary to understand it)

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+- Separation of roles / choices identifiers from the presentation layer. These identifiers should have representation in the metadata part as well.
+
+- Encryption layer. We assume now that exchanged data are public or that the access to the information is restricted by a separate layer of the metadata server. In the future we want to probably propose some encryption schema as a part of the protocol.
+
+- Proposal for "identity layer" using Atala PRISM. With identities we can associate and exchange some metadata information like nicknames, credibility etc.
+
+## Verifying Marlowe Contract Metadata on the chain
+
+### Terminology
```
palas, line 29: I would also leave terminology to the end. Best to explain it as you need it. Could add a glossary at the end for completion

palas: Also there are three types of metadata and I cannot tell the difference. I think it could be more useful to the reader to have a discussion of how they are different instead

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+- Localization (`l10n`) layer for the metadata format.
+
+- Separation of roles / choices identifiers from the presentation layer. These identifiers should have representation in the metadata part as well.
+
+- Encryption layer. We assume now that exchanged data are public or that the access to the information is restricted by a separate layer of the metadata server. In the future we want to probably propose some encryption schema as a part of the protocol.
+
+- Proposal for "identity layer" using Atala PRISM. With identities we can associate and exchange some metadata information like nicknames, credibility etc.
+
+## Verifying Marlowe Contract Metadata on the chain
```
palas, line 27: This jumps to detail directly. "Verifying Marlowe Contract Metadata on the chain". Who put metadata on-chain? Why do we have to verify it? What do we need to verify about it?

doc/metadata-server/README.md
palas, line 62: "full meaning of a particular instance of a contract" is a bit esoteric. I think it would be better to explain that we need to look at role distribution too in order to make sure that it was created correctly, because if, for example, if the creator of an escrow contract keeps the tokens for all parties it is not really an escrow contract.

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+- In multiparty agreement it can be not so important who is the owner of the initial transaction (contract participant, some third party auditor, layer etc.). On the other hand it is rather important that the attachment for a transaction was really dedicated for its semantic meaning.
+
+- Using these schema we can require distribution of role tokens to a specific, generic utility scripts with constant addresses (like "deposit scripts") and provide the same audited Marlowe contract with the same role distribution which can be safely reused and certified. By using this schema we can push a possible role distribution logic outside of our Marlowe interpreter / validator (we don't have to change our implementation) to role scripts (with single and simpler to audit responsiblity) and distribute reusable instances of Marlowe contracts together with verified set of metadata (authorized for this Contract and hard coded role to scripts distribution).
+
+## Extended Marlowe
```
palas, line 106: In the overview we could say why we are considering contracts in Marlowe and Extended Marlowe separately. Also can we generalise this so that it would be possible to use other kinds of templates (like a Haskell function, for example)?

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+### Terminology
```
palas, line 108: Again would leave terminology to the end of the document

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+          Nothing
+
+  -- | For the `InsMetadata` we reuse a similar procedure
+  -- | as in previous chapter
+  insMetadata <- extractInstanceMetadata
+
+  pure { insMetadata, extMetadata }
+```
+
+## Protocol
```
palas, line 177: This sounds like it could be more context if it was at the beginning, before the schema. But it could use a bit more introduction/context. What does the solution look like?

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+### Objectives
+
+- We want to avoid centralization of metadata registry.
+
+- We want to minimize blockchain resource usage in the default setup.
+
+- We want to use [existing or emerging Cardano standards](https://cips.cardano.org/) and tools.
+
+- Metadata format schema should be well defined but we should leave it open / extensible.
```
palas, line 187: Could this be folded together with the ones in the motivation? Also they would be more powerful if they had justification

palas (changes requested):
It is clear you put a lot of thought in the decisions and that you fleshed out the details a lot. But it is really hard for me to understand the thought process by reading the document. It kind of forces the reader to guess why the things are done the way they are. You explain some of the reasons, but they are not linked in any way to the solutions, and there is very little context.

I've written some suggestions about how I think it would be possible to improve the structure. I wrote some examples but the content in the examples may be wrong, is just about the way I would approach explaining it.

I think the tip that works best for me is just trying to think of the document as trying to explain things to someone using a whiteboard instead of trying to be very formal about it. I am not saying we don't need detail, but we shouldn't start with it

doc/metadata-server/README.md
paluh: I would like to leave the terminology here and explain the difference using it below.

doc/metadata-server/README.md
```diff
@@ -0,0 +1,327 @@
...
+    && checkValidator txValidator
+    && checkSignature txMetadata.payload pubKey MarloweTxMetadata.signature
+      then
+        Just MarloweTxMetadata.metadata
+      else
+        Nothing
+```
+
+> In the current implementation `MarloweParams` are used to derive the final validator. `MarloweParams` provide "low level" details (the currency symbol) which are used to implement roles handling on chain. From the perspective of the contract semantics they are not important. We treat them as implementation detail here hidden behind `checkValidator` but in the final implementation spec we should also handle them somehow.
+> It is worth to mention that in the context of [CIP-33](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0033) it seems beneficial to move `MarloweParams` to the `datum` so the validator script could be shared and the transaction size could be minimized.
```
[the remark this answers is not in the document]
paluh: @bwbush Wouldn't something like [`utxoRefsWithCurrency`](https://github.com/input-output-hk/plutus-apps/blob/main/plutus-contract/src/Plutus/Contract/Request.hs#L354) be enough to track all the pieces connected to our "contract currency" which remains efficient?

paluh:
@bwbush @palas I'm not sure what to do with this PR and spec. In the `Motivation` section I clearly describe `Marketplace` and `Run` as a specific drivers for this spec. I can probably extend it to more a generic (than Marlowe extended) "compilation to Marlowe" flow.

Do you think that I can close this PR here and start a MIPs discussion or open up a draft of a PR there?

PR #110, opened 2022-04-01
Enforce single Marlowe input and output

nau:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/marlowe/src/Language/Marlowe/Scripts.hs
+++ b/marlowe/src/Language/Marlowe/Scripts.hs
@@ -20,8 +20,8 @@
 {-# LANGUAGE TypeOperators         #-}
 {-# OPTIONS_GHC -Wno-name-shadowing #-}
 {-# OPTIONS_GHC -fno-ignore-interface-pragmas #-}
-{-# OPTIONS_GHC -fno-omit-interface-pragmas #-}
-{-# OPTIONS_GHC -fno-specialise #-}
+-- {-# OPTIONS_GHC -fno-omit-interface-pragmas #-}
+
 
 module Language.Marlowe.Scripts where
 import GHC.Generics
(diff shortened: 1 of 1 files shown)


marlowe/src/Language/Marlowe/Scripts.hs
```diff
@@ -161,6 +153,41 @@ smallMarloweValidator MarloweParams{rolesCurrency, rolePayoutValidatorHash}
...
+    findOwnInput :: ScriptContext -> Maybe TxInInfo
+    findOwnInput ScriptContext{scriptContextTxInfo=TxInfo{txInfoInputs}, scriptContextPurpose=Spending txOutRef} =
+        find (\TxInInfo{txInInfoOutRef} -> txInInfoOutRef == txOutRef) txInfoInputs
+    findOwnInput _ = Nothing
+
+    ownInput :: TxInInfo
+    ownInput@TxInInfo{txInInfoResolved=TxOut{txOutAddress=ownAddress}} =
+        case findOwnInput ctx of
+            Just ownTxInInfo ->
+                case filter (sameValidatorHash ownTxInInfo) (txInfoInputs scriptContextTxInfo) of
```
bwbush, line 165: What about the situation where two scripts have the same `rolePayoutValidatorHash`, but different `rolesCurrency`s? (Such a situation could occur because the validator does not enforce the relationship between these two parts of `MarloweParams`.) Then the script addresses would be different.

For example, could someone who possessed two corresponding role tokens for both contracts (i.e., the same token name but different currency symbols) submit a double-satisfaction transaction? It seems like both validations would succeed because each script's call to `validatePartyWitness` would find that the role had properly authorized. Both scripts would validate the double-satisfaction output, where `payoutToTxOut` verify that it paid to the common role address.

Would a `sameRoleValidatorHash` be safer than `sameValidatorHash`? Forcing payments to different role validator addresses would seem to completely prevent double satisfaction, since single output can be to more than one address.

bwbush: I agree that it's ugly and awkward; it also might increase execution costs. I'm fine with staying with `sameValidatorHash`. If future PlutusTx builtins make it efficient, we can revisit this.

In general, I think we're going to need an off-chain linting (or security) tool for Marlowe contracts and transactions: it would flag a `MarloweParams` as being non-standard/suspicious (e.g., having a broken relationship between the payout validator and the currency symbol). It can also warn of double-satisfaction dangers: I realized that the Marlowe validator can protect against double-satisfaction for multiple Marlowe in the same transaction, but it cannot protect against double-satisfaction with non-Marlowe scripts in the same transaction because those other scripts would need their own double-satisfaction protection.

bwbush (approved):
I just ran the basic on-chain tests on this branch, and they all succeeded. :1st_place_medal: Thanks!

PR #116, opened 2022-04-14
SCP-3792 restart marloweapp

jhbertra:
Changes:

- Kill and restart MarloweApp after 3 ~minute~ block timeout (blocks are counted by the number of `SlotChange` messages we get from the PAB, ignoring any smaller slots than previously seen, e.g. when the PAB hits a rollback).
- Kill and restart MarloweApp on startup if it has no available endpoints

### DIFF
--- a/marlowe-dashboard-client/src/Capability/PlutusApps/MarloweApp.purs
+++ b/marlowe-dashboard-client/src/Capability/PlutusApps/MarloweApp.purs
@@ -30,7 +30,7 @@ import Data.Tuple.Nested (type (/\), (/\))
 import Data.UUID.Argonaut (UUID)
 import Effect.Aff (Aff, Error, forkAff, joinFiber)
 import Effect.Aff.Class (class MonadAff, liftAff)
-import Env (_applyInputBus, _createBus, _redeemBus)
+import Env (_applyInputBus, _createBus, _marloweAppTimeout, _redeemBus)
 import Language.Marlowe.Client (MarloweError)
 import Marlowe.PAB (PlutusAppId)
 import Marlowe.Semantics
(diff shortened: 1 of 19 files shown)


marlowe-dashboard-client/src/Capability/PlutusApps/MarloweApp.purs
```diff
@@ -51,20 +57,36 @@ class MarloweApp m where
...
+    -> m
+         ( AjaxResponse
+             (UUID /\ Aff (Maybe (Either MarloweError MarloweParams)))
+         )
   applyInputs
     :: PlutusAppId
     -> MarloweParams
     -> TransactionInput
-    -> m (AjaxResponse (Aff (Either MarloweError Unit)))
+    -> m (AjaxResponse (Aff (Maybe (Either MarloweError Unit))))
```
hrajchert, line 68: This is a weird one, why Maybe Either? What is the semantic meaning of this? Can't we create a new data type that wraps MarloweError and whatever this Maybe should represent?

hrajchert (approved):
Just a comment in one file. Haven't tested functionality

marlowe-dashboard-client/src/Capability/PlutusApps/MarloweApp.purs
jhbertra: Yeah, that's probably a good idea at this point... I didn't want to incur the overhead of copying all the instance boilerplate to a newtype...

The meaning is:

```
Nothing: The request timed out
Just (Left err): The request completed with an error
Just (Right a): The request completed successfully
```

jhbertra: I'll address this on my local branch, I'm editing these files there as well

PR #117, opened 2022-04-15
[Draft] Universal Marlowe Script: combine Validator and MintingPolicy in a single script

nau:
PoC of an idea to combine validator and mintingPolicy in a single script.

It's possible to produce a script that can be run in both Validation and Minting contexts.
In such a case we can mint Marlowe role tokens and ensure we created a Marlowe contract that uses these tokens within the same transaction. Thus, we can guarantee that no one can reuse the same tokens for different Marlowe contracts produced this way.

The thing is that tokens' `CurrencySymbol == ValidatorHash`, and we can check during minting that the Tx contains a script TxOut with `ValidatorHash == CurrencySymbol`.

### DIFF
--- a/marlowe/marlowe.cabal
+++ b/marlowe/marlowe.cabal
@@ -179,6 +179,7 @@ test-suite marlowe-test
         template-haskell -any,
         streaming -any,
         plutus-pab -any,
+        plutus-core,
         async -any,
         prettyprinter -any,
         purescript-bridge -any,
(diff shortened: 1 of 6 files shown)


bwbush:
@palas, the SCP-4107 was formatted as a Marlowe Improvement Proposal (MIP): https://github.com/input-output-hk/MIPs/pull/1.

bwbush:
`Language.Marlowe.Scripts` from this PR has been cherry-picked, modified, and merged into https://github.com/input-output-hk/marlowe-cardano/tree/SCP-4107, which is derived from https://github.com/input-output-hk/marlowe-cardano/tree/sprint-61.

PR #137, opened 2022-05-11
Try -j1 for darwin builds

hamishmack:
This workaround is for an error that started showing up when building with macOS 12:
   `cannot dlopen until fork() handlers have completed`


michalrus:
Thank you! :stars: :stars: :stars:

hamishmack:
We switched to ghc 8.10.7 and this was not needed.

PR #141, opened 2022-05-12
SCP-3964 - Make CurrencyInput into a Halogen component

palas:
This fixes an issue that was preventing Kraken oracle from working in playground


hrajchert (approved):
You could also remove the Hooks library as it is no longer used (or create a ticket to do that in another PR)

palas:
> You could also remove the Hooks library as it is no longer used (or create a ticket to do that in another PR)

It seems it is still used by the MetadataTab: https://hydra.iohk.io/build/14566163/nixlog/1/tail 🤷‍♂️
but it looks that is the last

PR #146, opened 2022-05-17
SCP-3912 add blockly datetime picker

hrajchert:
This PR adds a DateTime picker to Blockly to be able to select constant times

Can be previewed at https://marlowe-playground-hernan.plutus.aws.iohkdev.io/

blockly_date_time

Pre-submit checklist:
- Branch
- PR
    - [X] Self-reviewed the diff
    - [X] Reviewer requested

### DIFF
--- a/marlowe-playground-client/src/Blockly/DateTimeField.js
+++ b/marlowe-playground-client/src/Blockly/DateTimeField.js
@@ -1,0 +1,109 @@
+// Register a custom DateTime Field to blockly:
+// Used as base:
+// * https://developers.google.com/blockly/guides/create-custom-blocks/fields/customizing-fields/creating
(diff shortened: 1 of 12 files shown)


ladamesny (approved):
Thanks @hrajchert the code looks great. I appreciate all the notes. I simulated two simple contracts and tested the timeout with both a datetime picker field and a time param field, and it timed-out as expected.

palas (approved):
That is very cool, I didn't know you could tune Blockly this much.
Two thoughts, I am not sure whether they are possible but two things I noticed:
- The clear button gives "dd/mm/yy --:--", it would be nicer if it gave the epoch (0) or the current time
- We are not displaying seconds, but they are significant, so maybe we should add them even if it makes the block a bit larger

PR #149, opened 2022-05-19
SCP-3995 Report tip with slot change, and only when tip changes.

bwbush:
The PAB now reports slot changes as
```
{"tag":"SlotChange","tip":{"getSlot":3001420},"current":{"getSlot":1521393}}
```
and only does so when the tip advances.

@hrajchert, please adjust Marlowe Run, which is broken by this change in message format and content. Thanks!


jhbertra (approved):
LGTM

PR #169, opened 2022-06-21
scp 4141 fix warning crash

hrajchert:
This PR fixes an issue in the Blockly editor, where a warning that pointed to a Timeout crashed.
The problem was that we were using the `number` rule that doesn't have Position information instead of using the `%number` token that does.

The fix can be checked in [this deploy](https://marlowe-playground-hernan.plutus.aws.iohkdev.io/)
Pre-submit checklist:
- Branch
- PR
    - [X] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/marlowe-playground-client/grammar.ne
+++ b/marlowe-playground-client/grammar.ne
@@ -90,7 +90,7 @@ number
 
 timeout
    -> hole {% ([hole]) => hole %}
-    | number {% ([n]) => opts.mkTerm(opts.mkExtendedTimeValue(n))({startLineNumber: n.line, startColumn: n.col, endLineNumber: n.line, endColumn: n.col + n.toString(10).length}) %}
+    | %number {% ([n]) => opts.mkTerm(opts.mkExtendedTimeValue(n.value))({startLineNumber: n.line, startColumn: n.col, endLineNumber: n.line, endColumn: n.col + n.toString(10).length}) %}
     | lparen "TimeParam" someWS string rparen {% ([start,{line,col},,s,end]) => opts.mkTerm(opts.mkExtendedTimeParam(s))({startLineNumber: start.line, startColumn: start.col, endLineNumber: end.line, endColumn: end.col}) %}
 
 string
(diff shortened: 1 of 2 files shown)


marlowe-playground-client/grammar.ne
```diff
@@ -90,7 +90,7 @@ number
 
 timeout
    -> hole {% ([hole]) => hole %}
-    | number {% ([n]) => opts.mkTerm(opts.mkExtendedTimeValue(n))({startLineNumber: n.line, startColumn: n.col, endLineNumber: n.line, endColumn: n.col + n.toString(10).length}) %}
```
palas, line 93: Out of curiosity. What does adding a "%" before the identifier do?

palas (approved):
Great. Thank you :)

marlowe-playground-client/grammar.ne
hrajchert: As this comment suggest

> The problem was that we were using the number rule that doesn't have Position information instead of using the %number token that does.

If we use the plain `number` then the grammar rule `number` is used, which you can see on line 89

```
number
   -> %number {% ([n]) => opts.mkBigInteger(n.value.replace('_', '')) %}
    | lparen %number rparen {% ([,n,]) => opts.mkBigInteger(n.value.replace('_', '')) %}
```

That rule states that a number is either the token `%number` directly or the token enclosed in parenthesis. The token number is provided by the lexer and defined by a regular expression in line 6

```
const lexer = moo.compile({
        WS: /[ \t]+/,
        number: /-?[0-9](?:_*[0-9])*/,
        base16: {match: /"(?:[0-9a-fA-F][0-9a-fA-F])*"/, value: x => x.slice(1, -1)},
``` 

Nearly allows you to define different types of lexer, the default one gives you the index number of the token, the one we use (moo) gives you the start line, char, end line, char of the token, which was what was missing from the When rule. 

So in line 93 when we did `startLineNumber: n.line`, it resulted in the `undefined` that ended up exploding in the warning tab.

PR #170, opened 2022-06-22
SCP-4012 Create images and compose file for marlowe-run + marlowe-playground

shlevy:


jhbertra:
In principle, I am for this change with the understanding that after #197 is merged into `main`, `marlowe-run` and `marlowe-playgound` will be removed from the repo, and so these docker images will no longer be deployed. If the purpose of this is to deploy docker images of the current (alonzo era) builds of these tools, then that's fine, but much of this work will either not be maintained (in the case of run) or have to be migrated over to `marlowe-playground`.

@shlevy can you confirm what the short and long-term plans are for these docker images and the docker-compose file? Aside from this, the PR will also likely serve as an instructive template when it comes to building docker images and a compose file for the runtime.

shlevy:
@jhbertra The original goal for this was to give the pioneers local environments for our components. Moving forward, I think it would still be a good idea to have images and compose files for the playground (as a single deployment, without run), and images and compose files for the runtime as we develop that. Does that answer your question?

jhbertra:
@shlevy once the PR is up to date with `main`, this is good to merge

jhbertra (approved):

PR #171, opened 2022-06-27
SCP-4145

paluh:
Save / load project bundles from playground. This functionality requires larger project refactoring - we have to extract auth flows and restructure `Save` / `Open` flows together with project data model.
That is why this PR won't be small.


bwbush:
@paluh, this branch is based on `sprint-59`, which is long outdated. Could it base branch be switched to `main` instead?

paluh:
@bwbush Yes but this rebase requires a bit of work from my side. I'm going to close this PR and move this task to the new `playground` repo. Thanks for looking into that.

PR #172, opened 2022-06-27
scp-4128 Add version to the module of marlowe Core and Extended

hrajchert:
This PR is the first step into getting Marlowe Core and Extended versioned. The initial refactor involves adding the version as part of the module name to both Extended and Core Marlowe. I've unified the PureScript and Haskell namespace and I separated the structure so that Extended and Core can be patched individually.

The main changes are:

* Haskell
  * `Language.Marlowe.Semantics` -> `Language.Marlowe.Core.V1.Semantics`
  * `Language.Marlowe.Semantics.Types` -> `Language.Marlowe.Core.V1.Semantics.Types`
  * `Language.Marlowe.Extended` -> `Language.Marlowe.Extended.V1`
* PureScript 
  * Split types and instances of `Marlowe.Semantics` -> `Language.Marlowe.Core.V1.Semantics.Types`
  * Split evaluation functions of `Marlowe.Semantics` -> `Language.Marlowe.Core.V1.Semantics`
  * `Marlowe.Extended` -> `Language.Marlowe.Extended.V1`
  * `Marlowe.Extended.Metadata` -> `Language.Marlowe.Extended.V1.Metadata`
  * Moved Argonaut helpers to `purescript-web-common`

Still to decide for future work:
  * Haskell
    *  Pretty, Client, Util and Analysis are not versioned, is that fine? For adding version to pretty, should we add the Grammar first?
    * The plain `marlowe/src/Language/Marlowe.hs` export should be rethinked
  * PureScript (web-common-marlowe)
    * We should remove the `Marlowe/Deinstantiate.purs` hack once we have metadata sharing mechanism.
    * Marlowe/<HasParties, Template, Time> are not versioned for the moment. Should we?

The next step for me will involve adding a top-level construct for both representations that will serialize to JSON and include the version number. I think I will call it `MarloweModule` even if for the foreseeable future we don't import nor export stuff, as Marlowe `Contract` is already used for the continuation.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


bwbush (commented):
1. I think the Haskell module and name changes are fine.
2. I wonder if someday we might want to split Extended Marlowe into a separate package.
3. `Pretty`, `Client`, `Util` do not need versioning at this point and should be coded to to work with any version.
4. `Language.Marlowe.%` should probably be moved to `Util`.
5. `Language.Marlowe` should probably be versioned, maybe called `Language.Marlowe.Core.V1.Api`?

palas (changes requested):
It makes sense. We only need the types to be versioned, but it is good to have the semantics too because then we can have some tests for the conversion functions that we are going to need. Nice. It feels weird to have a file named `V1.hs` though :)

The only thing I see missing is this file, not sure what it is but probably something about deployment: `nix/tests/vm-tests/runghc-api-request.json`
the instructions in the tutorials: `doc/marlowe/tutorials/using-marlowe.rst`
the cookbook: `marlowe-cli/cookbook/english-auction.ipynb`
and these test: `marlowe/test/Spec/PAB/Workflow.hs`

hrajchert:
> It makes sense. We only need the types to be versioned, but it is good to have the semantics too because then we can have some tests for the conversion functions that we are going to need. Nice. It feels weird to have a file named `V1.hs` though :)
> 
> The only thing I see missing is this file, not sure what it is but probably something about deployment: `nix/tests/vm-tests/runghc-api-request.json` the instructions in the tutorials: `doc/marlowe/tutorials/using-marlowe.rst` the cookbook: `marlowe-cli/cookbook/english-auction.ipynb` and these test: `marlowe/test/Spec/PAB/Workflow.hs`

Yes, I didn't like the `V1.hs` file either, but sadly neither Haskell or PureScript have an `index.js` analogue to give the name to the directory. I'm happy to rename if we find a better name.

I'll upload the changes with the missing changes.

palas (approved):
The renaming looks good now. I don't understand the changes to the `marlowe-cli`, so probably @bwbush should review those

yveshauser (commented):
Agree, we could think of having a proper module `marlowe-extended` (which would probably allow to have shorter namespaces, as there is no need to introduce `Core`). Re-naming looks good.

paluh (approved):

yveshauser (approved):

PR #177, opened 2022-07-04
SCP-4067 - Make static analysis consider unreachable `Close` contracts

palas:
Modify reachability static analysis so that it also flags when a `Close` contract is unreachable (not just non-trivial contracts)
Deployed to: https://marlowe-playground-pablo.plutus.aws.iohkdev.io/


hrajchert (approved):
LGTM

Issue #174, opened 2022-06-29
Report warnings during `auto` contract execution


edsko:
The `auto` endpoint provides "automatic" contract execution for Marlow contracts that are amenable to this. A key function in this process is

```haskell
getAction :: MarloweTimeRange -> Party -> MarloweData -> PartyAction
```

which computes what action should happen next. Like regular contract execution, this depends on `reduceContractUntilQuiescent`, but unlike regular execution, it [silently swallows any `ReduceWarning`s](https://github.com/input-output-hk/marlowe-cardano/blob/294c0a06e961ed83887e63a8ad135fcb1425c179/marlowe/src/Language/Marlowe/Client.hs#L966) . This can be pretty confusing, because the perspective of the Marlow contract author, those "warnings" can be pretty serious; for example, if the contract stipulates that a payment should happen but there are insufficient funds, the payment is simply skipped. The execution log in this case will merely record the next action in the transaction, with zero indication that anything was skipped.

Perhaps it would make sense to bubble up these warnings from `getAction` and then log them as warnings (or even errors) in `autoExecuteContract` (in `marlowePlutusContract`)? (Happy to submit a PR.)

bwbush:
@edsko, thank you for pointing this out. I agree that it is important for the warnings to be reported. We'd be happy to review a PR if you'd like to submit one; if not, we will queue this for the Marlowe team to address.

edsko:
Thanks @bwbush . I have opened #181; it's only small.

bwbush:
The PR request for this was accepted.

PR #181, opened 2022-07-07
Report warnings during auto contract execution

edsko:
This provides a simple PR that addresses #174. I have not added tests that verify the warnings.

### DIFF
--- a/marlowe/src/Language/Marlowe/Client.hs
+++ b/marlowe/src/Language/Marlowe/Client.hs
@@ -827,7 +827,8 @@ marlowePlutusContract = selectList [create, apply, applyNonmerkleized, auto, red
     autoExecuteContract reqId params typedValidator party marloweData = do
         time <- currentTime
         let timeRange = (time, time + defaultTxValidationRange)
-        let action = getAction timeRange party marloweData
+        let (warnings, action) = getAction timeRange party marloweData
+        forM_ warnings $ \w -> logWarn $ "Warning: " <> show w
         case action of
             PayDeposit acc p token amount -> do
                 logInfo $ "PayDeposit " <> show amount <> " at within time " <> show timeRange
@@ -959,11 +960,11 @@ shelleyAddressToKeys (AddressInEra _ (Shelley.ShelleyAddress _ paymentCredential
             pure (ppkh,  Just . StakePubKeyHash . PubKeyHash . toBuiltin $ serialiseToRawBytes stakeHash)
 shelleyAddressToKeys _ = throwError $ OtherContractError $ Contract.OtherContractError "Byron Addresses not supported"
 
-getAction :: MarloweTimeRange -> Party -> MarloweData -> PartyAction
+getAction :: MarloweTimeRange -> Party -> MarloweData -> ([TransactionWarning], PartyAction)
 getAction timeRange party MarloweData{marloweContract,marloweState} = let
     env = Environment timeRange
     in case reduceContractUntilQuiescent env marloweState marloweContract of
-        ContractQuiescent _reduced _warnings _payments state contract ->
+        ContractQuiescent _reduced warnings _payments state contract -> (convertReduceWarnings warnings,
             -- here the contract is either When or Close
             case contract of
                 When [Case (Deposit acc depositParty tok value) _] _ _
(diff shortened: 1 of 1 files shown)


yveshauser (commented):
Thanks! Logging the warnings for sure makes sense

yveshauser (changes requested):
`getAction` only tries to detect what the next action for the auto execution should be - the warnings should rather be logged after `computeTransaction`

edsko:
> getAction only tries to detect what the next action for the auto execution should be - the warnings should rather be logged after computeTransaction

I don't fully understand. `getAction` is a pure function, it cannot do any logging, it merely computes what warnings there are alongside the action. The logging is done in `autoExecuteContract`, where there is other logging too. Not sure what you'd like to change, exactly?

yveshauser:
> > getAction only tries to detect what the next action for the auto execution should be - the warnings should rather be logged after computeTransaction
> 
> I don't fully understand. `getAction` is a pure function, it cannot do any logging, it merely computes what warnings there are alongside the action. The logging is done in `autoExecuteContract`, where there is other logging too. Not sure what you'd like to change, exactly?

The warnings are available after the call to `computeTransaction` in the `mkStep` function. The `TransactionOutput` has a field `txOutWarnings` that contains the warnings that should be logged. `getAction` only tries to determine the possible next actions but does not trigger any payments.

edsko:
Yes, but prior to this PR, any warnings encountered by `getAction` will be irrevocably lost (the `warnings` field of `ContractQueiest` was ignored). There can indeed be _additional_ warnings, depending on the action that is computed by `getAction`, but any warnings arising from computing what action that should be (and hence also independent of that action, because they arise primarily from the actions _before_ the action returned by `getAction` to be executed. The warnings that are returned as part of the output by `computeTransaction` are a different set.

For example, if a contract has two transfers, but the contract does not have sufficient  funds for the first transfer, prior to this PR, that first transfer is simply skipped over with no logging anywhere, and `computeTransaction` will not even get to see it.

But perhaps I am still misunderstanding you?

yveshauser:
> Yes, but prior to this PR, any warnings encountered by `getAction` will be irrevocably lost (the `warnings` field of `ContractQueiest` was ignored). There can indeed be _additional_ warnings, depending on the action that is computed by `getAction`, but any warnings arising from computing what action that should be (and hence also independent of that action, because they arise primarily from the actions _before_ the action returned by `getAction` to be executed). The warnings that are returned as part of the output by `computeTransaction` are a different set.
> 
> For example, if a contract has two transfers, but the contract does not have sufficient funds for the first transfer, prior to this PR, that first transfer is simply skipped over with no logging anywhere, and `computeTransaction` will not even get to see it.
> 
> But perhaps I am still misunderstanding you?

Yes, the logging of the warnings is missing. In order to add it, I would propose to put it as follows:
```
diff --git a/marlowe/src/Language/Marlowe/Client.hs b/marlowe/src/Language/Marlowe/Client.hs
index 307b5a10b..3c4592177 100644
--- a/marlowe/src/Language/Marlowe/Client.hs
+++ b/marlowe/src/Language/Marlowe/Client.hs
@@ -1259,7 +1259,8 @@ mkStep MarloweParams{..} typedValidator timeInterval@(minTime, maxTime) clientIn
                 txInputs = inputs }

         case computeTransaction txInput marloweState marloweContract of
-            TransactionOutput {txOutPayments, txOutState, txOutContract} -> do
+            TransactionOutput {txOutWarnings, txOutPayments, txOutState, txOutContract} -> do
+                forM_ txOutWarnings $ \w -> logWarn $ "Warning: " <> show w
                 let marloweData = MarloweData {
                         marloweContract = txOutContract,
                         marloweState = txOutState }
```
As in `computeTransaction` the inputs get applied and the contract is reduced until quiescent. Only here a contract gets actually reduced - `getAction` only tries to detect inputs for the auto-execution and does not reduce the contract. Do you agree, or am I missing something?

edsko:
I agree that it would be useful to log those warnings _as well_, but those are _different_ warnings than the warnings logged by this PR. The _combination_ of what you're doing and what I'm proposing would probably make the most sense: this PR as it should logged warnings that arose during `getAction`, and then the code you're adding will log any warnings that arise from the execution of the action returned by `getAction`.

edsko:
Specifically, `reduceContractUntilQuiescent` can skip actions if they cannot be executed, for example due to lack of funds (see `reduceContractStep`, case for `Pay`). Such actions will result in warnings being returned from `reduceContractUntilQuiescent`; these are the warnings `getAction` bubbles up in my PR, along with the action that is to be executed _after_ the contract was reduced until quiescent.

bwbush:
Merged as https://github.com/input-output-hk/marlowe-cardano/pull/207.

edsko:
Thanks @bwbush ! :)

PR #182, opened 2022-07-08
Make Marlowe chain agnostic

edsko:
In the literature, Marlowe is described as being agnostic to a particular choice of blockchain, be it UTxO based, account based, or otherwise. This is _mostly_ true for the implementation, but not quite. There are a few places where the Marlowe semantics refer to Cardano specific concepts; specifically, the Cardano-specific notion of a `Token`, and the Cardano-specific notion of identity (`PubKeyHash`). I am in the process of modifying the implementation so it abstracts over these concepts; of course, the validation function and off-chain could would still be specific to Cardano, but the core types and the core semantics would not be. This PR is work in progress; I'm not sure if this generalization would be welcomed by the Marlowe team. If not, of course feel free to say so and I will close this PR again. But if it is, then I will update this PR when I'm finished and make it ready for review.

### DIFF
--- a/marlowe-actus/src/Language/Marlowe/ACTUS/Domain/BusinessEvents.hs
+++ b/marlowe-actus/src/Language/Marlowe/ACTUS/Domain/BusinessEvents.hs
@@ -6,7 +6,7 @@ module Language.Marlowe.ACTUS.Domain.BusinessEvents where
 
 import Data.Aeson.Types (FromJSON, ToJSON)
 import GHC.Generics (Generic)
-import Language.Marlowe (Token, Value)
+import Language.Marlowe (PubKeyHash, Token, Value)
 
 {-| ACTUS event types
     https://github.com/actusfrf/actus-dictionary/blob/master/actus-dictionary-event.json
@@ -54,4 +54,4 @@ data RiskFactorsPoly a = RiskFactorsPoly
     deriving anyclass (FromJSON, ToJSON)
 
 type RiskFactors = RiskFactorsPoly Double
-type RiskFactorsMarlowe = RiskFactorsPoly (Value Token)
+type RiskFactorsMarlowe = RiskFactorsPoly (Value PubKeyHash Token)
--- a/marlowe-actus/src/Language/Marlowe/ACTUS/Domain/BusinessEvents.hs
+++ b/marlowe-actus/src/Language/Marlowe/ACTUS/Domain/BusinessEvents.hs
@@ -6,7 +6,7 @@ module Language.Marlowe.ACTUS.Domain.BusinessEvents where
 
 import Data.Aeson.Types (FromJSON, ToJSON)
 import GHC.Generics (Generic)
-import Language.Marlowe (Observation, Value)
+import Language.Marlowe (Observation, Token, Value)
 
 {-| ACTUS event types
     https://github.com/actusfrf/actus-dictionary/blob/master/actus-dictionary-event.json
@@ -54,4 +54,4 @@ data RiskFactorsPoly a = RiskFactorsPoly
     deriving anyclass (FromJSON, ToJSON)
 
 type RiskFactors = RiskFactorsPoly Double
-type RiskFactorsMarlowe = RiskFactorsPoly (Value Observation)
+type RiskFactorsMarlowe = RiskFactorsPoly (Value (Observation Token) Token)
--- a/marlowe-actus/src/Language/Marlowe/ACTUS/Domain/BusinessEvents.hs
+++ b/marlowe-actus/src/Language/Marlowe/ACTUS/Domain/BusinessEvents.hs
@@ -6,7 +6,7 @@ module Language.Marlowe.ACTUS.Domain.BusinessEvents where
 
 import Data.Aeson.Types (FromJSON, ToJSON)
 import GHC.Generics (Generic)
-import Language.Marlowe (Observation, Token, Value)
+import Language.Marlowe (Token, Value)
 
 {-| ACTUS event types
     https://github.com/actusfrf/actus-dictionary/blob/master/actus-dictionary-event.json
@@ -54,4 +54,4 @@ data RiskFactorsPoly a = RiskFactorsPoly
     deriving anyclass (FromJSON, ToJSON)
 
 type RiskFactors = RiskFactorsPoly Double
-type RiskFactorsMarlowe = RiskFactorsPoly (Value (Observation Token) Token)
+type RiskFactorsMarlowe = RiskFactorsPoly (Value Token)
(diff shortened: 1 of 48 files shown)


edsko:
Ok, this PR is now in principle complete, although I need to do a second generalization too (identity). I've updated all dependencies of `marlowe` also, although I've not attempted to _generalize_ most of them (i.e., most of them refer to `Token` specifically). We may wish to introduce some type synonyms for `Contract Token` and co (or indeed rename `Contract` and then let `type Contract = Contract_ Token` or whatever). Wasn't sure what the most convenient would be, so haven't done that yet. Happy to address that if there a specific request from you guys. 

(I am aware that this generalization might not be what you want at all... don't mean to be presumptuous, just keeping this PR up to date.)

edsko:
Ok, introduced type synonym for `Value` (and `Case`). The old type is now `Value_`, and the new `Value` is defined as
    
```haskell
type Value t = Value_ (Observation t) t
```
    
This avoids the awkward duplication that was happening everywhere, which I'd argue is useful even without the gneralization of this PR. It doesn't solve the type synonym problem for the non-generalized version, of course.

edsko:
Ok, I have now also generalized away from identity (and fixed the test failure that CI was reporting, which was due to Plutus `Map`s not really being maps at all..).

Note that I have _not_ generalized the "Extended Marlowe" language, which I think is what most users actually see? So my hope is that the impact of this generalization downstream is relatively small.

edsko:
I'm seeing the that the tests are failing due to the validator being too large. Not sure what's causing this, I didn't change any code, merely signatures. Conjecture: some dictionaries will need to be passed in, could that push it over the edge? Anyway, will leave this until I know that this is something the Marlowe team even thinks is a good idea or not.

bwbush:
> I'm seeing the that the tests are failing due to the validator being too large.

@edsko, changing the signatures likely caused the validator size to increase by about 3kB. We'll need to look into mitigating this. BTW, the Marlowe private testnet has more generous protocol parameters, so we could do on-chain testing there.

@palas, @hrajchert, @yveshauser, the change proposed in this PR is reminiscent of my Lean4 implementation, which aggressively parameterized the Marlowe types. I'm wondering if in the long term the Isabelle and Haskell in https://github.com/input-output-hk/marlowe should take this approach, but perhaps (due to validator size etc.) the Haskell in https://github.com/input-output-hk/marlowe-cardano might need to be bound to the concrete types for Cardano?

edsko:
> changing the signatures likely caused the validator size to increase by about 3kB. We'll need to look into mitigating this. BTW, the Marlowe private testnet has more generous protocol parameters, so we could do on-chain testing there.

@bwbush This is rather surprising though, no? After all, Plutus core is untyped, making more types polymorphic should not make a significant difference. My only guess so far is that there are some dictionaries being passed around that are resulting in more size? (Total guess, no evidence for it.) Or perhaps the Plutus compiler just does something slightly different when code is polymorphic? Do you have an intuition as to why this is making such a big difference?

I can take a closer look myself if you guys think that the generalization in this PR is useful.

(I actually want to generalize the on-chain client too, so that it can have different kinds of rules for different kinds of tokens; and I'd like to make the core contract language extensible. But those generalizations are a bit more localized, and I figured the one in this PR at least matches the spirit of Marlowe, as it makes the language chain independent.)

edsko:
I did some initial investigation, measuring the size of the validator, wrapper, and params separately. Out of curiosity, I also measured the size of the Plutus code after running it through the [Plutonomy optimizer for Plutus](https://github.com/well-typed/plutonomy). The results for `main` are

```
## Typed

scriptSize: 14677

validator: 10876 (optimized  8832)
wrapper  :  4373 (optimized  3319)
params   :    50 (optimized    39)
total    : 15299 (optimized 12190, 0.80%)

## Untyped

scriptSize: 12582

# edsko/parameterize-contract

## Typed

scriptSize: 17748 (increase of 3071)

validator: 13473 (optimized 11086)
wrapper  : 4789  (optimized  3703)
params   : 50    (optimized    39)
total    : 18312 (optimized 14828, 0.81%)

## Untyped

scriptSize: 15204 (increase of 2622)
```

(Here `scriptSize` is the measurement as it is done in the existing size tests; this doesn't _quite_ line up with measuring the components, I guess some optimization is happening when applying them.)

Two observations at this point:

* The size of the "untyped" validator is `15204` after this PR, which is only _just_ over the limit set by the test (15200). The limit is exceeded by quite a bit more for the typed validator, but I think that's mostly because the limit for the untyped validator was less strict. Initially I thought that the difference in how much the limit is exceeded was giving a clue about the cause of the problem, but I now think this is a red herring.
* Plutonomy can reduce the Marlowe code by about 20% (in its default configuration), which isn't _immediately_ relevant for this PR but worth keeping in mind.

I don't yet know exactly where the growth is coming from; I will try and take a look at that next.

hrajchert:
I think the repository that needs to be blockchain agnostic is https://github.com/input-output-hk/marlowe. We are in the process of updating the Marlowe specification and proofs, as they are a little bit outdated (doesn't have Extended Marlowe, still uses Slot instead of time and doesn't have merkelization). As part of the update we are planning to do a `test oracle` that test that a particular implementation (for example marlowe-cardano) is compliant with the blockchain agnostic semantics.

edsko:
I'm taking a closer look at where some of this extra size comes from; indeed, delay/force arguments arising from type arguments is one source, as you already point out, but I've identified at least one other (https://github.com/input-output-hk/plutus/issues/4782). I will continue looking at this a bit longer.

edsko:
Ok, I looked more at this, and wrote a blog post with some of my findings; the final section is of particular relevance I guess: https://well-typed.com/blog/2022/08/plutus-cores/#the-cost-of-polymorphism .

FWIW, I tried moving the calls to `compile` into a separate module from the definition of `rolePayoutScript`, `smallTypedValidator`, `smallUntypedValidator`; this will give the `ghc` optimizer a chance to look at the code before the Plutus compiler plugin does. However, as I already anticipated in the blog post, this actually made things worse; 18318/16286 (two separate scripts) instead of 17748/15204, so roughly increase of 3%/7% respectively. Also tried adding `-fspecialize-aggressively`, which seemed to have zero effect; also tried adding `-O1` in addition, which made it worse still 18675/16660. This is somewhat disappointing, because I thought part of the problem of my generalization of Marlowe was type class dictionaries being passed around, and in small scale tests `-fspecialize-aggressively` does manage to get rid of that. But I guess in the context of the larger script, it also has other consequences elsewhere. 

I will come back to this again, but might be a little while before I do.

edsko:
Ok, opened a PR against `marlowe` at https://github.com/input-output-hk/marlowe/pull/135 that attempts the same generalization there. We are still experimenting, but it's probably much easier there, and if it works for us, then I will close this PR for now. (I was halfway rebasing this PR, when I realized that due to some changes the `marlowe` repo might now work for us).

edsko:
Ok, closing this in favour of https://github.com/input-output-hk/marlowe/pull/135 .

PR #185, opened 2022-07-14
SCP-4247 Filtered Chain Sync Protocol specification

jhbertra:
Adds draft spec for filtered chain sync protocol to be implemented by Marlowe Runtime Chain Index.

### DIFF
--- a/marlowe-runtime/docs/filtered-chain-sync-spec-1.0.md
+++ b/marlowe-runtime/docs/filtered-chain-sync-spec-1.0.md
@@ -17,7 +17,7 @@ only needs to process a tiny fraction of the full blockchain.
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
-interpreted as described in RFC 2119.
+interpreted as described in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
 
 An implementation is not compliant if it fails to satisfy one or more of the
 MUST or REQUIRED level requirements for the protocols it implements. An
@@ -40,13 +40,13 @@ objects of, the protocol.
 - query: Part of the payload of a query next message which the client sends to the server to tell it what to send next.
 - error: A protocol level error which is triggered when one party sends an invalid message. Results in the termination of the connection.
 - rejection: An application level error which is triggered when the server rejects a syntactically valid, semantically invalid query sent by the client. Does not result in the termination of the connection.
-- roll forward: A message sent by the server to the client in response to a query.
+- roll forward: A message sent by the server to the client to advance its position forward to a new block. One possible response to a query.
 - result: Part of the payload of a roll forward message which contains the result of a query.
-- roll backward: A messagge sent by the server to the client when the client's position has been rolled back.
-- tip: The latest chronological point in the blockchain
-- point: A point in the blockchain up to and including the tip
+- roll backward: A messagge sent by the server to the client to revert its position to a previous block. One possible response to a query.
+- tip: The latest chronological point in the blockchain that the connected node has recorded.
+- point: A point in the blockchain up to and including the tip. The precise representation is implementation-specific.
 - position: The point where a client is currently "focused."
-- schema version: A named and numberedversion of the query schema supported by the parties of the protocol. Used in the handshake to verify that the server supports the query schema supported by the client.
+- schema version: The version of the query schema supported by the participants in the protocol. Used in the handshake to verify that the server supports the query schema supported by the client.
 - handshake: A proceedure used to establish a connection which ensures the client and server understand the same queeries.
 - agency: Used to describe which role is permited to send a message in a given state.
 
@@ -69,6 +69,18 @@ result of the query. If the response is a roll backward message, the payload
 will include the new position of the client, and the new tip of the blockchain.
 The position will be the most recent common ancestor point of the client's
 previous position and the new tip.
+
+### A Note on Intersections
+
+The original Chain Sync Protocol provides a way to fast forward the client to
+an "intersection point" - that is, a point which intersects with the set of
+blocks the client may have already visited during a previous session. This
+mechanism allows clients to connect without having to start from the genesis
+block each time.
+
+The Filtered Chain Sync Protocol does not provide such a mechanism, because the
+query mechanism provides a much more flexible way to acheive the same result,
+making intersection points redundant.
 
 ## State Machine
 
@@ -106,7 +118,7 @@ its position to `Genesis` when it receives a `ConfirmHandshake` message.
 The `Idle` state is the state from which a client MAY send a query to the
 server. It does so by sending a `QueryNext` message. The payload of the
 `QueryNext` MUST include the position of the client, and a query. The
-`QueryNext` message transitions the state machine to the `Next CanWait` state.
+`QueryNext` message transitions the state machine to the `Next CanAwait` state.
 The client MAY request the server terminates the connection by sending a `Done`
 message. The `Done` message transitions the state machine to the `Done` state.
 
@@ -133,9 +145,9 @@ tip.
   - `RollForward` to client
   - `RollBackward` to client
 
-##### Next CanWait state
+##### Next CanAwait state
 
-The `Next CanWait` state is a sub-state of `Next wait` in which the server MUST
+The `Next CanAwait` state is a sub-state of `Next wait` in which the server MUST
 decode and decide how to handle the query receved in the `QueryNext` message.
 If the server cannot decode the query, it MUST reject the query with a
 `QueryRejected` message. The payload of the message SHOULD indicate that the
@@ -197,8 +209,8 @@ The `RequestHandshake` is sent by the client to initiate the handshake.
 The `RequestHandshake` message is encoded as follows, in the following order:
 
 - A `tag` octet consisting of the value `0x01`
-- 4 octets representing a 4-byte integer length that describes the length of
-  the following string.
+- 4 octets containing the length of the following string as a big-endian 32-bit
+  integer.
 - A UTF-8 encoded string of octets containing the schema version.
 
 #### ConfirmHandshake message
@@ -240,7 +252,7 @@ The `RejectHandshake` message is encoded as follows, in the following order:
 The `QueryNext` is sent by the client to request more data from the blockchain.
 
 - Origin State: `Idle`
-- Target State: `Next CanWait`
+- Target State: `Next CanAwait`
 - Payload: current position, query
 - Sender: client
 - Receiver: server
@@ -260,7 +272,7 @@ The `QueryNext` message is encoded as follows, in the following order:
 The `QueryRejected` is sent by the server to indicate that the query has been
 rejected.
 
-- Origin State: `Next CanWait`
+- Origin State: `Next CanAwait`
 - Target State: `Idle`
 - Payload: rejection reason
 - Sender: server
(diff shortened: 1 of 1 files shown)


marlowe-runtime/docs/filtered-chain-sync-spec-1.0.md
```diff
@@ -0,0 +1,375 @@
...
+specified query, the client is able to dramatically reduce the volume of data
+it must process compared to running the Chain Sync Protocol. This is
+particularly useful for writing the off-chain code for a smart contract, which
+only needs to process a tiny fraction of the full blockchain.
+
+### Requirements
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in RFC 2119.
```
bwbush, line 20: add hyperlink

marlowe-runtime/docs/filtered-chain-sync-spec-1.0.md
```diff
@@ -0,0 +1,375 @@
...
+- state: A member of the set of states specified by the state machine of the protocol.
+- message: A member of the set of state transitions specified by the state maching of the protocol.
+- payload: Data associated and sent with a message.
+- query: Part of the payload of a query next message which the client sends to the server to tell it what to send next.
+- error: A protocol level error which is triggered when one party sends an invalid message. Results in the termination of the connection.
+- rejection: An application level error which is triggered when the server rejects a syntactically valid, semantically invalid query sent by the client. Does not result in the termination of the connection.
+- roll forward: A message sent by the server to the client in response to a query.
+- result: Part of the payload of a roll forward message which contains the result of a query.
+- roll backward: A messagge sent by the server to the client when the client's position has been rolled back.
+- tip: The latest chronological point in the blockchain
```
bwbush, line 46: Technical, at the node that one is connected to, which may or may not be the latest chronological point globally.

marlowe-runtime/docs/filtered-chain-sync-spec-1.0.md
```diff
@@ -0,0 +1,375 @@
...
+- point: A point in the blockchain up to and including the tip
+- position: The point where a client is currently "focused."
+- schema version: A named and numberedversion of the query schema supported by the parties of the protocol. Used in the handshake to verify that the server supports the query schema supported by the client.
```
bwbush, line 49: typo

marlowe-runtime/docs/filtered-chain-sync-spec-1.0.md
```diff
@@ -0,0 +1,375 @@
...
+- message: A member of the set of state transitions specified by the state maching of the protocol.
+- payload: Data associated and sent with a message.
+- query: Part of the payload of a query next message which the client sends to the server to tell it what to send next.
+- error: A protocol level error which is triggered when one party sends an invalid message. Results in the termination of the connection.
+- rejection: An application level error which is triggered when the server rejects a syntactically valid, semantically invalid query sent by the client. Does not result in the termination of the connection.
+- roll forward: A message sent by the server to the client in response to a query.
+- result: Part of the payload of a roll forward message which contains the result of a query.
+- roll backward: A messagge sent by the server to the client when the client's position has been rolled back.
+- tip: The latest chronological point in the blockchain
+- point: A point in the blockchain up to and including the tip
```
bwbush, line 47: Can we commit to a precise definition of "point" in this document? `BlockHash` is essential along with `BlockNo` and/or `SlotNo`.

marlowe-runtime/docs/filtered-chain-sync-spec-1.0.md
```diff
@@ -0,0 +1,375 @@
...
+### Operation
+
+The Filtered Chain Sync Protocol is a stateful client-server protocol. When a
+client establishes a connection with a server, a handshake is performed. If the
+handshake succeeds, the idle state is reached. From the idle state, the client
+has the agency to send a query or terminate the connection. If the client
+terminates the connection, by sending a `Done` message, then the server closes
+the connection and the protocol terminates. If the client sends a query, by
+sending a `QueryNext` message, the server can choose to respond immediately to
+the query, reject the query with a rejection reason, or ask the client to wait
```
bwbush, line 62: Do we really need `Wait`? It just tells the client to do wait it would be doing anyway. Does it have information content? Surely the underlying transport layer would ensure message receipt, keep-alives/pings, etc.

marlowe-runtime/docs/filtered-chain-sync-spec-1.0.md
```diff
@@ -0,0 +1,375 @@
...
+for a response. When the client receives a response or a rejection, it returns
+to the idle state, and the loop repeats.
+
+A response can be a roll forward or roll backward message. If the response is a
```
bwbush, line 66: Does the state machine need more failure transitions? For instance, what if the server accepts the query and transitions to the `MustWait` state, but there is a server-side failure (e.g., lost connect to the node or database)? It appears that the state machine has already committed to a roll forward or roll back that it might not be able to service. Should there be an option to report a failure, either retryable or not?

marlowe-runtime/docs/filtered-chain-sync-spec-1.0.md
```diff
@@ -0,0 +1,375 @@
...
+roll forward, the payload will include the new position of the client and the
+result of the query. If the response is a roll backward message, the payload
+will include the new position of the client, and the new tip of the blockchain.
+The position will be the most recent common ancestor point of the client's
+previous position and the new tip.
+
```
bwbush, line 72: Also, what happens when the server starts to shut down. Does it notify clients, or just drop the connnections?

marlowe-runtime/docs/filtered-chain-sync-spec-1.0.md
```diff
@@ -0,0 +1,375 @@
...
+- Payload: Schema version
+- Sender: client
+- Receiver: server
+
+##### Encoding
+
+The `RequestHandshake` message is encoded as follows, in the following order:
+
+- A `tag` octet consisting of the value `0x01`
+- 4 octets representing a 4-byte integer length that describes the length of
```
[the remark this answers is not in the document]
jhbertra: Big endian

marlowe-runtime/docs/filtered-chain-sync-spec-1.0.md
jhbertra: This is from the original Chain Sync protocol. The server sends a `Wait` message to the client when it reaches the tip, which lets the client know it may take an arbitrary amount of time to receive a response. I've always thought this was a somewhat odd design decision myself, but I decided to preserve it.

jhbertra: I think the idea is to give the client a chance to run an action when it is informed it must wait for a reply.

marlowe-runtime/docs/filtered-chain-sync-spec-1.0.md
jhbertra: In principle, the server could lose and regain its connection to the node while in a `Next any` state. If it can reconnect and resume from where it was, things proceed as if no interruption occurred. If it fails to find its old position, the client may receive a `RollBackward` message. Either way, this event is opaque to the client. I don't think the protocol should represent cases like these; we risk a leaky abstraction if it does.

In the case of a database connection error, the server could either try to reconnect or (more likely) die. If the server dies, it will server the connection at the transport layer. The client may try to re-establish the lost connection as it sees fit.

marlowe-runtime/docs/filtered-chain-sync-spec-1.0.md
jhbertra: Hmm, good question. I think it needs to sever the transport-layer connection. The problem is that the server cannot notify clients in the `Idle` state because the client has agency (i.e. it will not be listening for messages on its end of the channel). It could, in theory, send `Closed` messages to all clients waiting for a response, but that doesn't cover all clients. So I think the appropriate action is to terminate the connection at the transport layer.

marlowe-runtime/docs/filtered-chain-sync-spec-1.0.md
bwbush: Okay, maybe we should state that we Marlowe's point is identical to Node's point?

jhbertra:
> My only substantial concern is about the necessity of the `Wait` state.

I assume you meant the `Wait` message. I explained why I included it in the original review comment thread. I could certainly be convinced that it is unnecessary. It would simplify the state machine and the API a bit, at the cost of making it less expressive.

marlowe-runtime/docs/filtered-chain-sync-spec-1.0.md
bwbush: Thanks for reminding me. Yes, let's keep this, both for consistency and for signaling the client that they might experience a long wait.

marlowe-runtime/docs/filtered-chain-sync-spec-1.0.md
bwbush: Sounds good.

marlowe-runtime/docs/filtered-chain-sync-spec-1.0.md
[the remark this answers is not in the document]
bwbush: I just realized that we're committing to a binary protocol (good), but not commenting on serialization format. Should the specification eventually specify types, serialization etc.? or is that a separate document?

bwbush:
> > My only substantial concern is about the necessity of the `Wait` state.
> 
> I assume you meant the `Wait` message. I explained why I included it in the original review comment thread. I could certainly be convinced that it is unnecessary. It would simplify the state machine and the API a bit, at the cost of making it less expressive.

I'm convinced we should keep `Wait`.

marlowe-runtime/docs/filtered-chain-sync-spec-1.0.md
jhbertra: I can make this more explicit here that the precise representation is implementation-specific.

bwbush:
@jhbertra, should the specification mention that the intersection-point functionality will be available via a query? Without such a statement, one might interpret that clients always have to work their way forward from genesis.

jhbertra:
> @jhbertra, should the specification mention that the intersection-point functionality will be available via a query? Without such a statement, one might interpret that clients always have to work their way forward from genesis.

That is technically an accurate interpretation :slightly_smiling_face:. However, I will include this note because it significantly deviates from the original protocol, which we reference as the basis for this one.

marlowe-runtime/docs/filtered-chain-sync-spec-1.0.md
```diff
@@ -0,0 +1,375 @@
...
+- client: The role played by the participant which sends queries and receives chain data.
+- server: The role played by the participant which responds to queries maintains chain data.
+- state: A member of the set of states specified by the state machine of the protocol.
+- message: A member of the set of state transitions specified by the state maching of the protocol.
+- payload: Data associated and sent with a message.
+- query: Part of the payload of a query next message which the client sends to the server to tell it what to send next.
+- error: A protocol level error which is triggered when one party sends an invalid message. Results in the termination of the connection.
+- rejection: An application level error which is triggered when the server rejects a syntactically valid, semantically invalid query sent by the client. Does not result in the termination of the connection.
+- roll forward: A message sent by the server to the client in response to a query.
+- result: Part of the payload of a roll forward message which contains the result of a query.
```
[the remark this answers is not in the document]
jhbertra: Done

marlowe-runtime/docs/filtered-chain-sync-spec-1.0.md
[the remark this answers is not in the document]
jhbertra: Can you help me understand the difference between the two? From my perspective, this specification is explicit about the serialization format of the messages.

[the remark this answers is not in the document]
bwbush: Sorry, I was reading the raw markdown, not the rendered markdown, and didn't see the heading levels correctly. I think the specification is indeed explicit enough.

bwbush (approved):

PR #187, opened 2022-07-18
SCP-4262 Connect Marlowe Chain Index to Cardano Node

jhbertra:
This PR adds a pipelined client for the Chain Sync Protocol. Currently, the handling logic is a stub that logs some stats to stdout.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


marlowe-chain-sync/app/Main.hs
```diff
@@ -0,0 +1,86 @@
+module Main where
```
bwbush, line 1: Eventually, we should probably move `main :: IO ()` to the CLI component for Marlowe runtime.

marlowe-chain-sync/app/Options.hs
```diff
@@ -0,0 +1,80 @@
...
+  , getOptions
+  ) where
+
+import Cardano.Api (NetworkId (..), NetworkMagic (..))
+import Data.Maybe (fromMaybe)
+import qualified Options.Applicative as O
+import System.Environment (lookupEnv)
+import Text.Read (readMaybe)
+
+data Options = Options
```
bwbush, line 12: Eventually move to CLI component.

bwbush (approved):
Looks good, but I haven't compiled an run it against a network.

Let's keep configuration, command-line options, and (perhaps) file i/o separate so we can refactor to have other components/infrastructure handle those.

marlowe-chain-sync/app/Options.hs
jhbertra: _Maybe_ yes. See my other comment for more details.

marlowe-chain-sync/app/Main.hs
jhbertra: I understand that no action required on this for now, but this highlights the need for us to discuss and get on the same page regarding the plan for the CLI, as this doesn't align with what I had imagined. In my mind, the CLi does not control the lifecycle of runtime processes, it simply is an interface that connects to an already running runtime via IPC to issue commands and queries.

Not saying one way or another that I agree or disagree, merely drawing attention to the fact that there is currently a lack of alignment here.

PR #200, opened 2022-07-26
SCP-4263 Persist Chain Data

jhbertra:
This PR adds:

- A convenient Docker-based development configuration for PostgreSQL
- Database migrations via Sqitch https://sqitch.org/docs/manual/sqitchtutorial/
- Bulk block persistence to PostgreSQL via Hasql https://hackage.haskell.org/package/hasql


marlowe-chain-sync/app/Main.hs
```diff
@@ -1,86 +1,58 @@
...
-            writeTVar ttip tip
-      where
-        connectionInfo = LocalNodeConnectInfo
-          { localConsensusModeParams = CardanoModeParams $ EpochSlots 21600
-          , localNodeNetworkId = networkId
-          , localNodeSocketPath = nodeSocket
-          }
-        getHeaderAtPoint _ = pure Origin
-        getIntersectionPoints _ _ = pure []
+    localNodeConnectInfo :: LocalNodeConnectInfo CardanoMode
```
bwbush, line 50: _General comment (FYI only):_ Since we may run this eventually with Hydra, where the node connection has a different type, we may want to keep the node-specific functions as separate as possible from the pure computations that use them.

marlowe-chain-sync/deploy/asset.sql
```diff
@@ -0,0 +1,42 @@
...
+  ( id SERIAL
+  , policyId BYTEA NOT NULL
+  , name BYTEA NOT NULL
+  , CONSTRAINT pk_asset PRIMARY KEY (id)
+  , CONSTRAINT unique_asset UNIQUE (policyId, name)
+  );
+
+CREATE TABLE chain.assetOut
+  ( txOutId BYTEA NOT NULL
+  , txOutIx BIGINT NOT NULL
```
bwbush, line 16: Technically, we don't need `BIGINT`, even though this maps to `Integer`, because it is impossible for `txOutIx` to ever be greater than about 10k.

marlowe-chain-sync/deploy/block.sql
```diff
@@ -0,0 +1,19 @@
...
+BEGIN;
+
+SET client_min_messages = 'warning';
+
+CREATE TABLE chain.block
+  ( id BYTEA NOT NULL
+  , slotNo BIGINT NOT NULL
+  , blockNo BIGINT NOT NULL
+  , rollbackToBlock BYTEA
+  , rollbackToSlot BIGINT
```
bwbush, line 13: Are we sure we don't need `rollbackToBlockNo`, too? It's probably redundant, but we should think about edge cases.

marlowe-chain-sync/deploy/asset.sql
jhbertra: Good to know!

marlowe-chain-sync/deploy/asset.sql
```diff
@@ -0,0 +1,42 @@
...
+  , slotNo BIGINT NOT NULL
+  , assetId INT NOT NULL
+  , quantity BIGINT NOT NULL
+  , CONSTRAINT fk_assetMint_asset FOREIGN KEY(assetId) REFERENCES chain.asset(id)
+      ON DELETE RESTRICT
+  );
+
+COMMIT;
+
+CREATE INDEX assetMint_slotNo ON chain.assetMint USING BTREE (slotNo);
```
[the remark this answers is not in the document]
jhbertra: It doesn't really matter (I like being explicit). Generally, Btree is a sensible default unless you have a strong reason to prefer another.

marlowe-chain-sync/deploy/initial_partition.sql
```diff
@@ -0,0 +1,27 @@
...
+  END LOOP;
+END;
+$$;
+
+SELECT chain.createInitialPartitions('block');
+SELECT chain.createInitialPartitions('tx');
+SELECT chain.createInitialPartitions('txOut');
+SELECT chain.createInitialPartitions('txIn');
+
+DROP FUNCTION chain.createInitialPartitions;
```
bwbush, line 25: FYI:, `CREATE FUNCTION pg_temp.` is also possible.

marlowe-chain-sync/deploy/block.sql
jhbertra: `SlotNo` and `BlockHeaderHash` are sufficient to uniquely identify a block (and this is what we build the indexes on). Technically, `BlockHeaderHash` is sufficient, but having the slot no allows us to make use of the composite index and improve query performance further.

marlowe-chain-sync/deploy/initial_partition.sql
jhbertra: Thanks for the tip

marlowe-chain-sync/docs/marlowe-chain-sync.1
```diff
@@ -22,8 +22,8 @@ chain using the Filtered Chain Sync Protocol.
```
bwbush, line 22: Historically, we haven't created `man` pages for Marlowe tools. Do we know whether these will work properly on Windows and Mac? @ladamesny might want to test on Mac and @dino-iog on Windows.

marlowe-chain-sync/secrets/postgres_password
```diff
@@ -0,0 +1 @@
+9kZ@o7j4OkMC
```
bwbush, line 1: In the long term, we should probably avoid even default secrets in the commit.

marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
```diff
@@ -0,0 +1,564 @@
...
+ commitGenesisBlock
+ getHeaderAtPoint
+ getIntersectionPoints
+ getGenesisBlock
+
+-- GetGenesisBlock
+
+getGenesisBlock :: GetGenesisBlock Session
+getGenesisBlock = GetGenesisBlock $ statement () $ refineResult (decodeResults . V.toList)
+  [vectorStatement|
```
bwbush, line 59: :sunglasses:

marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
```diff
@@ -0,0 +1,564 @@
...
+          AND block.id     = $2 :: bytea
+      |]
+    where
+      decodeResults :: Maybe Int64 -> Either Text (WithOrigin BlockHeader)
+      decodeResults Nothing        = Left $ "No block found at " <> T.pack (show point)
+      decodeResults (Just blockNo) = Right $ At $ BlockHeader slotNo hash $ BlockNo $ fromIntegral blockNo
+
+decodeTxId :: ByteString -> Either Text TxId
+decodeTxId txId = case deserialiseFromRawBytes AsTxId txId of
+  Nothing    -> Left "Invalid TxId"
```
bwbush, line 102: Should we include the bytes that couldn't be decoded in the error message itself?

marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
```diff
@@ -0,0 +1,564 @@
...
+
+-- GetIntersectionPoints
+
+getIntersectionPoints :: GetIntersectionPoints Session
+getIntersectionPoints = GetIntersectionPoints $ statement () $ rmap decodeResults
+  [vectorStatement|
+    SELECT slotNo :: bigint, id :: bytea
+    FROM   chain.block
+    WHERE  rollbackToBlock IS NULL
+    ORDER BY slotNo DESC LIMIT 2160
```
bwbush, line 118: The Ouroboros security parameter is 2160 _blocks_, not 2160 _slots_, so this limit is not large enough. Also, I don't know that _k_ will ever change, but maybe we should flag this hardcoding so we can use a parameter someday?

marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
```diff
@@ -0,0 +1,564 @@
...
+-- CommitRollback
+
+commitRollback :: CommitRollback Session
+commitRollback = CommitRollback \case
+  ChainPointAtGenesis -> sql $ BS.intercalate "\n"
+    [ "BEGIN;"
+    , "TRUNCATE TABLE chain.tx;"
+    , "TRUNCATE TABLE chain.txOut;"
+    , "TRUNCATE TABLE chain.txIn;"
+    , "DELETE FROM chain.asset;"
```
bwbush, line 136: Why `DELETE` instead of `TRUNCATE`?

marlowe-chain-sync/secrets/postgres_password
jhbertra: Yes, though this isn't sensitive info. Given that, it would probably be better if the `postgres` user for the dev database not to have a password at all.

marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
jhbertra: We could

marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
jhbertra: Yeah, I want to parameterize this, but do not know the most appropriate place to read it from. And this does limit by 2160 blocks, not slots, as it is a `LIMIT`on the `block` table.

marlowe-chain-sync/deploy/tx.sql
```diff
@@ -0,0 +1,51 @@
+-- Deploy chain:tx to pg
+-- requires: block
+
+BEGIN;
+
+SET client_min_messages = 'warning';
+
```
bwbush, line 7: Is the lack of primary key, foreign key, and check constraints here (and throughout the schema) for optimization?

marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
```diff
@@ -0,0 +1,564 @@
...
+    decodeResult :: (Int64, ByteString) -> ChainPoint
+    decodeResult (slotNo, hash) = ChainPoint (SlotNo $ fromIntegral slotNo) (HeaderHash $ toShort hash)
+
+-- CommitRollback
+
+commitRollback :: CommitRollback Session
+commitRollback = CommitRollback \case
+  ChainPointAtGenesis -> sql $ BS.intercalate "\n"
+    [ "BEGIN;"
+    , "TRUNCATE TABLE chain.tx;"
```
bwbush, line 133: Even though we don't have foreign key constraints, I wonder if would be better to truncate the most dependent tables first, just in case we ever to do add foreign keys (and forget to reorder this code)?

marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
```diff
@@ -0,0 +1,564 @@
...
+    , "TRUNCATE TABLE chain.block;"
+    , "COMMIT;"
+    ]
+  ChainPoint slotNo hash -> statement (slotNoToParam slotNo, headerHashToParam hash)
+    [resultlessStatement|
+      WITH blockUpdates AS
+        ( UPDATE chain.block
+             SET rollbackToSlot  = $1 :: bigint
+               , rollbackToBlock = $2 :: bytea
+           WHERE slotNo > $1 :: bigint
```
bwbush, line 148: Should this be `>=` or `>`? My memory is that the rollback messages report a new block that should replace the old one at the slot and where everything subsequent should be discarded. Does the rollback strategy replace the block at the rollback slot and delete the transactions there? Maybe I'm mistaken.

marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
```diff
@@ -0,0 +1,564 @@
...
+              ) AS assetInputs
+            ON CONFLICT (policyId, name) DO NOTHING
+            RETURNING id, policyId, name
+          )
+        , assetIds AS
+          ( SELECT * FROM newAssets
+            UNION
+            SELECT asset.*
+            FROM   assetOutInputs AS assetOut
+            JOIN   chain.asset    AS asset    ON  asset.policyId = assetOut.policyId
```
bwbush, line 352: More succinct with `USING` instead of `ON`.

bwbush (approved):
I read through the code and schemas. Sorry if some of my comments are pedantic.

My only concern is the handling of rollbacks, but I'd need to walk thorough the code meticulously to check that. Perhaps we should make a list of rollback scenarios and check the code against those.

marlowe-chain-sync/secrets/postgres_password
bwbush: I'm just concerned about the case where someone blindly deploys the chain sync without changing the default password.

marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
bwbush: Sorry, my mistake--I should have realized that when I read the query!

BTW, some of the testnets have a non-standard security parameter.

marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
bwbush: Right, the assets table is a table that has those.

jhbertra: You can use `TRUNCATE TABLE <table> CASCADE` for this actually.

I'm also pushing a fix that re-adds the genesis block after doing all this truncation... still faster to truncate and re-add than to exclude it using a `DELETE`

marlowe-chain-sync/deploy/tx.sql
jhbertra: It's a necessary tradeoff for acceptable insert performance. It is justified because it is a copy of a verified data source (the node already enforces constraints), so we can afford to be more relaxed by avoiding UNIQUE, PRIMARY KEY, and FOREIGN KEY constraints.

marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
jhbertra: If we do, we'll know because we'd need to update these to use `CASCADE` anyway (`TRUNCATE TABLE tbl` fails if `tbl` is referenced by a foreign key).

marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
jhbertra: I think you're mistaken - I'm pretty confident that the rollback `point` indicates the most recent block that is _not_ rolled back.

e.g. If the client has blocks `A B C D E F` and receives a `MsgRollBackward C`, then the client should delete everything after `C`, leaving `A B C`. The next `MsgRollForward` will return the next block `D'` after `C` in the new fork of the chain.

jhbertra: My evidence for this is that the chain sync server always starts by sending a `MsgRollback` to the client's starting block You will start by receiving a `MsgRollback ChainPointAtGenesis` if you do not start from an intersection point. If you receive a `MsgIntersectFound p` for some point `p` then send a `MsgRequestNext`, you will always receive a `MsgRollBackward p tip` before you receive a `MsgRollForward`.

Also, the `MsgRollBackward` only sends the `point`, not the block, so it doesn't tell you to "replace the old one at this point."

jhbertra:
> I read through the code and schemas. Sorry if some of my comments are pedantic.
> 
> My only concern is the handling of rollbacks, but I'd need to walk through the code meticulously to check that. Perhaps we should make a list of rollback scenarios and check the code against those.

Yes, absolutely we should do this (and write test cases for them ideally). What are your specific concerns?

PR #206, opened 2022-07-28
SCP-3128: Update plutus-core to fix broken haddock links.

shlevy:


shlevy:
Note that this is present in the current state so we can consider merging this and opening a separate ticket if desirable.

Ah yeah, the way this page works currently is we build a fake combined haddock index for a bunch of packages at once rather than keeping to the typical haddock structure of one haddock index per package. Options, in order of ease of implementation:

1. Keep things as-is
2. Remove the broken links
3. Simply serve separate haddocks, with some optional simple landing page pointing to packages of interest
4. Update the haddock combining script to have working combined indexes and contents

shlevy:
To be honest this docs page was part of what I inherited when joining the team and it's not clear to me that anyone is or should be using it, as opposed to just doing a local cabal haddock if you need it and waiting for the internal hackage work to proceed. But if there are real users we think are served by this then let's see what they need.

bwbush:
`Index` and `Contents` are really useful, especially if there is no `hoogle`.

bwbush:
> To be honest this docs page was part of what I inherited when joining the team and it's not clear to me that anyone is or should be using it, as opposed to just doing a local cabal haddock if you need it and waiting for the internal hackage work to proceed. But if there are real users we think are served by this then let's see what they need.

Yes, I had to create https://github.com/input-output-hk/marlowe-dependency-docs, so I'd have a usable hoogle+haddock. Maybe it's unrealistic to modify the `marlowe-cardano` deployment to do the same--we don't really know if or how much that haddock is used.

bwbush:
> 2. Remove the broken links

I vote for this option, so long as there is a goal deploy a hackage for Cardano.

@palas?

bwbush (approved):
Ideally, we should remove the broken "Contents" and "Index" links.

shlevy:
"status update" on the internal hackage work: https://input-output-rnd.slack.com/archives/CG386Q1NE/p1657034501259179?thread_ts=1656431670.008409&cid=CG386Q1NE

doc describing the plan: https://docs.google.com/document/d/1PGmrvcNmfLP7HhGZpcgEee6zlHsQ-hozCC2XkjzHcbI/edit#heading=h.fxrd36cq33g7

palas:
Could we add a symbolic link from `/doc/haddock/share/doc/` to `/doc` or something like that? it seems that `/doc/haddock/doc-index.html` works. If that is the only problem...

palas:
Otherwise it may be best to keep as is, cause I wouldn't have been able to find it without the broken link

shlevy:
@bwbush Fixed the broken links.

bwbush (approved):
My spot test of links worked. Thanks!

PR #218, opened 2022-08-10
Some work done to upgrade era from Alonzo to Babbage

dino-iog:
I'd like to merge this with cardano-node-1.35 before it diverges too far but don't want to break anyone's work. Here's what's happened so far:

Work has also been started here to get the non-PAB tests working on the
public networks. To facilitate this, changes have been made to
Language.Marlowe.CLI.Transaction buildFaucet (not buildFaucet') to allow
the use of any wallet addr/skey as a "faucet" to fund another address.

The `marlowe-cli util faucet` command now uses buildFaucet (not
buildFaucet' as it did before) and switches have been added for
specifying the source fund addr/skey.

This relates to (but does not yet close) SCP-4364 and SCP-4146

### DIFF
--- a/marlowe-cli/src/Language/Marlowe/CLI/Command/Util.hs
+++ b/marlowe-cli/src/Language/Marlowe/CLI/Command/Util.hs
@@ -33,7 +33,7 @@ import Language.Marlowe.CLI.Command.Parse (parseAddressAny, parseNetworkId, pars
                                            parseTokenName)
 import Language.Marlowe.CLI.Merkle (demerkleize, merkleize)
 import Language.Marlowe.CLI.Sync (watchMarlowe)
-import Language.Marlowe.CLI.Transaction (buildClean, buildFaucet', buildMinting, querySlotting, selectUtxos)
+import Language.Marlowe.CLI.Transaction (buildClean, buildFaucet, buildMinting, querySlotting, selectUtxos)
 import Language.Marlowe.CLI.Types (CliError, OutputQuery)
 import Plutus.V1.Ledger.Api (TokenName)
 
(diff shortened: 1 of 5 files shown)


paluh (approved):

ladamesny (approved):
Thanks @dino-iog looks good to me!

marlowe-cli/src/Language/Marlowe/CLI/Command/Util.hs
```diff
@@ -278,12 +281,14 @@ faucetCommand network socket =
...
+    <*> (O.optional . O.option O.auto)     (O.long "submit"          <> O.metavar "SECONDS"                              <> O.help "Also submit the transaction, and wait for confirmation."                                                         )
+    <*> O.option parseAddressAny           (O.long "faucet-address"  <> O.metavar "ADDRESS"                              <> O.help "The faucet addresses to provide funds."                                                                          )
+    <*> O.strOption                        (O.long "required-signer" <> O.metavar "SIGNING_FILE"                         <> O.help "File containing a required signing key."                                                                         )
+    <*> O.some (O.argument parseAddressAny $                            O.metavar "ADDRESS"                              <> O.help "The addresses to receive the funds."                                                                             )
```
jhbertra, line 291: Horizontal/tabulated alignment causes needlessly noisy diffs ;)

paluh: Yes, maybe we can enforce formatting tool to replace this approach? @bwbush do you have anything against changing the formatting across the project?

jhbertra: Normally I'm not that hung up about formatting choices - but horizontal alignment (though I know it's super common in Haskell) causes a specific problem in that it significantly increases the likelihood of merge conflicts that are not easy to resolve (diff and merge algorithms don't handle horizontal differences nearly as well as vertical ones)

dino-iog: I'm not a fan of this horiz alignment with semigroup things either but not sure what to do. The column got wider and it looked weird to not move them all. In my own stuff I usualy opt for using more vertical space which would make these files crazy long and perhaps harder to see the "big picture" of an individual command's switches on one screen height.
Maybe keep one-line but no columns, just one space between everything.

Also, I tend to not bother qualified importing the `Options.Applicative` in my code because the `O.` is very noisy. Basically just "giving up" on the explicit imports because most of these options parsing modules are using functions from that library anyway.

jhbertra: To be clear @dino-iog I'm not requesting a change here. I think if we wanted to take a position on this it should be uniformly applied across the codebase as @paluh suggested.

dino-iog:
These changes were merged with newer commits. Closing.

PR #221, opened 2022-08-10
SCP-4161 Draft of on-chain specification for Marlowe validators.

bwbush:
Please attend to the following items in this review:
- Is sufficient context provided?
- Is the notation clear?
- Are any on-chain constraints/requirements missing or incorrect?
- Are the constraints explained in sufficient detail?
- Is the specification's location the best folder for it?
- Should a [CIP-52](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0052/Tx-spec.md) section be added?

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


marlowe/specification/ReadMe.md
```diff
@@ -0,0 +1,346 @@
...
+        -- FIXME: We should write out `Value.leq` ourselves, so the specification does not depend upon Plutus's convenience functions.
+```
+
+
+#### *Constraint 15.* Sufficient payment
+
+Sufficient funds are paid to the public keys or roles.
+```haskell
+all sufficient (txOutPayments transactionOutput)
+  where sufficient (Payment _ (Party (PK hash)) value) = Value.leq value (foldMap (paidToHash hash) (txInfoOutputs $ scriptContextTxInfo scriptContext))
```
palas, line 296: Now I get why it is not obvious how `leq` behaves, it is the Map of Maps artefact which is tricky to handle because it is not just a number

palas: Also it seems that `foldMap` is working on the assumption that `Value` is a monoid that has addition as operation? Probably would be good to be explicit. Definitely this specification is helping me understand things better, nice :)

palas (changes requested):
Thanks, it looks really good, and it did help me understand the implementation much better :)
I still didn't understand the diagram 100%, but I think the legend addition could help, and also being familiar with Plutus datatypes probably helps too

marlowe/specification/ReadMe.md
```diff
@@ -0,0 +1,346 @@
...
+
+
+## Participants in Marlowe Contracts
+
+Marlowe contracts identify each participant by either a *public-key hash (PKH)* or a *role*. Marlowe semantics allow PKHs and roles to be freely mixed in contract logic.
+- A Marlowe PKH corresponds to the hash of a Cardano payment verification key, `Hash PaymentKey`.
+- A Marlowe role corresponds to a Cardano native token with the `TokenName`  for the role. (In Cardano, role names may be zero to thirty-two bytes long.) All of the role tokens in a given Marlowe contract must use the same  `CurrencySymbol` .
+
+
+## Life Cycle of a Marlowe Contract
```
simonjohnthompson, line 16: To be strictly accurate, the document should distinguish between a "contract" and a "contract instance" (or "an instance of a Marlowe contract"). Most of the discussion here is about the latter. The point is that the same contract can be executed multiple times, and indeed concurrently.

marlowe/specification/ReadMe.md
```diff
@@ -0,0 +1,346 @@
+# On-Chain Transaction Specification for Marlowe
+
+
+## Scope
+
+This document defines the specification for Marlowe semantics's interface with the Cardano blockchain. Marlowe utilizes three Plutus scripts: (i) a monetary policy for the roles currency used in some Marlowe contracts; (ii) the Marlowe application validator script that enforces Marlowe semantics; and (iii) the Marlowe payout-validator script that allows the holder of a role token to withdraw funds paid by the Marlowe application.
+
```
simonjohnthompson, line 7: Would be useful to have an overview of what the document contains. E.g. a list of sections.

marlowe/specification/ReadMe.md
```diff
@@ -0,0 +1,346 @@
...
+
+## Life Cycle of a Marlowe Contract
+
+The execution of a Marlowe contract proceeds as a linear sequence of applications of inputs at the contract's script address.
+1. Role tokens are typically minted prior to or within the creation transaction of the contract, though this is not enforced on-chain.
+2. The creation transaction for the contract stores the state of the contract in the datum at the contract's script address.
+3. Each application of input to the contract updates the state/datum at that same address.
+4. When a contract closes, there is no output to the contract's script address.
+5. If the contract pays funds to a role during application of inputs, those funds are sent to the address of the Marlowe payout-validator script.
```
simonjohnthompson, line 23: Would expect 4 (line 22) to follow 5 (line 23).

marlowe/specification/ReadMe.md
```diff
@@ -0,0 +1,346 @@
...
+
+Thus each Marlowe contract is a finite linear sequence of continuations at the script address, from creation to closure.
```
simonjohnthompson, line 25: no need for "linear"

marlowe/specification/ReadMe.md
```diff
@@ -0,0 +1,346 @@
...
+
+
+## Monetary Policy for Role Tokens
+
+Any Cardano monetary policy may be used to mint the role tokens used in a Marlowe contract. For security in standard Marlowe use cases, a one-time or locked minting policy such as `Plutus.Contracts.Currency.OneShotCurrency` is recommended. Exotic use cases might employ other monetary policies. It is the responsibility both of the developer of the off-chain code managing a contract and also of the user of the contract to verify that the monetary policy of the role tokens meets their security requirements.
```
simonjohnthompson, line 30: This could be tightened further. Would we want, by default, precisely one token per role?

There is a standard monetary policy in the implementation. Is there any reason that this is outside the scope of this document?

marlowe/specification/ReadMe.md
```diff
@@ -0,0 +1,346 @@
...
+
+
+## Representation of Marlowe Semantics in Plutus
+
+Marlowe semantics are embodied in the Haskell function `computeTransaction` that is generated from [the Isabelle implementation of the Marlowe DSL](https://github.com/input-output-hk/marlowe/tree/master/isabelle/Core). Note that `computeTransaction`  uses the `PlutusTx` types and functions instead of the standard-prelude or GHC-base Haskell ones.
```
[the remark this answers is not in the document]
simonjohnthompson: I was wondering about the "generated from" too.

marlowe/specification/ReadMe.md
```diff
@@ -0,0 +1,346 @@
...
+```haskell
+computeTransaction :: TransactionInput -> State -> Contract -> TransactionOutput
+```
+
+With the exception of `Input`, [the Isabelle types for Marlowe](https://github.com/input-output-hk/marlowe/blob/master/isabelle/Core/SemanticsTypes.thy) directly correspond to their Haskell representations, so below we only provide the definitions that will be later referenced in this document.
```
simonjohnthompson, line 40: Would it be better to have a stand-alone document that contains all the (Haskell) definitions needed, e.g. in an appendix, rather than having an external dependency like this?

marlowe/specification/ReadMe.md
```diff
@@ -0,0 +1,346 @@
...
+    TransactionOutput
+    {
+      txOutWarnings :: [TransactionWarning]
+    , txOutPayments :: [Payment]
+    , txOutState    :: State
+    , txOutContract :: Contract
+    }
+  | Error TransactionError
+```
+
```
simonjohnthompson, line 72: The treatment of Merkelization is confusing. I suggest either

1. A short overview about the Merkelization process and impact before it is discussed in the spec, or
2. Give the spec without Merkelization first, and then add a separate section which discusses the changes that it requires.

marlowe/specification/ReadMe.md
```diff
@@ -0,0 +1,346 @@
...
+Consider the application of the Marlowe validator and Marlowe semantics:
+```haskell
+validationResult = smallMarloweValidator marloweData marloweInput scriptContext
+
+transactionOutput = computeTransaction transactionInput inState inContract
+```
+
+Furthermore, let `marloweValidatorHash :: ValidatorHash` be the hash of the Marlowe semantics script.
+
+The validation fails (via returning `False` for `validationResult` or via the throwing of an error) if any of the following constraints does not hold.
```
simonjohnthompson, line 144: Should the two forms of failure be explained further? Under what circumstances can an error be thrown?

marlowe/specification/ReadMe.md
```diff
@@ -0,0 +1,346 @@
...
+filter (isMaroweAddress . txOutAddress . txInfoResolved) (txInfoInputs $ scriptContextTxInfo scriptContext) ≡ [TxInInfo _ (TxOut _ inValue _)]
+  where isMarloweAddress (Address (ScriptCredential hash)) = hash == marloweValidatorHash
+        isMarloweAddress _ = False
+```
+Hereafter we use `inValue` as the script value being spent.
+
+
+#### *Constraint 3.* Single Marlowe output
+
+There is a single output to the Marlowe script address unless the contract closes.
```
simonjohnthompson, line 172: in which case there is no output.

marlowe/specification/ReadMe.md
```diff
@@ -0,0 +1,346 @@
...
+        valueOfPayment (Payment _ (Party _) value)) = value
+        valueOfPayment _ = mempty
+```
+
+
+#### *Constraint 7.* Input state
+
+The input state of the transaction is used in the semantics computation.
+```haskell
+marloweState marloweData ≡ inState
```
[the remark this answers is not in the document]
simonjohnthompson: It's a constraint,  not a definition AFAIK

marlowe/specification/ReadMe.md
```diff
@@ -0,0 +1,346 @@
...
+        valueOfDeposit (IDeposit _ _ (Token currency name) count) = Value.singleton currency name count
+        valueOfDeposit _ = mempty
+        valueOfPayment (Payment _ (Party _) value)) = value
+        valueOfPayment _ = mempty
+```
+
+
+#### *Constraint 7.* Input state
+
+The input state of the transaction is used in the semantics computation.
```
simonjohnthompson, line 216: Needs rephrasing: doesn't seem to express the force of the constraint. Maybe "The input state of the transaction is the state used in the semantics computation."?

marlowe/specification/ReadMe.md
```diff
@@ -0,0 +1,346 @@
...
+```haskell
+marloweState marloweData ≡ inState
+```
+
+
+#### *Constraint 8.* Input contract
+
+The input contract of the transaction is used in the semantics computation.
```
simonjohnthompson, line 224: Similar comment to line 216

marlowe/specification/ReadMe.md
```diff
@@ -0,0 +1,346 @@
...
+
+The output contract of the transaction matches the result of the semantics computation.
+```haskell
+Just $ txOutContract transactionOutput ≡ fmap marloweContract . fromBuiltinData $ getDatum outDatum
+```
+
+
+#### *Constraint 12.* Merkleized continuations
+
+All of the merkleized contracts referenced in the input are provided to the script.
```
simonjohnthompson, line 256: provided by? how?

marlowe/specification/ReadMe.md
```diff
@@ -0,0 +1,346 @@
...
+
+All accounts in the state have positive balances.
+```haskell
+all ((> 0) . snd) (accounts $ marloweState marloweData) ≡ True
+```
+
+
+#### *Constraint 14.* Inputs authorized
+
+The required signature or at-least-one role token is present in the transaction for each deposit or choice.
```
simonjohnthompson, line 274: why hyphenated at-least-one here? Is it a technical term?

marlowe/specification/ReadMe.md
simonjohnthompson: Is this the tightest possible spec? Under what circumstances is the relation > rather than =? Would be worth including rationale for this

marlowe/specification/ReadMe.md
```diff
@@ -0,0 +1,346 @@
...
+```
+
+
+## Plutus Validator for Marlowe Payouts
+
+
+The Marlowe payout validator for roles is a Plutus script parameterized by currency symbol.
+```haskell
+rolePayoutValidator :: CurrencySymbol -> TokenName -> () -> ScriptContext -> Bool
+```
```
simonjohnthompson, line 314: Why the argument of type `()`?

simonjohnthompson (commented):
Thanks for this. I added some comments.

marlowe/specification/ReadMe.md
bwbush: Thanks for these comments. I'll make the following changes:
1. Specify the monoid laws that `Value` must conform to.
2. Specify the behavior required of `leq`. (I believe that `Value` comparison was the cause of the MinSwap vulnerability, so we need to be really clear about comparisons.)
3. Add the rationale for `>` instead of `==`. Briefly, `>` is required so that transactions conform to the min-ada requirement. For example, it is impossible to send less than one Ada or to send a token w/o any Ada.

marlowe/specification/ReadMe.md
bwbush: I'll edit throughout for clarify regarding contract vs contract instance.

marlowe/specification/ReadMe.md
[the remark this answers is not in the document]
bwbush: I'll also check with @hrajchert about this, since he is looking at Isabelle->Haskell code generation. There is a possibility that we could move from "translated" to "generated" soon.

marlowe/specification/ReadMe.md
```diff
@@ -0,0 +1,346 @@
...
+### Relationship between Marlowe Validator and Semantics
+
+The arguments are `computeTransaction` must be constructed as follows:
+1. The `txInterval` of `TransactionInput` is derived from the `txInfoValidTimeRange . scriptContextInfo` of the `ScriptContext`, as detailed below.
+2. The `txInputs` of `TransactionInput` is derived from the `MarloweInput` provided as the `Redeemer` and the `txInfoData . scriptContextTxInfo` of the `ScriptContext`, as detailed below.
+3. The `State` is the `marloweState` of the `MarloweData` provided as the `Datum`.
+4. The `Contract` is the `marloweContract` of the `MarloweData` provided as the `Datum`.
+5. The new `Datum` at the script address is the `MarloweData` with the same `marloweParams` as originally, but with the new `txOutState` and `txOutContract` of the `TransactionOutput`.
+
+![Relationship between Marlowe validator and semantics.](semantics2plutus.svg)
```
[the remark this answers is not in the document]
bwbush: Yes, I had to make too many compromises because I used graphviz to create the diagram. I'll redraw it using InkScape.

marlowe/specification/ReadMe.md
```diff
@@ -0,0 +1,346 @@
...
+fromBuiltinData redeemer       ≡ (Just markloweInput :: Maybe MarloweInput)
+fromBuiltinData scriptContext' ≡ (Just scriptContext :: Maybe ScriptContext)
+```
+
+
+#### *Constraint 2.* Single Marlowe script input
+
+The output of exactly one Marlowe UTxO is spent in the transaction.
+```haskell
+filter (isMaroweAddress . txOutAddress . txInfoResolved) (txInfoInputs $ scriptContextTxInfo scriptContext) ≡ [TxInInfo _ (TxOut _ inValue _)]
```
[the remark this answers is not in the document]
bwbush: It was typed from memory. I'm in the process of implementing each of the constraints as a pair of on- and off-chain tests. I'll probably find other typos in the code of the document and will fix all of them as I write implement the tests.

marlowe/specification/ReadMe.md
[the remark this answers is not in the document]
bwbush: @palas and @simonjohnthompson, I debated quite a bit about using `≡` vs `=` vs `==` vs $\stackrel{?}{=}$. What do you think would be the most appropriate notation here, given the limitations of unicode and markdown?

marlowe/specification/ReadMe.md
bwbush: Thanks, I'll expand this and add better links to the Isabelle specification.

marlowe/specification/ReadMe.md
bwbush: Fundamentally, all failures in Plutus are calls to `error` and all successes are returning `()`. The typed Plutus validators obscure this by using `Bool` as the return type.

I'll try to clean up the discussion in this section.

marlowe/specification/ReadMe.md
bwbush: The signature of the function requires a redeemer, but we don't need one, so we just use unit.

marlowe/specification/ReadMe.md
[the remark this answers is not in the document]
paluh: I think that the bottom `MarloweData` "constructor signature" should be modified to
`MarloweParams -> State -> Contract -> MarloweData` or should be changed to some kind of "product representation" (like `tuple` or `record`).

marlowe/specification/ReadMe.md
```diff
@@ -0,0 +1,346 @@
...
+    {
+      marloweParams   :: MarloweParams
+    , marloweState    :: State
+    , marloweContract :: Contract
+    }
+
+data MarloweParams =
+    MarloweParams
+    {
+      rolePayoutValidatorHash :: ValidatorHash
```
paluh, line 105: Do we want to consider hard coding the `rolePayoutValidatorHash` (it is going to be constant) value in the validator itself? We can save some space in the `Tx` and save some decoding/encoding CPU cycles.

marlowe/specification/ReadMe.md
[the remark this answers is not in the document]
paluh: We require `roleCurrency` as a part of `MarloweParams`. Do you think that we can relax this check by allowing many Marlowe validators with different currencies?

marlowe/specification/ReadMe.md
paluh: If we want to handle multiple Marlowe validators with distinct role currencies we can mark PK payouts using `datum` containing `roleCurrency` value so we can prevent double spending I think.

marlowe/specification/ReadMe.md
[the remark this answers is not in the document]
paluh: In the current context (I assume `Address (ScriptCredential marloweValidatorHash) == ownAddress`) isn't this approach vulnerable to double spending when we have PKH based payouts? If I understand this correctly we can use multiple Marlowe scripts because every one of them can have `ownAddress` and they pay to the same PKH and we are not able to distinguish which payout is dedicated for which script.

marlowe/specification/ReadMe.md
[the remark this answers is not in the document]
hrajchert: It is not generated, it just follows the Isabelle implementation. And after we implement the test oracle we can link to it to say how we make sure it follows it closely.

marlowe/specification/ReadMe.md
hrajchert: Currently the Isabelle specification doesn't include merkelization. We should probably add a section here that discusses it and how "it should not affect semantics", even if we don't provide assisted proofs.

marlowe/specification/ReadMe.md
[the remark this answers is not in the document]
bwbush: Thanks for the clarifications.

marlowe/specification/ReadMe.md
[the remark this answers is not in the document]
bwbush: @paluh, I'll add a `FIXME` in the document for the `roleCurrency`/`MarloweParams` issue and revisit it after the validator is frozen.

marlowe/specification/ReadMe.md
bwbush: @paluh, yes, I'll leave a `FIXME` in the document and re-sync it with the validator once that is frozen.

bwbush: @simonjohnthompson, I will tighten up the specification slightly by defining `leq` to require `==` for Ada but `<=` for non-Ada tokens. We cannot make this full equality because of the min-UTxO ledger constraint.

marlowe/specification/ReadMe.md
bwbush: I will add an appendix with a correspondence table between Isabelle and Haskell.

marlowe/specification/ReadMe.md
bwbush: Yes, I added a `FIXME` to address this after the validator is finalized.

marlowe/specification/ReadMe.md
bwbush: fixed in 8be04323e4ec24c4e923ed214094e85c916ab8e0

marlowe/specification/ReadMe.md
bwbush: fixed in 8be04323e4ec24c4e923ed214094e85c916ab8e0

marlowe/specification/ReadMe.md
bwbush: fixed in 8be04323e4ec24c4e923ed214094e85c916ab8e0

marlowe/specification/ReadMe.md
bwbush: fixed in 8be04323e4ec24c4e923ed214094e85c916ab8e0

marlowe/specification/ReadMe.md
bwbush: fixed in 8be04323e4ec24c4e923ed214094e85c916ab8e0

marlowe/specification/ReadMe.md
bwbush: fixed in 8be04323e4ec24c4e923ed214094e85c916ab8e0

marlowe/specification/ReadMe.md
bwbush: I added an early section on Merkleization in 8be04323e4ec24c4e923ed214094e85c916ab8e0.

marlowe/specification/ReadMe.md
bwbush: I added a `FIXME` so we can revisit this discussion/issue before finalizing.

marlowe/specification/ReadMe.md
bwbush: fixed in 8be04323e4ec24c4e923ed214094e85c916ab8e0

marlowe/specification/ReadMe.md
bwbush: done in 8be04323e4ec24c4e923ed214094e85c916ab8e0

marlowe/specification/ReadMe.md
bwbush: fixed in 8be04323e4ec24c4e923ed214094e85c916ab8e0

marlowe/specification/ReadMe.md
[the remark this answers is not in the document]
bwbush: fixed in 8be04323e4ec24c4e923ed214094e85c916ab8e0

marlowe/specification/ReadMe.md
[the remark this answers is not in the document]
bwbush: fixed in 8be04323e4ec24c4e923ed214094e85c916ab8e0

marlowe/specification/ReadMe.md
bwbush: fixed in 8be04323e4ec24c4e923ed214094e85c916ab8e0

marlowe/specification/ReadMe.md
bwbush: I'm adding an appendix to this document that will contain sufficient information to make it self contained.

marlowe/specification/ReadMe.md
bwbush: fixed in 8be04323e4ec24c4e923ed214094e85c916ab8e0

marlowe/specification/ReadMe.md
bwbush: fixed in 8be04323e4ec24c4e923ed214094e85c916ab8e0

marlowe/specification/ReadMe.md
bwbush: fixed in 8be04323e4ec24c4e923ed214094e85c916ab8e0

marlowe/specification/ReadMe.md
bwbush: fixed in 8be04323e4ec24c4e923ed214094e85c916ab8e0

marlowe/specification/ReadMe.md
bwbush: fixed in 8be04323e4ec24c4e923ed214094e85c916ab8e0

marlowe/specification/ReadMe.md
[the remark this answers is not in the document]
bwbush: I edited the diagram and added a one-paragraph caption preceding it in 2bf586c1c37999da481052591c239fa814a43bf8.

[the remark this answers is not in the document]
palas: Thanks, I think I get it now

marlowe/specification/ReadMe.md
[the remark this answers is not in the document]
palas: if it is a constraint `≡` is fine with me, assuming it makes sense mathematically

PR #225, opened 2022-08-12
Scp 4364 nonpab tests

dino-iog:
In this branch Jamie has done the heavy-lifting of upgrading to Babbage in the Haskell code, including generalizing the types. Unless anyone objects, I think we should merge this with cardano-node-1.35 sooner rather than later.


marlowe-cli/test/double-satisfaction.sh
```diff
@@ -334,8 +343,8 @@ if cardano-cli transaction build --testnet-magic "$MAGIC" --alonzo-era
                                  --invalid-hereafter "$TIMEOUT_SLOT"          \
                                  --out-file tx-2.raw
 then
-  echo "Double satisfaction was prevented."
+  echo "Failure! Double satisfaction occurred"
   exit 1
 else
-  echo "Double satisfaction was not prevented."
+  echo "Success. Double satisfaction was prevented"
```
jhbertra, line 349: @bwbush Just double checking that this edit is correct - if the command succeeds it means that double satisfaction occurs and that the test should fail. This check is the opposite of what it was before - was it wrong before?

bwbush: The message was wrong, but the exit code was correct. The `marlowe-cli/run-nonpab-tests.sh` correctly reported success or failure.

marlowe-cli/app/Main.hs
```diff
@@ -44,4 +44,3 @@ fromGit = Nothing
 #else
 fromGit = Just ((\v -> if null v then mempty else " @ " <> v) . unpack . strip $ pack $(gitRevFromGit))
 #endif
-
```
bwbush, line 47: FYI: the `gitRevFromGit` function is pretty much useless and we probably should remove it now or soon.

jhbertra (approved):

marlowe-cli/examples/cfd/run-cfd.sh
```diff
@@ -131,15 +141,15 @@ echo "The oracle mints the role tokens."
...
+$MARLOWE_CLI_BIN -- util mint --testnet-magic "$MAGIC"                  \
                       --socket-path "$CARDANO_NODE_SOCKET_PATH"         \
                       --required-signer "$ORACLE_PAYMENT_SKEY"          \
                       --change-address  "$ORACLE_ADDRESS"               \
                       --expires "$MINT_EXPIRES"                         \
                       --out-file /dev/null                              \
                       --submit=600                                      \
                       "$ORACLE_ROLE" "$PARTY_ROLE" "$COUNTERPARTY_ROLE" \
-| sed -e 's/^PolicyID "\(.*\)"$/\1/'                                    \
+| tail -n +2 | sed -e 's/^PolicyID "\(.*\)"$/\1/'                       \
```
bwbush, line 152: I actually think we should change the `putStrLn` for most of the `marlowe-cli` commands so that one doesn't need to use `sed`: i.e., `marlowe-cli` should just print the relevant hashes (or hash#ix for transactions). This is less human readable, but it makes scripting much cleaner. However, it will break all pre-existing scripts.

marlowe-cli/src/Language/Marlowe/CLI/Command/Parse.hs
```diff
@@ -138,13 +139,13 @@ parseTxIx = TxIx <$> O.auto
 
 
 -- | Parser for `TxOut` information.
-parseTxOut :: O.ReadM (AddressAny, Value)
+parseTxOut :: IsShelleyBasedEra era => O.ReadM (AddressInEra era, Value)
```
bwbush, line 142: I like the switch to `AddressInEra`!

marlowe-cli/src/Language/Marlowe/CLI/Export.hs
```diff
@@ -116,20 +123,15 @@ exportMarlowe :: MonadError CliError m
...
+    contract :: Contract <- decodeFileStrict contractFile
+    state    :: State <- decodeFileStrict stateFile
+    inputs   :: [Input] <- mapM decodeFileStrict inputFiles
+    marloweInfo@MarloweInfo{..} <- liftEither =<< asksEra \era -> buildMarlowe marloweParams era costModel network stake contract state inputs
     let
       ValidatorInfo{..} = validatorInfo
       DatumInfo{..}     = datumInfo
       RedeemerInfo{..}  = redeemerInfo
-    maybeWriteJson outputFile (marloweInfo :: MarloweInfo AlonzoEra) -- FIXME: Generalize eras.
+    doWithShelleyBasedEra $ maybeWriteJson outputFile marloweInfo
```
bwbush, line 134: Nice!

jhbertra: The magic of GATDs: summoning typeclass constraints out of thin air!

marlowe-cli/test/double-satisfaction.sh
```diff
@@ -71,11 +76,11 @@ cardano-cli query utxo --testnet-magic "$MAGIC" --address "$PARTY_A_ADDRESS"
...
-marlowe-cli util select --testnet-magic "$MAGIC"                  \
-                        --socket-path "$CARDANO_NODE_SOCKET_PATH" \
-                        --lovelace-only 50000000                  \
-                        "$PARTY_A_ADDRESS"                        \
-| sed -n -e '1{s/^TxIn "\(.*\)" (TxIx \(.*\))$/\1#\2/;p}'         \
+$MARLOWE_CLI_BIN -- util select --testnet-magic "$MAGIC"                \
+                        --socket-path "$CARDANO_NODE_SOCKET_PATH"       \
+                        --lovelace-only 50000000                        \
+                        "$PARTY_A_ADDRESS"                              \
+| tail -n +2 | sed -n -e '1{s/^TxIn "\(.*\)" (TxIx \(.*\))$/\1#\2/;p}'  \
```
bwbush, line 83: Why does this skip two lines of UTxOs (via `tail -n +2`)? Wouldn't this only work if there are at least three UTxOs at the address? (Also, the `tail -n +2` would have been redundant anyway, since the `sed` could just select the third line if that is what we want.)

bwbush (changes requested):
I read the changes in the Haskell code, which look fine. I think that the shell scripts have `tail -n +2` in places where they should not. (Remember that `marlowe-cli util select` does not print a two-line header the way that `cardano-cli query utxo does`.)

Because of this, the `run-nonpab-tests.sh` fails for me, starting from a faucet with a single pure-ada UTxO with 1000 ada:
```
 ./run-nonpab-tests.sh --faucet-address $(cat $TREASURY/payment.testnet.address) --faucet-skey-file $TREASURY/payment.skey
FAIL: test/double-satisfaction.sh
```
See [attached log file](https://github.com/input-output-hk/marlowe-cardano/files/9324614/double-satisfaction.log), which indicates UTxO-selection failure because of the `tail -n +2`.

Also, do we really want to merge this now, or wait until all tests succeed?

marlowe-cli/src/Language/Marlowe/CLI/Command/Parse.hs
paluh: Yeah, this is nice.

jhbertra:
@bwbush I agree that we should make sure the tests actually pass before merging

marlowe-cli/src/Language/Marlowe/CLI/Command/Parse.hs
```diff
@@ -208,18 +209,27 @@ readAssetIdEither s =
...
+                   . IsShelleyBasedEra era
+                  => String                    -- ^ The string to be read.
+                  -> Either String (AddressInEra era)  -- ^ Either the address or an error message.
+readAddressEither s = do
+  era <- eraAsType
+  case deserialiseAddress (AsAddressInEra era) $ T.pack s of
     Nothing      -> Left "Invalid address."
     Just address -> Right address
+  where
+    eraAsType :: Either String (AsType era)
```
paluh, line 228: I have to for sure pair with you @jhbertra so we can together read `HasTypeProxy` ;-)

marlowe-cli/src/Language/Marlowe/CLI/Command/Role.hs
```diff
@@ -72,7 +73,7 @@ data RoleCommand =
 
 
 -- | Run a related command.
-runRoleCommand :: MonadError CliError m
+runRoleCommand :: (MonadError CliError m, MonadReader (CliEnv era000) m)
```
paluh, line 76: Cosmetics `s/era000/era`?

bwbush:
@dino-iog, @paluh, @jhbertra, we might want to collect lessons-learned (perhaps in [the Transaction Creation github discussion](https://github.com/input-output-hk/marlowe-cardano/discussions/213) or in [SCP-3973](https://input-output.atlassian.net/browse/SCP-3973) from this era generalization because it is relevant for Marlowe Runtime's Transaction Creation. BTW, we'll be able to replace `marlowe-cli`'s transaction-creation code with that from Marlowe Transaction Creation eventually.

bwbush:
Also, please bump the `marlowe-cli` version to `0.6.0.0`.

paluh (approved):

marlowe-cli/test/double-satisfaction.sh
dino-iog: I was trying to make it work with both `marlowe-cli` or `cabal run marlowe-cli --` but this is broken for `marlowe-cli` and probably too complicated to bother with. I'll get rid of that and we can use rely on having the binary built with nix.

PR #232, opened 2022-08-24
SCP-4276 Follow a Contract

jhbertra:
Changes include:

- Adding a `Follower` component with a dedicated CLI executable to the runtime that follows a single contract
- Extending `chainseekd` with a new query API for direct access. Currently supports querying some data from the Genesis parameters.
- Refactoring protocols that pass query GADTs (e.g. chain seek, job, query) to associate them with a `Tag` GADT. This helps a lot serialization a lot by separating the questions "which query are we decoding" and "how do I decode the query / result / error?"

### DIFF
--- a/marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
+++ b/marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
@@ -398,6 +398,7 @@ queryTxIns slotNo txInId = HT.statement (Api.unTxId txInId, fromIntegral slotNo)
     SELECT txIn.txOutId :: bytea
          , txIn.txOutIx :: smallint
          , txOut.address :: bytea
+         , txOut.datumBytes :: bytea?
          , txIn.redeemerDatumBytes :: bytea?
       FROM chain.txIn  as txIn
       JOIN chain.txOut as txOut ON txOut.txId = txIn.txOutId AND txOut.txIx = txIn.txOutIx
@@ -405,10 +406,11 @@ queryTxIns slotNo txInId = HT.statement (Api.unTxId txInId, fromIntegral slotNo)
   |] (Fold.foldMap (Set.singleton . decodeTxIn) id)
   where
     decodeTxIn :: ReadTxInRow -> Api.TransactionInput
-    decodeTxIn (txId, txIx, address, redeemerDatumBytes) = Api.TransactionInput
+    decodeTxIn (txId, txIx, address, datumBytes, redeemerDatumBytes) = Api.TransactionInput
       { txId = Api.TxId txId
       , txIx = fromIntegral txIx
       , address = Api.Address address
+      , datumBytes = Api.fromPlutusData . toPlutusData . unsafeDeserialize' <$> datumBytes
       , redeemer = Api.Redeemer . Api.fromPlutusData . toPlutusData . unsafeDeserialize' <$> redeemerDatumBytes
       }
 
@@ -476,6 +478,7 @@ type ReadTxInRow =
   ( ByteString       -- TxIn's TxId
   , Int16            -- TxIn's TxIx
   , ByteString       -- TxIn's Address
+  , Maybe ByteString -- TxIn's datumBytes
   , Maybe ByteString -- TxIn's redeemerDatumBytes
   )
 
--- a/marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
+++ b/marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
@@ -310,7 +310,7 @@ performFindConsumingTx Api.TxOutRef{..} point = do
       , tokenName
       , quantity
       ) | slotNo <= pointSlot point = MoveAbort $ Api.UTxOSpent $ Api.TxId spendingTxId
-        | otherwise                 = MoveArrive (decodeBlockHader (slotNo, hash, blockNo)) Api.Transaction
+        | otherwise                 = MoveArrive (decodeBlockHeader (slotNo, hash, blockNo)) Api.Transaction
           { txId = Api.TxId spendingTxId
           , validityRange = case (validityLowerBound, validityUpperBound) of
               (Nothing, Nothing) -> Api.Unbounded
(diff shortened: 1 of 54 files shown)


marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
```diff
@@ -319,6 +324,29 @@ performFindConsumingTx Api.TxOutRef{..} point = do
           }
     readFirstTxRow _ = MoveWait
 
+performFindConsumingTxs :: Set Api.TxOutRef -> Api.ChainPoint -> Transaction (PerformMoveResult (Map Api.TxOutRef Api.UTxOError) (Map Api.TxOutRef Api.Transaction))
+performFindConsumingTxs utxos point = do
```
bwbush, line 328: This function is complex enough that it might warrant some internal comments.

marlowe-protocols-test/src/Test/Network/Protocol/ChainSeek.hs
```diff
@@ -0,0 +1,81 @@
...
+      -> ClientStIdle q point tip m b
+      -> m (a, Maybe b)
+    runIdle = \case
+      Do action script' -> \client -> action *> runIdle script' client
+      Halt a -> \_ -> pure (a, Nothing)
+      ExpectDone a -> \case
+        SendMsgDone b -> pure (a, Just b)
+        SendMsgQueryNext move _ _ -> liftIO do
+          expectationFailure $ "Expected MsgDone, got MsgQueryNext (" <> showQuery move <> ")"
+          undefined
```
bwbush, line 54: Is this work in progress? `undefined`?

marlowe-runtime/app/Main.hs
```diff
@@ -1,8 +1,4 @@
 module Main where
 
-import qualified MyLib (someFunc)
-
 main :: IO ()
-main = do
-  putStrLn "Hello, Haskell!"
-  MyLib.someFunc
+main = undefined
```
bwbush, line 4: So the plan is to put APIs in `marlowe-runtime`. Will the `main` here evolve into a CLI entry point? For clients, for control, or both?

marlowe-runtime/src/Language/Marlowe/Runtime/History/Follower.hs
```diff
@@ -0,0 +1,559 @@
...
+data ClosedPreviousState v
+  = ClosedPreviousOpen (FollowerState v)
+  | ClosedPreviousClosed (FollowerStateClosed v)
+
+data FollowerStateClosed v = FollowerStateClosed
+  { payouts       :: Map Chain.TxOutRef (Payout v)
+  , previousState :: PreviousState (ClosedPreviousState v)
+  }
+
+exctractCreation :: FollowerDependencies -> Chain.Transaction -> Either ExtractCreationError SomeCreateStep
```
bwbush, line 246: typo in fn name

marlowe-protocols-test/src/Test/Network/Protocol/ChainSeek.hs
jhbertra: no... it't not. `expectationFailure` has a bad type of `String -> Expectation` where `type Expectation = IO ()`. But it throws an exception, so the type should actually be `String -> IO a`. Putting `undefined` here is a bluff to work around this - it should never be evaluated.

marlowe-runtime/src/Language/Marlowe/Runtime/History/Follower.hs
```diff
@@ -0,0 +1,559 @@
...
+  { payouts       :: Map Chain.TxOutRef (Payout v)
+  , scriptOutput  :: TransactionScriptOutput v
+  , previousState :: Maybe (PreviousState (FollowerState v))
+  }
+
+data ClosedPreviousState v
+  = ClosedPreviousOpen (FollowerState v)
+  | ClosedPreviousClosed (FollowerStateClosed v)
+
+data FollowerStateClosed v = FollowerStateClosed
```
bwbush, line 241: Would it be worthwhile to move all the type declarations to a module `Language.Marlowe.Runtime.History.Follower.Types`?

marlowe-runtime/app/Main.hs
jhbertra: This `main` will eventually evolve into the CLI for the runtime daemon as a singleton process.

marlowe-runtime/src/Language/Marlowe/Runtime/History/Follower.hs
jhbertra: Not these ones, they are internal implementation details, not part of the API (I don't want them used elsewhere as they may change)

marlowe-runtime/src/Language/Marlowe/Runtime/History/Follower.hs
```diff
@@ -0,0 +1,559 @@
...
+    payoutDatum <- note (InvalidPayoutDatum txOut) $ fromChainPayoutDatum version rawPayoutDatum
+    pure $ Payout assets payoutDatum
+  let output = TransactionOutput{..}
+  pure Transaction{..}
+
+wouldCloseContract :: MarloweVersion v -> UTCTime -> UTCTime -> Redeemer v -> Datum v -> Bool
+wouldCloseContract version validityLowerBound validityUpperBound redeemer datum = case version of
+  MarloweV1 ->
+    let
+      utcTimeToPOSIXTime = POSIXTime . round . (* 1000) . utcTimeToPOSIXSeconds
```
bwbush, line 530: I think this is okay, but we have to be careful about rounding milliseconds in times. (We've previously encountered sub-millisecond failures in `computeTransaction` and in the validator because of this.) Let's make sure we eventually have a thorough set of tests for times and slots. Also, we might want to consider debug logging failures from `computeTransaction` so these time-skew failures don't slip by us.

marlowe/src/Language/Marlowe/Core/V1/Semantics.hs
```diff
@@ -104,7 +104,7 @@ import Text.PrettyPrint.Leijen (comma, hang, lbrace, line, rbrace, space, text,
     when positive balances are payed out on contract closure.
 -}
 data Payment = Payment AccountId Payee Money
-  deriving stock (Haskell.Show)
+  deriving stock (Haskell.Show, Haskell.Eq)
```
bwbush, line 107: This induced a compilation failure for `exe:marlowe-test`. Here is the patch that fixes it:
```
diff --git a/marlowe/test/Spec/Marlowe/Orphans.hs b/marlowe/test/Spec/Marlowe/Orphans.hs
index 7a9b351db..23ee77c1d 100644
--- a/marlowe/test/Spec/Marlowe/Orphans.hs
+++ b/marlowe/test/Spec/Marlowe/Orphans.hs
@@ -10,9 +10,6 @@ module Spec.Marlowe.Orphans (
 import Language.Marlowe.Core.V1.Semantics
 
 
-deriving instance Eq Payment
-
-
 deriving instance Eq ApplyWarning
```

I recommend running `cabal build all` and `cabal test all` for PRs.

bwbush (approved):
Excellent implementation! I especially like the `IsQuery`/`Query`, the versioning, and the test mini-DSL. The tests ran successfully. I gather that this is still work in progress because of several `undefined`s.

There's one patch that needs applying to fix an orphan instance that is no longer needed downstream, since `cabal test all` fails otherwise.

marlowe-protocols-test/src/Test/Network/Protocol/ChainSeek.hs
bwbush: I suggest adding an inline comment after `undefined`s like this that point. Other alternatives are `Test.Hspec.Expectations.Lifted.expectationFailure` (from `hspec-expectations-lifted`) or even `forever`. Anyway, either the comment or the function name probably should indicate what is going on here.

marlowe/src/Language/Marlowe/Core/V1/Semantics.hs
jhbertra: Thanks for that! I wonder why this was defined here instead of in the original module - is it to save space in the compiled plutus code? Also, RE:

> I recommend running `cabal build all` and `cabal test all` for PRs.

But that's why we have a build server, so I don't have to do this locally :smirk:

marlowe-runtime/src/Language/Marlowe/Runtime/History/Follower.hs
jhbertra: Yeah, they should be logged as warnings

marlowe-runtime/src/Language/Marlowe/Runtime/History/Follower.hs
jhbertra: In a subsequent PR, I will extract these types as they will begin to be shared some more.

marlowe-runtime/src/Language/Marlowe/Runtime/History/Follower.hs
jhbertra: Thank you!

marlowe-protocols-test/src/Test/Network/Protocol/ChainSeek.hs
jhbertra: thanks!

marlowe/src/Language/Marlowe/Core/V1/Semantics.hs
jhbertra: I'll actually just revert this change, as I don't need it any longer

PR #237, opened 2022-08-31
SCP-4275 Implement Query API for Marlowe Contract History

jhbertra:


marlowe-protocols/src/Network/Protocol/ChainSeek/Types.hs
```diff
@@ -13,17 +14,15 @@ import Data.Kind (Type)
 import Data.String (IsString)
 import Data.Text (Text)
 import qualified Data.Text.Encoding as T
+import Data.Type.Equality (type (:~:))
```
bwbush, line 17: :sunglasses:

marlowe-runtime/marlowe-follower/Options.hs
```diff
@@ -0,0 +1,51 @@
...
+getOptions :: IO Options
+getOptions = execParser $ info (helper <*> parser) infoMod
+  where
+    parser = Options <$> portParser <*> queryPortParser <*> hostParser <*> contractIdParser
+
+    portParser = option auto $ mconcat
+      [ long "port-number"
+      , value 3715
+      , metavar "PORT_NUMBER"
+      , help "The port number of the chain seek server. Default value: 3715"
```
bwbush, line 24: Thanks for including default values in the help! :smile:

marlowe-runtime/src/Language/Marlowe/Protocol/Sync/Codec.hs
```diff
@@ -0,0 +1,134 @@
...
+      MsgIntersectFound blockHeader -> do
+        putWord8 0x0c
+        put blockHeader
+      MsgIntersectNotFound -> putWord8 0x0d
+
+  getMessage :: GetMessage MarloweSync
+  getMessage tok = do
+    tag <- getWord8
+    case tag of
+      0x01 -> case tok of
```
bwbush, line 68: Are these purely internal intra-Runtime tags, or do we need to document these for external (non-Runtime) use?

marlowe-runtime/marlowe-history-cli/Main.hs
```diff
@@ -1,14 +1,27 @@
...
+import qualified Data.Text as T
 import qualified Data.Text.IO as T
 import Data.Void (Void, absurd)
 import GHC.Base (when)
-import Language.Marlowe.Runtime.Core.Api (ContractId, parseContractId, renderContractId)
+import qualified Language.Marlowe.Core.V1.Semantics as V1
+import Language.Marlowe.Pretty (pretty)
+import qualified Language.Marlowe.Protocol.Sync.Client as MarloweSync
+import Language.Marlowe.Protocol.Sync.Codec (codecMarloweSync)
+import qualified Language.Marlowe.Protocol.Sync.Types as Sync
```
bwbush, line 19: I'm receiving compile errors due to unused imports:
```
marlowe-history-cli/Main.hs:19:1: error: [-Wunused-imports, -Werror=unused-imports]                                                  
    The qualified import of ‘Language.Marlowe.Protocol.Sync.Types’ is redundant                                                      
      except perhaps to import instances from ‘Language.Marlowe.Protocol.Sync.Types’                                                 
    To import instances alone, use: import Language.Marlowe.Protocol.Sync.Types()                                                    
   |                                                                                                                                 
19 | import qualified Language.Marlowe.Protocol.Sync.Types as Sync                                                                   
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
                                                                                                                                     
marlowe-history-cli/Main.hs:33:1: error: [-Wunused-imports, -Werror=unused-imports]                     
    The import of ‘Peer, PeerHasAgency, PeerRole’                                                                                    
    from module ‘Network.TypedProtocol’ is redundant                                                                                 
   |                                                                                                                                 
33 | import Network.TypedProtocol (Driver (..), Peer (..), PeerHasAgency (..), PeerRole (..), runPeerWithDriver)                     
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                     
```

marlowe-runtime/marlowe-follower/Options.hs
jhbertra: I think there is actually an `optparse-applicative` combinator that adds this. I'll look into this, it would be nice to have this generated instead.

marlowe-runtime/src/Language/Marlowe/Protocol/Sync/Codec.hs
jhbertra: No, this protocol is an external-facing API, and these tag values are part of the binary serialization schema for the messages. So yes, this does need to be documented. Haskell clients will be able to use the typed interface which already implements this serialization scheme, but implementations in other languages will need to explicitly follow the specification.

marlowe-runtime/marlowe-history-cli/Main.hs
jhbertra: Odd, the CI should be failing in that case...

bwbush: I did a fresh clone, `nix develop`, and `cabal build all`.

(This is not blocking my review: I fixed the problem locally, so I can test the executables.)

bwbush:
Another `optparse-applicative` annoyance is that we don't get a list of commands by default at top-level without `--help`. This isn't something to fix now, but I thought I'd note it.
```
$ marlowe-history-cli
Missing: (COMMAND | COMMAND)

Usage: marlowe-history-cli [--command-port PORT_NUMBER] 
                           [--query-port PORT_NUMBER] [--sync-port PORT_NUMBER] 
                           [-h|--host HOST_NAME] (COMMAND | COMMAND)

  Example Chain Seek client for the Marlowe Chain Sync
```

bwbush (approved):
I executed the history and follower commands on the `preview` network without any problems.

The code looks great, but I have a lingering concern about our use of UTC time in some places. Eventually, we'll need thorough testing of time conversions and the slotting throughout Runtime because even sub-second skews have caused validation to fail.

jhbertra:
I have also got lingering concerns, and I suspect that there is a bug still in rollback handling. Testing is becoming a critical priority.

PR #238, opened 2022-08-31
SCP-4267 SCP-4268 SCP-4269 Corrections to property-based tests

bwbush:
This contains three testing-related changes:
1. SCP-4267 allows the false-positives static-analysis test to timeout without failing.
2. SCP-4269 confirms the warning message for zero payment.
3. SCP-4268 fixes a logical error in the test for `getContinuation`.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marlowe/test/Spec/Marlowe/Semantics/Functions.hs
+++ b/marlowe/test/Spec/Marlowe/Semantics/Functions.hs
@@ -59,11 +59,6 @@ import Test.Tasty.QuickCheck (Arbitrary (..), Gen, Property, Testable (property)
 import qualified PlutusTx.AssocMap as AM (delete, empty, filter, fromList, insert, keys, lookup, member, null, toList)
 
 
--- | FIXME: Turn this off when semantics are fixed, see SCP-4269.
-_ALLOW_ZERO_PAYMENT :: Bool
-_ALLOW_ZERO_PAYMENT = True
-
-
 -- | Run the tests.
 tests :: TestTree
 tests =
@@ -746,7 +741,7 @@ checkReduceContractStepPay =
           && Party payee' == payee
           && case flattenMoney money of
                [(token', amount)] -> token == token' && amount == debit && state' `stateEq` newState
-               []                 -> _ALLOW_ZERO_PAYMENT && state' `stateEq` state
+               []                 -> state' `stateEq` state
                _                  -> False
       checkPayment (Payment account' (Account payee') money) state' =
         let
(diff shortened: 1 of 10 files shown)


palas (approved):
Looks good.

I guess it would make more sense as a cabal variable, since it is pretty deployment specific (we could think a slower machine needs a longer timeout). But I wouldn't have said it if you hadn't asked, and we may want to have a default value if the variable is unset

Another thing is, I am not sure the timeout function will kill `z3` if it is running (it may remain zombie). There is a way of passing timeout values through SBV, but those are not guaranteed to work either https://hackage.haskell.org/package/sbv-9.0/docs/Data-SBV.html#v:setTimeOut.

bwbush:
@palas, I will revise this to use a cabal variable.

A zombie `z3` won't impede the progression of testing, though that `z3` will use some resources on the VM where the test is performed. Since this error only happened about 2% of the time, I don't think it is worthwhile to dig deep in the process space to kill the `z3`.

bwbush:
@palas, would you please re-review the cabal flag addition and two commits with minor testing-related fixes? Thanks.

marlowe/test/Spec/Marlowe/Semantics/Functions.hs
```diff
@@ -882,11 +872,11 @@ checkGetContinuation =
...
            pure (input, case', contract, contractHash == contractHash')
     forAll' gen $ \(input, case', contract, hashesMatch) ->
-      _ALLOW_FAILED_CONTINUATION_ || case (input, case', getContinuation input case') of
-        (NormalInput{}    , Case{}          , contract') -> Just contract == contract'
-        (MerkleizedInput{}, MerkleizedCase{}, contract') -> (Just contract == contract') == hashesMatch
-        (_                , _               , Nothing  ) -> True
-        _                                                -> False
+      case (input, case', getContinuation input case') of
+        (NormalInput{}    , Case{}                                       , contract') -> Just contract == contract'
+        (MerkleizedInput _ contractHash _, MerkleizedCase _ contractHash', _        ) -> (contractHash == contractHash') == hashesMatch
```
palas, line 877: Wait, isn't this shortcutting the test? Not part of the PR but just noticed
```suggestion
        (MerkleizedInput _ contractHash _, MerkleizedCase _ contractHash', _        ) -> contractHash == contractHash'
```

palas (approved):
Thank you. PR looks good, but I noticed something strange

marlowe/test/Spec/Marlowe/Semantics/Functions.hs
bwbush: Thanks for catching this! The combination of the somewhat convoluted generator and my trying to make the pattern matching too succinct end in making one of the matches tautological. This is fixed in 51fafdc82035fb096495f534907ea823171cc3e5, which is verbose and straightforward.

PR #241, opened 2022-09-02
[DO NOT MERGE] Scp 4416/generate choice test data

ladamesny:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


marlowe/test/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -419,6 +419,64 @@ choiceNotInBounds bounds =
...
+
+interestingParties :: Party -> [Party]
+interestingParties (Role "Alice") = [Role "Alice", Role "Alic", Role "Alices"]
+
+interestingChoiceNum :: [Bound] -> [ChoiceNum]
+interestingChoiceNum bounds = concatMap interestingChoiceNum' bounds
+
+interestingChoiceNum' :: Bool -> Bound -> [ChoiceNum]
+interestingChoiceNum' True (Bound lower upper)
+  | lower > upper = Nothing
```
ladamesny, line 440: @bwbush Do we want to raise an error in this case? Not sure how to handle this.

marlowe/test/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -419,6 +419,64 @@ choiceNotInBounds bounds =
...
+        lower
+      , lower + 1
+      , upper - 1
+      , upper
+      , (lower + upper) `div` 2
+      ]
+
+interestingChoiceNum' False (Bound lower upper)
+  | lower > upper = Nothing
+  | lower == upper =
```
ladamesny, line 457: I think I can collapse this case with the `otherwise` case below - looks like the same output

marlowe/test/Spec/Marlowe/Semantics/Arbitrary.hs
bwbush: No, since the error would be caught and dealt with elsewhere. In this case nothing is interesting, so it should return `[]`.

marlowe/test/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -419,6 +419,64 @@ choiceNotInBounds bounds =
...
+      [
+        lower
+      , lower + 1
+      , upper - 1
+      , upper
+      , (lower + upper) `div` 2
+      ]
+
+interestingChoiceNum' False (Bound lower upper)
+  | lower > upper = Nothing
```
ladamesny, line 456: Let's give it cases (of anything - lower, upper and edges)

reasoning - when lower is greater than upper, we want to make sure it fails when we give it anything.

marlowe/test/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -419,6 +419,94 @@ choiceNotInBounds bounds =
...
+interestingParties (Role "Alice") = [Role "Alice", Role "Alic", Role "Alices"]
+
+interestingChoiceNum :: [Bound] -> [ChoiceNum]
+interestingChoiceNum bounds = concatMap interestingChoiceNum' bounds
+
+interestingChoiceNum' :: Bool -> Bound -> [ChoiceNum]
+interestingChoiceNum' True (Bound lower upper)
+  | lower > upper = []
+  | lower == upper = [lower]
+  | otherwise = Data.List.nub validValues lower upper
```
bwbush, line 442: Nice!

marlowe/test/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -419,6 +419,94 @@ choiceNotInBounds bounds =
...
+
+interestingChoiceNum' False (Bound lower upper)
+  | lower > upper = Data.List.nub allValues lower upper
+  | otherwise = Data.List.nub invalidValues lower upper
+
+validValues :: Int -> Int -> [Int]
```
bwbush, line 448: We might want to give this a slightly different name, since if `lower == upper`, it *doesn't* produce valid values. Or there could be guards in the functions it calls.

marlowe/test/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -419,6 +419,94 @@ choiceNotInBounds bounds =
...
+validValues lower upper =
+  validLowerValues lower
+  <> validUpperValues upper
+  <> valuesWithinRange lower upper
```
bwbush, line 452: Sometimes the base values are valid, too.

```
validValues lower upper =
  filter inRange
  $ validLowerValues lower
  <> validUpperValues upper
  <> valuesWithinRange lower upper
  <> baseValues
  where inRange x = x >= lower && x <= upper
```

marlowe/test/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -419,6 +420,35 @@ choiceNotInBounds bounds =
...
+interestingChoiceNum' False (Bound lower upper) = invalidValues lower upper
+
+validValues :: Integer -> Integer -> [Integer]
+validValues lower upper = [x | x <- testValueRange, x >= lower  && x <= upper]
+
+invalidValues :: Integer -> Integer -> [Integer]
+invalidValues lower upper = [x | x <- testValueRange, x < lower  || x > upper]
+
+testValueRange :: [Integer]
+testValueRange = [(-10000)..10000]
```
ladamesny, line 450: hey @bwbush I think I found a cleaner solution to this problem, thanks to your suggestion of thinking about this problem is a number line. However, this code is still dependent on a hard-coded number range. Is there a better way to solve this problem? Maybe we can circle back to this when it starts failing?

ladamesny, line 450:
```suggestion
testValueRange = [(-rangeLimit)..rangeLimit]

rangeLimit :: Integer
rangeLimit = 10000
```

marlowe/test/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -419,6 +420,38 @@ choiceNotInBounds bounds =
...
+validValues lower upper = [x | x <- testValueRange, x >= lower  && x <= upper]
+
+invalidValues :: Integer -> Integer -> [Integer]
+invalidValues lower upper = [x | x <- testValueRange, x < lower  || x > upper]
+
+testValueRange :: [Integer]
+testValueRange = [(-rangeLimit)..rangeLimit]
+
+rangeLimit :: Integer
+rangeLimit = 10000
```
ladamesny, line 453: Hi @bwbush! thanks to your suggestion of thinking about this problem as a number line, I think I was able to clean this code up a bit. However, it's still dependent on a hard-coded value in the `rangeLimit` function. We can also circle back on this when the tests start failing?

bwbush: Let's leave this as it is for right now. Later we'll use an `Arbitrary Integer` instance.

marlowe/test/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -419,6 +420,38 @@ choiceNotInBounds bounds =
...
+interestingChoiceNum' True (Bound lower upper)  = validValues lower upper
+interestingChoiceNum' False (Bound lower upper) = invalidValues lower upper
+
+validValues :: Integer -> Integer -> [Integer]
+validValues lower upper = [x | x <- testValueRange, x >= lower  && x <= upper]
+
+invalidValues :: Integer -> Integer -> [Integer]
+invalidValues lower upper = [x | x <- testValueRange, x < lower  || x > upper]
+
+testValueRange :: [Integer]
```
bwbush, line 449: Change to an infinite list like `[0, 1, -1, 2, -2]`.

bwbush (approved):
We reviewed this in meetings. It looks great!

PR #243, opened 2022-09-02
Implemented the typed-protocols Marlowe Sync client

dino-iog:
- Made new data and GADT definitions
- Implemented the Client hoist functions
- Implemented the Client peer functions
- Changed blockHeader* to header* for consistency
- Updated updateMaterialized

### DIFF
--- a/marlowe-runtime/marlowe-runtime.cabal
+++ b/marlowe-runtime/marlowe-runtime.cabal
@@ -64,6 +64,7 @@ library
   import: lang
   hs-source-dirs:   src
   exposed-modules:
+    Language.Marlowe.Protocol.Sync.Client
     Language.Marlowe.Protocol.Sync.Codec
     Language.Marlowe.Protocol.Sync.Server
     Language.Marlowe.Protocol.Sync.Types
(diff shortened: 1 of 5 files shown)


jhbertra:
Copying this over from the last PR:

The changes look great! If you want, you could also take a go at rewriting `runShowCommand` in `marlowe-runtime/marlowe-history-cli/Main.hs` to use the `MarloweSyncClient` instead of the raw `Peer` (which it currently uses). Something like this:

```hs
-- before
(result, _) <- runPeerWithDriver driver peerInit (startDState driver)
...

-- after
(result, _) <- runPeerWithDriver driver (marloweSyncClientPeer client) (startDState driver)
...
where
  client = MarloweSyncClient $ pure $ clientInit
  clientInit = ...
```

dino-iog:
Ok, started working on that. I'm getting errors because we don't have a `data ClientStNextCanReject`, do we need that in addition to `ClientStNext` ?

jhbertra:
I think you are working with the wrong types - check where you are importing `ClientStNext` from)

PR #252, opened 2022-09-12
SCP-4445: Update Readme with up to date NixOS instructions

bjornkihlberg:
The NixOS instructions were slightly outdated with deprecated fields. With this PR I can also demonstrate that I have correctly set up my PGP signature with git.


palas (approved):
Thanks, looks good

PR #254, opened 2022-09-12
SCP-4372 Stop stylish-haskell from forcing alignment

dino-iog:
This is a preliminary fix to remove the settings from stylish-haskell that were forcing column alignment on us in many parts of Haskell code. It also will not care about existing alignment in source files, which will minimize noisy changes to existing code.

I tested it by changing formatting in a source file of some LANGUAGE pragmas, imports, type signatures, Haddock comments and case statement `->` alignments to confirm it both ignores existing indentation and also allows no alignment.

### DIFF
--- a/.stylish-haskell.yaml
+++ b/.stylish-haskell.yaml
@@ -1,15 +1,5 @@
 ---
 steps:
-  - simple_align:
-      cases: always
-      top_level_patterns: always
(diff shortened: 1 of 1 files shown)


dino-iog:
SCP-4454-remove-stylish-alignment

PR #258, opened 2022-09-14
SCP-4451 history tests

jhbertra:
I wrote these tests by writing a pure model of the history store queries, writing property tests for that, then writing equivalence tests between the public API of the HistoryStore and the model.

The `HistoryScript` is a sequence of actions which create and advance contracts in a very simple mock of the blockchain, and which manipulate that mock by rolling it forwards and backwards. It has an `Arbitrary` instance that generates valid scripts. Using the script I can set up both the model store and the real store so that they have equivalent states. Once that is done, I can generate query parameters for the model and check them against both the real and the model implementation to make sure they produce the same result.

This will make it easy to test future implementations of the queries that use, for example, a real database, because we can just plug in a different set of queries and run the same store tests.


bwbush (approved):
Reviewed via google meet. Looks great!

PR #264, opened 2022-09-15
Post-Constant Validator Runtime Updates

jhbertra:
- [x] Adds structures and helpers for maintaining a repository of statically known script hashes and addresses
- [x] Adds tests to make sure these are kept up-to-date
- [x] Removes the check for close/create coincidences

### DIFF
--- a/marlowe-runtime/src/Language/Marlowe/Runtime/Core/AddressRegistry.hs
+++ b/marlowe-runtime/src/Language/Marlowe/Runtime/Core/AddressRegistry.hs
@@ -1,92 +1,67 @@
-{-# LANGUAGE DataKinds #-}
-{-# LANGUAGE DuplicateRecordFields #-}
-{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GADTs #-}
-{-# LANGUAGE KindSignatures #-}
-{-# LANGUAGE MultiParamTypeClasses #-}
-{-# LANGUAGE RankNTypes #-}
-{-# LANGUAGE TypeFamilies #-}
-{-# LANGUAGE TypeOperators #-}
 
 module Language.Marlowe.Runtime.Core.AddressRegistry
   where
 
-import Cardano.Api.Byron (NetworkId(..))
 import Data.Foldable (asum)
 import Data.Map (Map)
 import qualified Data.Map as Map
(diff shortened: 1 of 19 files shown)


paluh:
I'm not sure if we really want to transition from script hashes to script addresses (without delegation). In principle Marlowe contract can be published on an address with stake address reference.

Side note:
`marlowe-cli` provides a way to publish our reference scripts permanently on the chain under [predictable and unspendable address](https://github.com/input-output-hk/marlowe-cardano/blob/main/marlowe-cli/src/Language/Marlowe/CLI/Cardano/Api/Address/ProofOfBurn.hs#L48). This simplifies referencing a lot (every transaction which wants to use reference script has to provide an `TxIn` as reference input to the transaction). We can discuss if we want to make it a standard and use in the context of Marlowe script registry.

marlowe-runtime/src/Language/Marlowe/Runtime/Core/AddressRegistry.hs
```diff
@@ -0,0 +1,92 @@
...
+import qualified Data.Set as Set
+import Data.String (IsString(fromString))
+import Language.Marlowe.Runtime.ChainSync.Api (Address, ScriptHash)
+import Language.Marlowe.Runtime.Core.Api
+
+-- | The hash and address for a script.
+data ScriptAddressInfo = ScriptAddressInfo
+  { scriptAddress :: Address
+  , scriptHash :: ScriptHash
+  } deriving (Show, Eq, Ord)
```
bwbush, line 28: A given version of the Marlowe script will have a unique payment PKH, but the staking part of the address will vary. (There are important use cases where the staking address will be set to something other than `NoStakingKey`.) How are we handling this in the registry and in transaction detection?

bwbush (approved):
It's not clear to me how the staking part of a Marlowe validator address is handled. Will this code be robust with respect to different staking parts of the address?

jhbertra:
@paluh @bwbush thank you both for your feedback. I have updated the PR so that instead of having an `AddressRegistry` that keeps track of a set of static addresses, it is a `ScriptRegistry` that keeps track of a set of static script hashes. Construction and inspection of addresses will be outside the scope of this module, which really is only about pinning known script hashes after all.

PR #265, opened 2022-09-15
Project scaffolding for Tx Creation

jhbertra:


marlowe-runtime/src/Language/Marlowe/Runtime/Transaction/Api.hs
```diff
@@ -0,0 +1,322 @@
...
+import Language.Marlowe.Runtime.Transaction.Constraints (UnsolvableConstraintsError)
+import Network.Protocol.Job.Types (Command(..), SomeTag(..))
+
+-- | The low-level runtime API for building and submitting transactions.
+data MarloweTxCommand status err result where
+  -- | Construct a transaction that starts a new Marlowe contract. The
+  -- resulting, unsigned transaction can be signed via the cardano API or a
+  -- wallet provider. When signed, the 'Submit' command can be used to submit
+  -- the transaction to the attached Cardano node.
+  Create
```
paluh, line 32: We want to probably add `StakeAddressReference` here.

jhbertra: Good idea, added

marlowe-runtime/src/Language/Marlowe/Runtime/Transaction/Api.hs
```diff
@@ -0,0 +1,363 @@
...
+    -- ^ The Marlowe version to use
+    -> WalletAddresses
+    -- ^ The wallet addresses to use when constructing the transaction
+    -> ContractId
+    -- ^ The ID of the contract to apply the inputs to.
+    -> Maybe UTCTime
+    -- ^ The "invalid before" bound of the validity interval. If omitted, this
+    -- is computed from the contract.
+    -> Maybe UTCTime
+    -- ^ The "invalid hereafter" bound of the validity interval. If omitted, this
```
bwbush, line 77: I don't think it is feasible to compute this from the contract. We can compute an _upper bound_, but that likely will be too far in the future and the node will reject the transaction.

Babbage has some rules about validity ranges crossing epoch boundaries etc. I suppose we could fetch those rules and then compute the minimum of the contract's upper bound and the ledger's upper bound.

marlowe-runtime/src/Language/Marlowe/Runtime/Transaction/Api.hs
```diff
@@ -0,0 +1,363 @@
...
+  ApplyInputs
+    :: ScriptDataSupportedInEra era
+    -> MarloweVersion v
+    -- ^ The Marlowe version to use
+    -> WalletAddresses
+    -- ^ The wallet addresses to use when constructing the transaction
+    -> ContractId
+    -- ^ The ID of the contract to apply the inputs to.
+    -> Maybe UTCTime
+    -- ^ The "invalid before" bound of the validity interval. If omitted, this
```
bwbush, line 74: Possible, but we'd need the current slot number of the node, etc.

marlowe-runtime/src/Language/Marlowe/Runtime/Transaction/Api.hs
```diff
@@ -0,0 +1,363 @@
...
+import Data.Type.Equality (type (:~:)(Refl))
+import Data.Void (Void, absurd)
+import GHC.Generics (Generic)
+import Language.Marlowe.Runtime.ChainSync.Api (Address, BlockHeader, TokenName, TxId, getUTCTime, putUTCTime)
+import Language.Marlowe.Runtime.Core.Api
+import Language.Marlowe.Runtime.Transaction.Constraints (UnsolvableConstraintsError)
+import Network.Protocol.Job.Types (Command(..), SomeTag(..))
+
+-- | The low-level runtime API for building and submitting transactions.
+data MarloweTxCommand status err result where
```
bwbush, line 37: Is signing part of this design? Would it be appropriate for a `Sign` constructor to trigger a handoff to the actual signing? (Probably not, but I thought I'd mention this possiblity.)

bwbush (approved):
Looks great. The automatic interval selection might be impractical to implement, but we can try. Also, we may want to give the client some control over choice of collateral, since that's pretty standard practice in most Dapp wallets.

marlowe-runtime/src/Language/Marlowe/Runtime/Transaction/Api.hs
jhbertra: I didn't know about this - in the past, in Marlowe Run, we always computed the upper bound from the contract which is why I assumed we could do it in the runtime. Maybe as you say, we need to do a `min` comparison between what the contract says and what the ledger says.

marlowe-runtime/src/Language/Marlowe/Runtime/Transaction/Api.hs
jhbertra: That can easilly be obtained from `chainseekd` - in fact, I would go so far as to say this should be the _preferred_ way to do it, as it avoids that mess we ran into when trying to do it from Marlowe Run.

marlowe-runtime/src/Language/Marlowe/Runtime/Transaction/Api.hs
jhbertra: Signing is not part of this design. The runtime returns an unsigned TxBody for you to sign however you want, then accepts a signed Tx body in return. This allows wallet integration to be fully handled by clients of the runtime.

The CLI on the other hand should provide integrations with different wallet types (this might be our CIP-30 client actually).

PR #267, opened 2022-09-16
SCP-4465 add flag to control simulation mode for marlowe cli dsl

ladamesny:
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marlowe-cli/src/Language/Marlowe/CLI/Test/Script.hs
+++ b/marlowe-cli/src/Language/Marlowe/CLI/Test/Script.hs
@@ -101,6 +101,7 @@ import Language.Marlowe.CLI.Test.Types
   , ContractSource(..)
   , CurrencyNickname
   , CustomCurrency(CustomCurrency, ccCurrencySymbol)
+  , ExecutionMode(..)
   , MarloweContract(..)
   , PartyRef(RoleRef, WalletRef)
   , ScriptEnv(..)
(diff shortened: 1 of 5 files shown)


marlowe-cli/src/Language/Marlowe/CLI/Test/Script.hs
```diff
@@ -348,15 +351,16 @@ interpret so@Publish {..} = do
...
+      -- Seconds timeout <- view seTransactionTimeout
+      executionTimeout <- view seExecutionTimeout
       runSoCli so $ publishImpl
         connection
         waSigningKey
         Nothing
         waAddress
         publishingStrategy
         (CoinSelectionStrategy False False [])
-        timeout
```
ladamesny, line 359: @paluh I noticed here we weren't using `(Just timeout)` here. Maybe we need to update `publishImpl` to accept a Maybe for this argument, otherwise the simulation feature will not work correctly for this function, right?

marlowe-cli/src/Language/Marlowe/CLI/Transaction.hs
```diff
@@ -752,7 +752,8 @@ publishImpl connection signingKey expires changeAddress publishingStrategy coinS
     coinSelectionStrategy
     printStats
 
-  void $ submitBody connection txBody [signingKey] timeout
```
paluh, line 755: I would keep here the previous semantics - the function is called `publishImpl` and not `build*Impl` - it's main purpose is to actually do the publishing ;-)

We should just skip a call on our side when we don't have a `timeout` at hand.

paluh (approved):

paluh:
Thanks a lot Larry!

PR #268, opened 2022-09-16
Add transaction constraints module

jhbertra:
Adds definitions, documentation, and API stubs for a Marlowe-specific transaction constraints module.

Future work:
  - Implement `satisfiesConstraints` @paluh
  - Implement `solveConstraints` @dino-iog
  - Implement `buildCreateConstraints` @paluh @bjornkihlberg
  - Implement `buildApplyInputsConstraints` @paluh @bjornkihlberg
  - Implement `buildWithdrawConstraints` @paluh @bjornkihlberg
  - Implement necessary support queries (fetch UTXOs for addresses, fetch application and payout script UTXOs) @jhbertra

### DIFF
--- a/marlowe-runtime/src/Language/Marlowe/Runtime/Transaction/Constraints.hs
+++ b/marlowe-runtime/src/Language/Marlowe/Runtime/Transaction/Constraints.hs
@@ -1,0 +1,262 @@
+{-# LANGUAGE GADTs #-}
+
+module Language.Marlowe.Runtime.Transaction.Constraints
+  where
+
+import qualified Cardano.Api as Cardano
+import qualified Cardano.Api.Shelley as Cardano
+import qualified Data.Aeson as Aeson
+import Data.Function (on)
+import Data.Map (Map)
+import Data.Set (Set)
+import qualified Data.Set as Set
+import Language.Marlowe.Runtime.ChainSync.Api (PaymentKeyHash)
+import qualified Language.Marlowe.Runtime.ChainSync.Api as Chain
+import qualified Language.Marlowe.Runtime.Core.Api as Core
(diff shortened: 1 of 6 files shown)


jhbertra:
@paluh, since @dino-iog is going to be working on the constraint solver, and the constraint checker is essentially intended as a test fixture for the solver, it would probably make sense for the two of you to work on it together.

marlowe-runtime/src/Language/Marlowe/Runtime/Transaction/Constraints.hs
```diff
@@ -0,0 +1,268 @@
...
+
+data InputConstraint v
+  = NoInput
+  -- ^ Specifies that the transaction consumes no marlowe or payout script
+  -- inputs. Corresponds to the create transaction type.
+  --
+  -- Rules to check:
+  --   1. All inputs of the transaction do not come from any of the script
+  --      addresses (marlowe or payout) associated with marlowe version v.
+  | MarloweInput Chain.TxOutRef P.POSIXTime P.POSIXTime (Core.Redeemer v)
```
bwbush, line 141: Should this include the datum? or will that be looked up from history?

bwbush (approved):
Looks good. I imagine we might need some small refinements during implementation, but I'm not sure of specifics.

marlowe-runtime/src/Language/Marlowe/Runtime/Transaction/Constraints.hs
jhbertra: Do we need the datum to construct the input? In principle, this should be computable from the existing output, so specifying it here may be redundant.

PR #269, opened 2022-09-17
SCP-4461 - Round Marlowe time range to slot timestamps

paluh:
When performing initialization of Marlowe contract we perform a `POSIXTime` to slot round trip so we get back timestamps which should match exactly the ones which we get on the ledger in the `ScriptContext`.

I did somewhat related minor dependency cleanup in `marlowe` - we don't use `Slot` nor `Ada` any more so I was able to drop some dep.

Pre-submit checklist:
- Branch
    - [x] Tests are provided (I slightly changed the `run-simple.sh` to test rounding)
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


marlowe-cli/src/Language/Marlowe/CLI/Run.hs
```diff
@@ -384,16 +384,34 @@ prepareTransactionImpl marloweIn txInputs minimumTime maximumTime printStats =
...
-      merkleizeInputs marloweIn transactionInput
-        `catchError` (const $ pure transactionInput)  -- TODO: Consider not catching errors here.
+    let
+      toSlot = posixTimeToEnclosingSlot mtSlotConfig
+      toSlotNo = SlotNo . fromIntegral . toSlot
+
+      roundTxInterval ti@TransactionInput{txInterval=txInterval@(minimumTime, maximumTime)} = do
+        let
+          txInterval' =
+            ( slotToBeginPOSIXTime mtSlotConfig . toSlot $ minimumTime
```
bwbush, line 400: This may work in most cases, but I'm skeptical that using `slotToBeginPOSIXTime` behaves exactly like the function that a nodes uses to convert them before passing them to the Plutus interpreter (`Cardano.Ledger.Alonzo.TxInfo.transVITime` in `cardano-ledger-alonzo`).

bwbush (approved):
Looks good, except I'm skeptical about `Plutus.V1.Ledger.Api.slotToBeginPOSIXTime`. I think we should create a backlog ticket for extensive testing of time ranges, especially on networks with non-standard slot lengths.

marlowe-cli/src/Language/Marlowe/CLI/Run.hs
paluh: So started with `transVITime` but brings here a lot of complexity which requires communication with `cardano-node`. Should I make another pass over this?

paluh: I mean it uses a lot of pieces which we use in `querySlotConfig` plus protocol parameters (only to grab the `ProtocolVersion` under the hood) so it is of course feasible ;-)

PR #271, opened 2022-09-19
Scp 4447 chainseek tx submission

jhbertra:
Adds a proxy API for node Tx submission to chainseekd. Something we might want to think about is how we want to handle node connections going forward. As it stands, `chainseekd` is becoming the node proxy for Marlowe Runtime. Is this what we want? Its name is a bit misleading if that is the case.

### DIFF
--- a/marlowe-chain-sync/app/Main.hs
+++ b/marlowe-chain-sync/app/Main.hs
@@ -1,7 +1,17 @@
+{-# LANGUAGE GADTs #-}
 module Main
   where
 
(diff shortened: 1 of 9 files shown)


bwbush (approved):
This looks correct, but I didn't try executing it.

jhbertra:
> This looks correct, but I didn't try executing it.

Bit difficult to test until the other components are ready... I suppose we will test it eventually when we integration test the whole stack!

PR #192, opened 2022-07-20
SCP-4195 - Create top level Module construct for Marlowe Extended

hrajchert:
This PR adds a top-level construct called `Module` to help in the versioned serialization of a contract. The `Module` construct currently bundles the metadata together with the code, and could be a potential place to add `imports/exports` and functions.

The PR is only covering the `Language.x` changes to avoid conflicts with the PR @paluh is doing. We will later need to change the UI that currently uses the global metadata object to work with the `Module`.

Pre-submit checklist:
- Branch
    - [ ] Tests are provided (if possible)
    - [X] Commit sequence broadly makes sense
    - [X] Key commits have useful messages
    - [X] Relevant tickets are mentioned in commit messages
    - [X] Formatting, materialized Nix files, PNG optimization, etc. are updated
- PR
    - [X] Self-reviewed the diff
    - [X] Useful pull request description
    - [X] Reviewer requested

PR #274, opened 2022-09-20
SCP-4319 Create top level Module construct in Extended

hrajchert:
This PR is the same as [PR-192](https://github.com/input-output-hk/marlowe-cardano/pull/192) but in Haskell instead of PureScript. 

It adds a top-level construct in Marlowe Extended called `Module` that helps with adding a version number in the Json serialization and its the candidate to extend to add `imports/exports` in the future (hence the name), and also functions that can be converted to merkelized Marlowe Core.

This PR is related to the following MIP discussion: https://github.com/input-output-hk/MIPs/discussions/7 

I've also fixed the serialization of timeout to match the PureScript counterpart.
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marlowe/src/Language/Marlowe/Extended/V1.hs
+++ b/marlowe/src/Language/Marlowe/Extended/V1.hs
@@ -43,6 +43,7 @@ import Data.Ratio ((%))
 import Data.Text (pack)
 import GHC.Generics
 import qualified Language.Marlowe.Core.V1.Semantics.Types as S
+import Language.Marlowe.Extended.V1.Metadata.Types (MetaData)
 import Language.Marlowe.ParserUtil (getInteger, withInteger)
 import Language.Marlowe.Pretty (Pretty(..), pretty)
 import Language.Marlowe.Util (ada)
@@ -53,8 +54,7 @@ import Text.PrettyPrint.Leijen (parens, text)
 
 data Timeout = TimeParam String
              | POSIXTime Integer
-  deriving stock (Eq, Show, Generic)
-  deriving anyclass (FromJSON, ToJSON)
+  deriving stock (Show,Generic,Eq)
 
 instance Pretty Timeout where
     prettyFragment (POSIXTime n)    = prettyFragment n
@@ -91,7 +91,7 @@ data Value = AvailableMoney S.AccountId S.Token
            | TimeIntervalEnd
            | UseValue S.ValueId
            | Cond Observation Value Value
-  deriving stock (Show,Generic)
+  deriving stock (Show,Generic,Eq)
   deriving anyclass (Pretty)
 
 data Observation = AndObs Observation Observation
@@ -105,22 +105,22 @@ data Observation = AndObs Observation Observation
                  | ValueEQ Value Value
                  | TrueObs
                  | FalseObs
-  deriving stock (Show,Generic)
+  deriving stock (Show,Generic,Eq)
   deriving anyclass (Pretty)
 
 data Action = Deposit S.AccountId S.Party S.Token Value
             | Choice S.ChoiceId [S.Bound]
             | Notify Observation
-  deriving stock (Show,Generic)
+  deriving stock (Show,Generic,Eq)
   deriving anyclass (Pretty)
 
 data Payee = Account S.AccountId
            | Party S.Party
-  deriving stock (Show,Generic)
+  deriving stock (Show,Generic,Eq)
   deriving anyclass (Pretty)
 
 data Case = Case Action Contract
-  deriving stock (Show,Generic)
+  deriving stock (Show,Generic,Eq)
   deriving anyclass (Pretty)
 
 data Contract = Close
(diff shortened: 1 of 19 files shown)


marlowe/src/Language/Marlowe/Extended/V1/Metadata/Types.hs
```diff
@@ -0,0 +1,130 @@
...
+import qualified Plutus.V1.Ledger.Value as Val
+
+data ContractType
+  = Escrow
+  | EscrowWithCollateral
+  | ZeroCouponBond
+  | CouponBondGuaranteed
+  | Swap
+  | ContractForDifferences
+  | Other
```
bwbush, line 32: Would it make sense for `Other` to be `Other String` so that the "other" type of contract could be named?

marlowe/src/Language/Marlowe/Extended/V1/Metadata/Types.hs
```diff
@@ -0,0 +1,130 @@
...
+import Data.Map.Ordered (OMap)
+import qualified Data.Map.Ordered as OMap
+import Data.Set (Set)
+import Data.Set.Ordered (OSet)
+import Data.Text.Encoding as Text
+import qualified GHC.Generics
+import Plutus.V1.Ledger.Api (TokenName)
+import qualified Plutus.V1.Ledger.Value as Val
+
+data ContractType
```
bwbush, line 25: This isn't much of a taxonomy. Do we really need a restrictive `ContractType` like this?

marlowe/src/Language/Marlowe/Extended/V1.hs
```diff
@@ -129,9 +129,30 @@ data Contract = Close
...
+                     --   https://github.com/input-output-hk/MIPs/discussions/7
+                     --   Version 1 of Marlowe Extended shouldn't be closed until the discussion
+                     --   is resolved.
+                     , contract :: Contract
+                     -- ^ Currently we have a single entrypoint for the full contract, but we could
+                     --   extend this to be a `OMap Identifier Contract` to enable functions.
+                     --   In order to guarantee that we can convert to Marlowe Core we would need to
+                     --   validate that a function can only call functions defined above (if bottom-up)
+                     --   and that they cannot call themselves
+                     }
```
bwbush, line 153: Should we consider storing the merkleized continuations here by adding `continuations :: Map ContractHash Contract`? I realize that Extended Marlowe support merkleization, but it will be unusable for non-trivial contracts if it doesn't. Maybe now is the time to design an approach for handling merkleization of Extended Marlowe?

marlowe/test/Spec/Marlowe/Serialization/ExtendedJson.hs
```diff
@@ -0,0 +1,50 @@
...
+    , testCase "Golden Swap module" testGoldenSwapModule
+    ]
+
+-- TODO: Do a small non-property round-trip with example contracts.
+
+
+-- | Checks that we can decode the Golden JSON Contract for Swap
+testGoldenSwapContract :: IO ()
+testGoldenSwapContract = do
+    filePath <- (\d -> d ++ "/test/Spec/Marlowe/Serialization/golden/swap-contract.json") <$> getCurrentDirectory
```
bwbush, line 36: You don't really need to prepend the current directory and `/`, since `eitherDecodeFileStrict` will read a path relative the current directory anyway. Also the `filepath` package provides OS-independent path navigation.

marlowe/test/Spec/Marlowe/Serialization/ExtendedJson.hs
```diff
@@ -0,0 +1,50 @@
...
+    mContract <- eitherDecodeFileStrict filePath
+    case mContract of
+        Left err              -> assertFailure err
+        Right (_ :: Contract) -> return ()
+
+-- | Checks that we can decode the Golden JSON Module for Swap
+-- | TODO: If we are more of these tests, add a helper function with a Proxy type
+testGoldenSwapModule :: IO ()
+testGoldenSwapModule = do
+    filePath <- (\d -> d ++ "/test/Spec/Marlowe/Serialization/golden/swap-module.json") <$> getCurrentDirectory
```
bwbush, line 46: ditto

marlowe/test/Spec/Marlowe/Serialization/ExtendedJson.hs
```diff
@@ -0,0 +1,50 @@
...
+
+import Data.Aeson (eitherDecodeFileStrict)
+import Language.Marlowe.Extended.V1 (Contract, Module)
+import System.Directory (getCurrentDirectory)
+import Test.Tasty (TestTree, testGroup)
+import Test.Tasty.HUnit (assertFailure, testCase)
+
+tests :: TestTree
+tests = testGroup "Extended Contract Serialization"
+    [ testCase "Golden Swap contract" testGoldenSwapContract
```
bwbush, line 26: Instead of hard-coding each golden test, could we have the test code discover all of the JSON files in the golden directory and then test each of them?

bwbush (approved):
Many of my comments might be for later, not for this PR.

marlowe/src/Language/Marlowe/Extended/V1/Metadata/Types.hs
hrajchert: Totally... I've set it the same as it is in PureScript for compatibility purposes, but I'd like to get rid of this field altogether (or find better taxonomy as you mentioned later). But I'd like that to happen after we resolve this discussion:
https://github.com/input-output-hk/MIPs/discussions/7

marlowe/src/Language/Marlowe/Extended/V1/Metadata/Types.hs
hrajchert: same comment as above

marlowe/src/Language/Marlowe/Extended/V1.hs
hrajchert: I was thinking of Extended to work with the concept of functions and Core to work with the concept of merkle trees. That would make Extended more usable, right now the examples would weight the same as the unmerkelized Core.

marlowe/test/Spec/Marlowe/Serialization/ExtendedJson.hs
hrajchert: I've added a comment to use a `Proxy` if we find ourselves doing more than these 2 tests. But I'm not sure I'm able just to do a smart discovery because we need to provide the type that we are trying to decode for each file.

marlowe/test/Spec/Marlowe/Serialization/ExtendedJson.hs
hrajchert: done

PR #275, opened 2022-09-21
SCP4448 Runtime Tx Submission

jhbertra:
Adds an implementation for the `SubmitJob` that delegates to chain seek (both for submitting the tx and awaiting the block).


bwbush (approved):
I haven't tried executing this, but it looks good.

PR #276, opened 2022-09-21
Update constraints

jhbertra:
I've updated the constraints module based on the changes I discussed with @bwbush and @dino-iog. Notably, the policyID for minted role tokens must be computed when solving the constraints. This means that `buildCreateConstraints` will need to do some preliminary coin selection to get a UTXO that can be used to compute the policyID. In addition, the min UTXO requirement for the contract must also be performed by `buildCreateConstraints` and reflected in the `SendToMarloweScript` constraint (in both the `Assets` and the `accounts` of the `Datum`).


bwbush (approved):
Looks good.

paluh (approved):

PR #285, opened 2022-09-30
Disable mainnet

jhbertra:
Disabling tx creation on mainnet for the time being, pending auditor approval of the core infrastructure.


bwbush (approved):
Perfect!

PR #289, opened 2022-10-03
Contract Discovery

jhbertra:
- Adds contract discovery implementation.
- Also fixes numerous bugs in the current CLI commands and History module.


bwbush (approved):
Looks great!

PR #304, opened 2022-10-14
SCP-4519 Fix so chain seek returns only unspent UTxOs

bwbush:
The query for utxos was returning all utxos, not just the unspent ones because there was no left join against the tx in table, with a filter for null on the right side of the join. This commit adds the join.

I tested using the `marlowe-runtime/examples/create.ipynb` notebook and debug statements to verify that `chainseekd` now returns only unspent outputs.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


paluh (approved):
Nice catch. Thanks a lot!

PR #317, opened 2022-10-20
Added a create test with many utxos in the party addresses

dino-iog:
This is more than one utxo in each but all with ADA only.

The test script was made from a duplicate of the `create` Jupyter notebook with some modifications.
We can add a script to run several together if we want too like `marlowe-cli/examples` has.

### DIFF
--- a/marlowe-cli/flake.lock
+++ b/marlowe-cli/flake.lock
@@ -3119,13 +3119,12 @@
         "sphinxcontrib-haddock": "sphinxcontrib-haddock"
       },
       "locked": {
-        "lastModified": 0,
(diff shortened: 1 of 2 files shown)


bwbush (changes requested):
The output of Cell 22 shows that there is only one UTxO at the address used to create the contract. How is this a "test with many utxos in the party address"?

dino-iog:
@bwbush I had made a mistake and accidentally pulled another test's notebook page into this branch. I got rid of the commits that brought in `create-close-native-tok.ipynb` and fixed the correct notebook `create-many-utxos.ipynb`

dino-iog:
Just rebased this against `main` now that #335 has been merged. Should be fine after the checks pass.

PR #335, opened 2022-10-28
SCP-4598 SCP-4604 SCP-4625 SCP-4626 Bug fixes to enable comprehensive Marlowe Runtime contract

bwbush:
- SCP-4598 creates a comprehensive contract for Marlowe Runtime testing
- SCP-4604 fixes coin selection when native assets are at the script address. Coin selection was not accounting for the value coming from the script address because that was not part of the wallet context. I added the marlowe context to the UTxO set so that coin selection could look up the value consumed from the script.
- SCP-4625 prevents coin selection from erroneously attempting to spend reference inputs.
- SCP-4626 fixes the incorrect redeemer that was used for payouts.

All together, these eliminate outstanding bugs that prevented [the comprehensive test contract](https://github.com/input-output-hk/marlowe-cardano/blob/SCP-4604/marlowe-runtime/examples/token-bid.ipynb) from being successfully executed with Marlowe Runtime.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


bwbush:
Sorry, here are more details:
1. selectCoins builds a UTxO map from the wallet context only: https://github.com/input-output-hk/marlowe-cardano/blob/7b67fc7cc4646d3eb06c6ea1c7e853ddef56c3a8/marlowe-runtime/src/Language/Marlowe/Runtime/Transaction/Constraints.hs#L536.
2. The incoming value (from the constraints) is computed from the UTxO map: https://github.com/input-output-hk/marlowe-cardano/blob/7b67fc7cc4646d3eb06c6ea1c7e853ddef56c3a8/marlowe-runtime/src/Language/Marlowe/Runtime/Transaction/Constraints.hs#L562.
3. However, if a script UTxO is being spent, then it is providing value, too. This is ignored in the inputsFromBody computation because the filter (flip elem inputTxIns . fst) will not find those script inputs, since there weren't included in the UTxO set.
4. This problem also had existed in balanceTx, but it was fixed at https://github.com/input-output-hk/marlowe-cardano/blob/7b67fc7cc4646d3eb06c6ea1c7e853ddef56c3a8/marlowe-runtime/src/Language/Marlowe/Runtime/Transaction/Constraints.hs#L798-L800. That code is almost exactly what is needed in selectCoins, except of the TxOut context.
5. I consolidated the code embedded into balanceTx into a top-level function allUtxos that balanceTx could still use and that selectCoins would be able to use so that it has both the marlowe and wallet contexts.

bwbush:
> However, if a script UTxO is being spent, then it is providing value, too.

We didn't see this error in cases where only lovelace was at the script address, since there is plenty of lovelace around for balancing. However, we do see the error if the script holds native tokens that the wallet doesn't, since then the script is the only source of the tokens required by the output constraints.

dino-iog (approved):

dino-iog:
Ok with me to merge this.

paluh:
Code changes looks ok - I reviewed them on Friday. I'm going test the notebooks on my branch later no.
Thanks a lot @bwbush!

PR #339, opened 2022-11-01
Scp 4569 slow utxo query

jhbertra:
Query was slow because it was filtering on the non-indexed `address` field. The index is on the MD5 hash of the address instead. Updated the `WHERE` clause of the query in question to take advantage of this, and added a comment to the file about this very non-obvious requirement for querying by address.

### DIFF
--- a/marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
+++ b/marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
@@ -5,6 +5,25 @@
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE QuasiQuotes #-}
 
+{- IMPORTANT PERFORMANCE WARNING - non-obvious requirement for filtering on addresses
+
+   If you want to write a query that filters on chain.txOut.address, be aware
+   that the database does not index the address column directly (because of some
+   Byron addresses that are prohibitively large). Instead, it indexes an MD5 hash
+   of the address.
+
+   This means that if you want your query to be at all performant, it is not
+   enough to say, for example
+
+    WHERE txOut.address = myAddress
+
+   you have to do something like this:
+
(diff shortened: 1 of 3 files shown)


nix/dev/compose.nix
```diff
@@ -98,7 +98,7 @@ let
       # TODO Connect volumes here to top-level volumes
       volumes = [
         "postgres:/var/lib/postgresql/data"
-        "./postgres/init.sql:/docker-entrypoint-initdb.d/init.sql:r"
+        "./postgres/init.sql:/docker-entrypoint-initdb.d/init.sql"
```
jhbertra, line 101: This broke with `podman`

marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
```diff
@@ -792,11 +814,11 @@ getUTxOs = GetUTxOs $ fmap Api.UTxOs . \case
              , asset.name :: bytea?
              , assetOut.quantity :: bigint?
           FROM chain.txOut         AS txOut
+          JOIN addresses           AS addr     ON addr.address = txOut.address AND CAST(MD5(addr.address) AS uuid) = CAST(MD5(txOut.address) AS uuid)
```
paluh, line 817: Out of curiosity - do this `JOIN` fail if we skip direct address equality check `addr.address = txOut.address`?

paluh (approved):

marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Database/PostgreSQL.hs
jhbertra: only if there is an MD5 hash collision. It's not a significant amount of overhead to add this recheck condition to the filter, the important thing is that it does a bitmap heap scan instead of a seq scan to produce the initial result set.

PR #356, opened 2022-11-16
SCP-4679 Unit tests for `Language.Marlowe.Runtime.Transaction.Constraints.solveInitialTxBodyContent`

jhbertra:


marlowe-runtime/src/Language/Marlowe/Runtime/Transaction/Constraints.hs
```diff
@@ -199,27 +199,10 @@ mustSendMerkleizedContinuationOutput contract =
...
--- | Get the total amount required to be paid to the given role.
-getTotalForRole
-  :: forall v
-   . Core.IsMarloweVersion v
-  => Core.PayoutDatum v
-  -> TxConstraints v
-  -> Chain.Assets
-getTotalForRole role = case Core.marloweVersion @v of
-  Core.MarloweV1 -> fromMaybe mempty . Map.lookup role . payToRoles
-
```
jhbertra, line 222: These were unused

marlowe-chain-sync/src/Language/Marlowe/Runtime/Cardano/Api.hs
```diff
@@ -251,7 +251,7 @@ assetsToCardanoValue Assets{..} = fold <$> sequence
 
 assetsFromCardanoValue :: C.Value -> Assets
 assetsFromCardanoValue value = Assets
-  { ada = maybe 0 fromCardanoLovelace $ C.valueToLovelace value
+  { ada = fromCardanoLovelace $ C.selectLovelace value
```
jhbertra, line 254: Caught a bug! `valueToLovelace` returns `Just` if and only if the `value` is ADA-only.

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/ConstraintsSpec.hs
```diff
@@ -0,0 +1,734 @@
...
+            matches (TxOut outAddress (TxOutValue MultiAssetInBabbageEra value) _ _)
+              | selectAsset value cardanoAssetId > 0 = Just (outAddress, value)
+              | otherwise = Nothing
+            matches (TxOut _ (TxOutAdaOnly era _) _ _) = case era of
+          let matchingOuts = mapMaybe matches txOuts
+          case matchingOuts of
+            [(outAddress, value)] -> do
+              fold
+                [ check
+                    (fmap fst (valueToList value) == [cardanoAssetId])
```
bwbush, line 137: Is this really needed? Why can't the output contain irrelevant tokens? The coin-selection and balancing algorithms could (in principle) decide to consolidate outputs to the same address into a single output. Irrelevant tokens might be in that output if they had to be consumed in input. In practice, the coin selection and balancing algorithms don't do this, but a future version might.

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/ConstraintsSpec.hs
```diff
@@ -0,0 +1,734 @@
...
+  -> [String]
+mustSendMarloweOutputViolations MarloweV1 MarloweContext{..} TxConstraints{..} TxBodyContent{..} =
+  case marloweOutputConstraints of
+    MarloweOutputConstraintsNone -> []
+    MarloweOutput assets datum ->
+      case find ((== marloweAddress) . extractAddress) txOuts of
+        Nothing -> ["No output found to Marlowe address"]
+        Just txOut -> fold
+          [ check (extractValue txOut == assets) "Wrong assets sent to Marlowe Address."
+          , check (extractDatum txOut == Just (toChainDatum MarloweV1 datum)) "Wrong assets sent to Marlowe Address."
```
bwbush, line 219: Shouldn't this error message mention the datum, perhaps "wrong datum sent for Marlowe Address"?

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/ConstraintsSpec.hs
```diff
@@ -0,0 +1,734 @@
...
+                   && extractDatum txOut == Just (toChainPayoutDatum MarloweV1 roleToken)
+      matchingOutput = find isMatch txOuts
+    case matchingOutput of
+      Nothing -> ["No matching output found to payout validator"]
+      Just txOut ->
+        let
+          totalToRole = extractValue txOut
+        in
+          check
+            (totalToRole == assets)
```
bwbush, line 252: Should this be an inequality? What about the case when the lovelace in the output is rounded up to satisfy min-UTxO?

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/ConstraintsSpec.hs
```diff
@@ -0,0 +1,734 @@
...
+  -> [String]
+requiresSignatureViolations MarloweV1 utxos TxConstraints{..} TxBodyContent{..} = do
+  pkh <- Set.toList signatureConstraints
+  let
+    inInput ref = case Chain.lookupUTxO ref utxos of
+      Nothing -> False
+      Just Chain.TransactionOutput{..} -> case Chain.paymentCredential address of
+        Just (Chain.PaymentKeyCredential pkh') -> pkh' == pkh
+        _ -> False
+    inExtraKeyWits = case txExtraKeyWits of
```
bwbush, line 326: Technically, one doesn't need an extra key witness if the same key witness is already needed to spend one of the UTxOs in the transaction. Including this extra "extra key witness" is harmless in these cases, aside from increasing the transaction size a bit. I don't think we need to change anything here.

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/ConstraintsSpec.hs
```diff
@@ -0,0 +1,734 @@
...
+
+genRoleToken :: Gen Chain.AssetId
+genRoleToken = Chain.AssetId
+  <$> (hedgehog $ fromCardanoPolicyId . PolicyId <$> genScriptHash)
+  <*> genRole
+
+-- NOTE just a random list of names that won't conflict with anything generated
+-- by Gen.Cardano.Api.Typed
+genRole :: Gen Chain.TokenName
+genRole = elements
```
bwbush, line 606: Someday we should unify the `Gen` instances here (and in other packages) with those in `marlowe-test`, so we have less code to maintain.

bwbush (approved):
I made a few minor comments about technicalities, but this unit test seems thorough and accurate.

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/ConstraintsSpec.hs
jhbertra: This is checking the output of the constraint solver only (before coin selection and balancing). It's not strictly needed, but it does enforce that the outputs are created cleanly (the constraint solver could decide to aggregate payouts and role tokens, for example, but having those separate makes consuming them easier).

I'm certainly open to a debate on this point however.

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/ConstraintsSpec.hs
jhbertra: Copy :spaghetti: mistakes...

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/ConstraintsSpec.hs
jhbertra: `solveInitialTxBody` is run pre-minUTxO adjustment (That's why we wanted to write these tests for that function rather than `solveConstraints`, so we would have easier invariants to check).

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/ConstraintsSpec.hs
jhbertra: I'm actually checking `inExtraKeyWits || inInputs` on line 331 below

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/ConstraintsSpec.hs
jhbertra: Somehow, yes - although it's a bit tricky because context matters (in the case of Chain.TokenName, it could be a random token or a role token)

PR #357, opened 2022-11-17
SCP-4705 GET /contracts/:contractId

jhbertra:


marlowe-runtime/web-server-test/Spec.hs
```diff
@@ -46,15 +67,18 @@ instance Arbitrary Web.PolicyId where
...
   arbitrary = pure $ Web.Metadata Null
 
-instance Arbitrary Web.TxStatusHeader where
+instance Arbitrary Web.TxStatus where
   arbitrary = oneof
     [ pure Web.Unsigned
     , pure Web.Submitted
-    , Web.Confirmed <$> arbitrary
+    , pure Web.Confirmed
     ]
```
bjornkihlberg, line 75: I don't see this as a blocker but here's a slightly simplified version:
```haskell
arbritrary = elements
  [ Web.Unsigned
  , Web.Submitted
  , Web.Confirmed
  ]
```
[`elements :: [a] -> Gen a`](https://hackage.haskell.org/package/QuickCheck-2.14.2/docs/Test-QuickCheck.html#v:elements)

marlowe-runtime/web-server/Language/Marlowe/Runtime/Web/Server.hs
```diff
@@ -79,12 +81,14 @@ compile $ SelectorSpec "server"
   , "http" ≔ Inject ''ServeRequest
   , "api" ≔ Inject ''ApiSelector
   , ["contract", "indexer"] ≔ Inject ''ContractHeaderIndexerSelector
+  , ["history"] ≔ Inject ''HistoryClientSelector
```
bjornkihlberg, line 84: This is new syntax to me. Is this part of the `DataKinds` extension? Is it representing a type as a value?

marlowe-runtime/web/Language/Marlowe/Runtime/Web/Orphans.hs
```diff
@@ -0,0 +1,415 @@
+{-# LANGUAGE OverloadedLists #-}
+
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Language.Marlowe.Runtime.Web.Orphans
```
bjornkihlberg, line 5: Is it possible to unit test this module such that a programmer who isn't familiar with this code (eg us in three months 😉) won't leave the schema out of date?

marlowe-runtime/web-server-test/Spec.hs
jhbertra: Thanks, well spotted

marlowe-runtime/web/Language/Marlowe/Runtime/Web/Orphans.hs
jhbertra: Yes - all the schemas are tested in `web-server-test`

PR #365, opened 2022-11-25
SCP-4715

bjornkihlberg:
DONE Gather pieces
DOING Put pieces together
TODO Clean up code

### DIFF
--- a/marlowe-runtime/.gitignore
+++ b/marlowe-runtime/.gitignore
@@ -1,0 +1,1 @@
+client_session_key.aes
--- a/marlowe-runtime/cip30-demo/Main.hs
+++ b/marlowe-runtime/cip30-demo/Main.hs
@@ -1,14 +1,30 @@
 {-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TypeFamilies #-}
 
 module Main
   where
 
+import Cardano.Api (BabbageEra)
+import Control.Exception (throwIO)
+import Data.Aeson ((.:))
+import qualified Data.Aeson as Aeson
+import Data.Text (Text)
+import qualified Language.Marlowe.Runtime.ChainSync.Api as ChainSync.Api
+import qualified Language.Marlowe.Runtime.Core.Api as Core.Api
+import qualified Language.Marlowe.Runtime.Transaction.Api as Transaction.Api
+import Network.Protocol.Driver (RunClient, runClientPeerOverSocket)
+import Network.Protocol.Job.Client (JobClient, jobClientPeer)
+import qualified Network.Protocol.Job.Client as JobClient
+import Network.Protocol.Job.Codec (codecJob)
+import Network.Socket (SocketType(..))
+import qualified Network.Socket as Socket
 import qualified Text.Blaze as Blaze
 import Text.Julius (juliusFile)
 import Yesod (Yesod)
 import qualified Yesod
+import qualified Yesod.Core.Types
 
 data CIP30 = CIP30
 
@@ -21,9 +37,22 @@ newtype WalletName = WalletName String
     , Blaze.ToMarkup
     )
 
+newtype PostWalletCreateRequestDTO = PostWalletCreateRequestDTO { source :: Text }
+  deriving Show
+
+data PostWalletCreateResponseDTO = PostWalletCreateResponseDTO
+  { unsignedTransaction :: Text
+  , contractId :: Text
+  }
+
+instance Aeson.FromJSON PostWalletCreateRequestDTO where
+  parseJSON = Aeson.withObject "PostWalletCreateRequestDTO" \o ->
+    PostWalletCreateRequestDTO <$> o .: "source"
+
 Yesod.mkYesod "CIP30" [Yesod.parseRoutes|
-/            HomeR   GET
-/#WalletName WalletR GET
+/                   HomeR         GET
+/#WalletName        WalletR       GET
+/#WalletName/create WalletCreateR GET POST
 |]
 
 instance Yesod CIP30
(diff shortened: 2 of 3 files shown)


marlowe-runtime/.gitignore
```diff
@@ -0,0 +1 @@
+client_session_key.aes
```
jhbertra, line 1: 3 Comments:

1. Please don't add per-directory .gitignores (I know we have them elsewhere, but this is a legacy thing and should be cleaned up and moved to the root .gitignore)
2. What was this for?
3. For keeping local key files in repo, we already have a `treasury/` entry in our root .gitignore. If you need to keep a key somewhere in the repo hierarchy, make a `treasury` folder in the repo root and put your private stuff in there.

marlowe-runtime/cip30-demo/Main.hs
```diff
@@ -1,14 +1,30 @@
 {-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE RankNTypes #-}
```
jhbertra, line 2: I don't see anything in here that needs RanNTypes - is this because of TH code that Yesod is generating?

marlowe-runtime/cip30-demo/Main.hs
```diff
@@ -21,9 +37,22 @@ newtype WalletName = WalletName String
...
+  deriving Show
+
+data PostWalletCreateResponseDTO = PostWalletCreateResponseDTO
+  { unsignedTransaction :: Text
+  , contractId :: Text
+  }
+
+instance Aeson.FromJSON PostWalletCreateRequestDTO where
+  parseJSON = Aeson.withObject "PostWalletCreateRequestDTO" \o ->
+    PostWalletCreateRequestDTO <$> o .: "source"
```
jhbertra, line 50: If you derive `Generic` for `PostWalletCreateRequestDTO` you can omit this `parseJSON` definition.

marlowe-runtime/cip30-demo/Main.hs
```diff
@@ -21,9 +37,22 @@ newtype WalletName = WalletName String
     , Blaze.ToMarkup
     )
 
+newtype PostWalletCreateRequestDTO = PostWalletCreateRequestDTO { source :: Text }
```
jhbertra, line 40: You'll need more information besides the contract source, and the contract can be a stronger type than this (assuming we `JSON.parse` on the frontend before adding it to the `fetch` request). Check out `PostContractsRequest` in `L.M.R.Web.Types` to see what is needed. Additionally, we need to include the change address, extra addresses, and collateral UTxOs from the CIP-30 wallet in this request. The web server passes these via headers, but you can just put them in the request body here.

marlowe-runtime/cip30-demo/Main.hs
```diff
@@ -54,6 +83,70 @@ getWalletR walletName = Yesod.defaultLayout do
...
+    <input id="contractSource" type="text">
+    <button id="postContractSource">click
+    <script>
+      const contractSource = document.getElementById('contractSource')
+      const postContractSource = document.getElementById('postContractSource')
+      postContractSource.addEventListener("click", async () => {
+        const options = {
+          method: "POST",
+          headers: { "Content-Type": "application/json;charset=utf-8" },
+          body: JSON.stringify({ source: contractSource.value })
```
jhbertra, line 98: As mentioned above, let's parse the contract source here into JSON

```suggestion
          body: JSON.stringify({ source: JSON.parse(contractSource.value) })
```

jhbertra (commented):
Did a preliminary review of stuff so far

marlowe-runtime/.gitignore
bjornkihlberg: No problems! It seems to be generated by yesod. I could investigate it a little deeper to understand what it is.

marlowe-runtime/cip30-demo/Main.hs
bjornkihlberg: It's because of `RunClient` in
```haskell
runTxJobClient :: Socket.HostName -> Socket.PortNumber -> RunClient IO (JobClient Transaction.Api.MarloweTxCommand)
```
If I go with the same structure you did where it's wrapped in a `newtype` then it also needs the `RankNTypes` extension to work.

marlowe-runtime/cip30-demo/Main.hs
bjornkihlberg: I absolutely agree, the `createContract` procedure will force me to provide more data and I agree that it can be cleaned up and architectured better.

marlowe-runtime/cip30-demo/Main.hs
bjornkihlberg: Right, it's just a temporary thing. I will later add more fields to `PostWalletCreateRequestDTO` and then it will be derived correctly. I don't think it's treated as a record with only one field which means that I will need to do some rework in the inlined JavaScipt if I derive it now and I'd rather have the compiler help me later.

marlowe-runtime/.gitignore
bjornkihlberg: Fixed it with the advice given in [this stackoverflow thread](https://stackoverflow.com/questions/43395944/prevent-yesod-from-generating-client-session-key-aes-file).

marlowe-runtime/cip30-demo/Main.hs
bjornkihlberg: Do we know that the source will be given in JSON?

marlowe-runtime/cip30-demo/Main.hs
```diff
@@ -21,12 +38,25 @@ newtype WalletName = WalletName String
     , Blaze.ToMarkup
     )
 
+data PostWalletCreateRequestDTO = PostWalletCreateRequestDTO
+  { version :: Core.Api.SomeMarloweVersion
```
jhbertra, line 42: Remove this field and hard-code MarloweV1

marlowe-runtime/cip30-demo/Main.hs
```diff
@@ -21,12 +38,25 @@ newtype WalletName = WalletName String
...
+  , source :: Aeson.Value
```
jhbertra, line 43: Change this to `contract :: Contract` and instead of setting this to the text value of the form, call `JSON.parse` and assign the resulting object tot it.

PR #374, opened 2022-11-30
SCP-4687 Fix default validity interval calculations in marlowe-tx

jhbertra:
- [x] Update default calculation of `invalidBefore` in `buildApplyInputsConstraints` to use the current tip slot of the local node instead of the current time.
- [x] Update default calculation of `invalidHereafter` in `buildApplyInputsConstraints` to take the min value of the next contract timeout and the maximum slot boundary of the current era history.
- [x] Bonus: Fix slot conversion in HistoryFollower (And eradicate the misguided `SlotConfig` type from codebase).
- [x] Bonus: Fix bug where the validity lower bound was being saved instead of the validity upper bound in `chainseekd` :fearful: 

:rotating_light: NOTE RELATED TO LAST ITEM: Please drop and recreate your `chain seek` databases, as the validityIntervalUpperBound is incorrect on all transactions. You will need to re-index the chain from genesis to fix this. :rotating_light:


marlowe-runtime/src/Language/Marlowe/Runtime/Transaction/BuildConstraints.hs
```diff
@@ -287,11 +303,19 @@ buildApplyInputsConstraintsV1 systemStart eraHistory marloweOutput invalidBefore
     requiredParties = Set.fromList $ for redeemer $ marloweInputParty >>> maybeToList
     roleAssetId = toAssetId currencySymbol
 
-  invalidBefore' <- lift $ utcTimeToSlotNo invalidBefore
+    tipSlot' = toCardanoSlotNo tipSlot
+
+  invalidBefore' <- lift $ maybe (pure tipSlot') utcTimeToSlotNo invalidBefore
```
bwbush, line 308: We'll need extensive testing of Runtime's time-to-slot conversion, to make sure it is consistent with the validator's https://github.com/input-output-hk/marlowe-cardano/blob/92916702040c995ac51ee541fb6c28a382102c65/marlowe/src/Language/Marlowe/Scripts.hs#L159-L168.

We've had problems in the past where a small roundoff in the time interval propagates into the `minTime` in the datum differently in the on- and off-chain code, with the result that the transaction fails due to the on- and off-chain outputs having different `minTime`.

The code here looks correct, but it's hard enough to reason about this that we'll need test cases. I'll create a Jira ticket.

marlowe-runtime/src/Language/Marlowe/Runtime/Transaction/BuildConstraints.hs
```diff
@@ -396,11 +423,34 @@ buildApplyInputsConstraintsV1 systemStart eraHistory marloweOutput invalidBefore
...
+    nextMarloweTimeout (V1.Let _ _ c) = nextMarloweTimeout c
+    nextMarloweTimeout (V1.Assert _ c) = nextMarloweTimeout c
+
+    maxSafeSlot :: O.SlotNo
+    maxSafeSlot = getMaxSafeSlotFromSummary $ unInterpreter interpreter
+
+    unInterpreter :: Interpreter xs -> Summary xs
+    unInterpreter = unsafeCoerce -- Interpreter xs is a newtype wrapper around Summary xs.
+                                 -- Unfortunately, the provided query language is not able to extract the max safe slot,
+                                 -- So we have to reach unsafely into the internals to get
```
bwbush, line 438: Should we submit a ticket to the ouroboros repo?

bwbush (approved):
This is more robust and cleaner that the previous use of slot configurations.

I have a lingering concern about roundoff causing inconsistency between the off- and on-chain code, though I cannot point to any location in our codebase where this might be a problem. I created [SCP-4755](https://input-output.atlassian.net/browse/SCP-4755) for on-chain testing.

marlowe-runtime/src/Language/Marlowe/Runtime/Transaction/BuildConstraints.hs
jhbertra: Yes, perhaps that would be a good idea.

PR #386, opened 2022-12-08
SCP-4266 chainseekd logging

jhbertra:
- [x] adds JSON to stderr logging to chainseekd
- [x] adds detailed log configuration (hot-loaded at runtime :sunglasses:)


marlowe-chain-sync/src/Language/Marlowe/Runtime/Cardano/Api.hs
```diff
@@ -99,14 +99,6 @@ fromCardanoScriptData = \case
   C.ScriptDataNumber i -> I i
   C.ScriptDataBytes bs -> B bs
 
-toCardanoMetadata :: Metadata -> C.TxMetadataValue
```
bwbush, line 102: Is there an equivalent capability that provides this functionality? or is this truly not needed?  Clients will need to both read and write metadata.

bwbush (approved):
Thanks for all of the `ToJSON` instances! I've tested this on `preview` and `preprod`, and it seems to work fine.

PR #390, opened 2022-12-12
Add marlowe runtime apps to flake outputs

jhbertra:
- [x] Adds MRT components as `apps` to the Flake outputs
- [x] Updates documentation that previously referred to running these in a nix shell to run them using `nix run`


bwbush (approved):
Excellent!

PR #395, opened 2022-12-13
SCP-4816 Split chainseek into writer and reader processes

jhbertra:
Splits chainseekd into two processes - an indexer and a chain seek server. This is to support multiple chain seek servers per postgres database.


jhbertra:
Those are good questions. Here are my answers:

> Are there any concerns if the indexer and chainseekd are talking to different node instances?

No, that should actually be fine. The submission pipeline and the synchronization pipeline do not depend on one another. Obviously, it would be preferable if they were the same node instance though...

> Could there be some kind of cache incoherence if the database is significantly out of sync with the node of my chainseekd instance?

Yes, it would be like interacting with a node that is significantly behind the rest of the network. You pretty much need to wait for it to catch up if you want to submit a transaction or want up-to-date information.

marlowe-chain-sync/marlowe-chain-indexer/Main.hs
```diff
@@ -0,0 +1,83 @@
...
+import Language.Marlowe.Runtime.ChainIndexer.Genesis (computeByronGenesisBlock)
+import Logging (RootSelector(..), getRootSelectorConfig)
+import Observe.Event (narrowEventBackend)
+import Observe.Event.Backend (newOnceFlagMVar)
+import Observe.Event.Component (LoggerDependencies(..), logger)
+import Options (Options(..), getOptions)
+import System.IO (stderr)
+
+main :: IO ()
+main = run =<< getOptions "0.0.0.0"
```
bwbush, line 37: We should grab this from the cabal version.

jhbertra:
> In retrospect, we should have included this split in the original design.

I recall that I was in favour of doing this initially, but there was pushback based on not complicating the architecture :wink:

PR #401, opened 2022-12-15
SCP-4824 Enhanced Metadata Support

jhbertra:
- [x] adds support for transaction metadata in `apply` commands in CLI and `POST /contracts/{contract-id}/transactions` in REST server
- [x] adds transaction metadata to `tx-header` and `tx` DTOs in REST API.


bwbush (approved):
Thanks! So now we have metadata both in building transactions and in history.

PR #402, opened 2022-12-15
SCP-4762 Erroneous `TEApplyNoMatchError` when applying inputs in Marlowe Runtime

jhbertra:
- [x] rewrites `loadMarloweContext` to use `ChainSeek` directly instead of `MarloweSync`
- [x] removes `marlowe-tx`'s dependency on `marlowe-history`

Explanation of cause: because the `MarloweContext` was being loaded from `marlowe-history`, there was a slight chance that it would still have stale information in it (i.e. the datum from before the previous transaction was confirmed), so it would try to match the inputs against the old datum and fail.

The fix was to load the MarloweContext directly from `chainseekd` instead. this guarantees that the data in the `MarloweContext` is at least as fresh as the data seen when the previous tx was confirmed. Note that this does _not_ guarantee it will work - if a rollback occurs between confirming the previous tx and building the next one, the same problem could happen. However, that isn't a bug - it's a legitimate case that might happen.

### DIFF
--- a/marlowe-chain-sync/chainseekd/Main.hs
+++ b/marlowe-chain-sync/chainseekd/Main.hs
@@ -32,7 +32,7 @@ import Language.Marlowe.Runtime.ChainSync.Database (hoistDatabaseQueries)
 import qualified Language.Marlowe.Runtime.ChainSync.Database.PostgreSQL as PostgreSQL
 import Logging (RootSelector(..), getRootSelectorConfig)
 import Network.Protocol.ChainSeek.Server (chainSeekServerPeer)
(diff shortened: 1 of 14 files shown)


bwbush (approved):
Cool! There were no API changes that percolated into the webserver or scaling clients.

I tested this on-chain with the scaling tool and there were not failures in the short run I did. I'll do a long run overnight.

bwbush:
We did receive two `TEApplyNoMatchErrors` out of ~3000 contracts run. See https://input-output.atlassian.net/browse/SCP-4825 for the log file.

jhbertra:
I'll take a look at the logs - still going to merge this because it is clearly an improvement! I wonder if the errors you saw occurred because of a rollback.

PR #406, opened 2022-12-19
SCP-4826 Setup Cardano Integration Testing Environment

jhbertra:
Basic idea:

```hs
-- testnet contains file paths to node setup, e.g. socket files, config files, payment key pairs.
withLocalTestnet \testnet -> do
  -- test code here
```

To share the same local testnet between multiple hspec tests (because setting up the testnet takes a few seconds):

```hs
aroundAll withLocalTestnet do
  describe "context 1" do
    it "integration 1" \testnet -> -- test code here
    it "integration 2" \testnet -> -- test code here
  describe "context 2" do
    it "integration 3" \testnet -> -- test code here
    it "integration 4" \testnet -> -- test code here
```

Running `withLocalTestnet` creates a temporary workspace directory with a UUID in the path. When it completes successfully, the workspace dir is cleaned up. If the action fails with an exception, the exception will be rethrown with context information added. Additionally, the workspace directory will not be cleaned up to aid in diagnosis (e.g. by reading the SPO node log files).

`withLocalTestnet` will also spawn N SPO `cardano-node` processes, where N is a configurable number (via `withLocalTestnet'`). These processes are managed via `ResourceT` and are guaranteed to be cleaned up before the test action exits (normally or via an exception).


bwbush:
Is this behavior expected? If so, how should I start the testnet?
```console
$ cabal run exe:testbed
Up to date
testbed: ./configuration/alonzo-babbage-test-genesis.json: copyFile:atomicCopyFileContents:withReplacementFile:copyFileToHandle:openBinaryFile: does not exist (No such file or directory)
```

cardano-integration/src/Test/Integration/Cardano/Local.hs
```diff
@@ -0,0 +1,444 @@
...
+  , shelleyGenesisJson :: FilePath
+  , alonzoGenesisJson :: FilePath
+  } deriving (Show, Eq, Ord, Generic)
+
+instance ToJSON Network
+
+defaultOptions :: LocalTestnetOptions
+defaultOptions = LocalTestnetOptions
+  { slotDuration = 1000
+  , securityParam = 10
```
bwbush, line 129: This might be too low for rollback testing.

cardano-integration/src/Test/Integration/Cardano/Local.hs
```diff
@@ -0,0 +1,444 @@
...
+data LocalTestnetOptions = LocalTestnetOptions
+  { slotDuration :: Int
+  , securityParam :: Int
+  , numSpoNodes :: Int
+  , numDelegators :: Int
+  , numWallets :: Int
+  , totalBalance :: Int
+  } deriving (Show, Eq, Ord)
+
+data LocalTestnet = LocalTestnet
```
bwbush, line 64: This type is also consistent with a public testnet if we allow `delegators` and `spoNodes` to be empty. I'm not sure if we should rename this to be more general vs eventually creating a type for public testnets.

cardano-integration/src/Test/Integration/Cardano/Local.hs
```diff
@@ -0,0 +1,444 @@
...
+      , "minThd"            .= ("600000000000000" :: String)
+      , "thdDecrement"      .= ("50000000000000" :: String)
+      ]
+    , "txFeePolicy"       .= object
+      [ "multiplier"        .= ("43946000000" :: String)
+      , "summand"           .= ("155381000000000" :: String)
+      ]
+    ]
+  let byronGenesisDir = resolveWorkspacePath workspace "byron-genesis"
+  execCli_
```
bwbush, line 269: We could just call the function in the `cardano-cli` package, instead of exec'ing it.

configuration/byron-mainnet/configuration.yaml
```diff
@@ -0,0 +1,287 @@
...
+TraceBlockFetchProtocolSerialised: False
+
+# Trace BlockFetch server.
+TraceBlockFetchServer: False
+
+# Trace BlockchainTime.
+TraceBlockchainTime: False
+
+# Verbose tracer of ChainDB
+TraceChainDb: True
```
bwbush, line 168: We might want to turn off most of the irrelevant logging, except for `TraceChainDb` and `TraceMempool`, since those are the only ones that are actually useful for debugging transactions and rollbacks.

jhbertra:
> Is this behavior expected? If so, how should I start the testnet?
> 
> ```
> $ cabal run exe:testbed
> Up to date
> testbed: ./configuration/alonzo-babbage-test-genesis.json: copyFile:atomicCopyFileContents:withReplacementFile:copyFileToHandle:openBinaryFile: does not exist (No such file or directory)
> ```

What PWD are you running the command from? I could make it so that it resolves paths from the repo root no matter where you run it from. The idea is to run it like `cabal run testbed` from the repo root.

cardano-integration/src/Test/Integration/Cardano/Local.hs
jhbertra: I is just a default - you can provide a different value.

cardano-integration/src/Test/Integration/Cardano/Local.hs
jhbertra: I don't really know right now - we can always refactor when we add the public testnet support. Things will be clearer then.

cardano-integration/src/Test/Integration/Cardano/Local.hs
jhbertra: True, we could. I'll look into this. I guess this way, we don't need to depend on `cardano-cli` at the cabal level.

configuration/byron-mainnet/configuration.yaml
jhbertra: Yes, I was thinking that as well.

bwbush:
It looks like it requires running from the top folder of the repo: 
```console
$ git clone git@github.com:input-output-hk/marlowe-cardano.git -b scp-4826-cardano-test-environment
. . .
$ cd marlowe-cardano/
$ nix develop
. . .
$ cd cardano-integration/
$ cabal run exe:testbed
. . .
testbed: ./configuration/alonzo-babbage-test-genesis.json: copyFile:atomicCopyFileContents:withReplacementFile:copyFileToHandle:openBinaryFile: does not exist (No such file or directory)
$ cd ..
$ cabal run exe:testbed
Testnet started, press enter to exit
```

bwbush (approved):
It works if I launch from the top folder of the repo.

I think it would be very useful if `Main.main` printed out the location of the workspace that it created. (I see than it is in `$TMP`, but I already have six folders in `$TMP/workspaces/` so I had to check timestamps to see which one was just created.

jhbertra:
> It works if I launch from the top folder of the repo.
> 
> I think it would be very useful if `Main.main` printed out the location of the workspace that it created. (I see that it is in `$TMP`, but I already have six folders in `$TMP/workspaces/` so I had to check timestamps to see which one was just created.

Yeah, I will add this. I didn't intend the `testbed` executable to survive this PR, I just needed a way to test locally. I can rework it into something a bit more robust.

PR #415, opened 2023-01-05
SCP-4748 Addition of test client for `marlowe-test-spec` from `marlowe` repository

bwbush:
This PR runs tests of the Cardano implementation of Marlowe against the `marlowe-test-spec` framework of https://github.com/input-output-hk/marlowe.

Note that this PR depends upon the acceptance of https://github.com/input-output-hk/marlowe/pull/155 in the `marlowe` repository.

Pre-submit checklist:
- Branch
    - [x] [Test report is updated](https://github.com/input-output-hk/marlowe-cardano/blob/main/marlowe/test/test-report.md) (if relevant)
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


marlowe/test/test-report.md
```diff
@@ -400,7 +400,7 @@ Several other property-based tests, in the [`Spec.Marlowe.Marlowe`](src/Spec/Mar
 
 ## Test Oracle for Marlowe
```
hrajchert, line 401: I would rename this for Test Spec, as oracle can be misinterpreted

marlowe-test/src/Data/Jsonable.hs
```diff
@@ -0,0 +1,126 @@
...
+import Data.Proxy (Proxy(..))
+import Test.Tasty.QuickCheck (Arbitrary(..), Gen)
+
+
+-- | A list of known types with `FromJSON`, `ToJSON`, and `Arbitrary` instances.
+type KnownJsonable = [JsonableType]
+
+
+-- | A type with `FromJSON`, `ToJSON`, and `Arbitrary` instances.
+data JsonableType = forall a. (ToJSON a, FromJSON a, Arbitrary a) => JsonableType {typeKey :: String, typeValue :: Proxy a}
```
hrajchert, line 47: It might be too much to require Arbitrary for all the types, as we only require them for Token and Party. I see that you had to add arbitrary instances for Payment for example. No need to change it, just an observation

hrajchert (approved):
it LGTM. On the running inside marlowe-test or via a separate command I don't have a strong opinion. I would leave it as it is unless we find a strong reason to change it.

On the benefits, the tests runs much faster as it doesn't involve spawning multiple process. On the drawbacks, it might be too dependend on changes to the spec test driver. For example, at some point I plan to modify the way the `ClientProcess` is passed. Right now I manually pass the reference and I see that you pass it when adding the test suite, eventually I'll make a custom monad to rely on a type class routing. By doing that it will break this tests when it shouldn't, as it is an internal change.

marlowe-test/spec-client/Main.hs
```diff
@@ -0,0 +1,59 @@
...
+main =
+  do
+    hSetBuffering stdin LineBuffering
+    hSetBuffering stdout LineBuffering
+    stream <- lines <$> getContents
+    sequence_
+      [
+        case A.eitherDecode $ LBS8.pack chunk of
+          Right request -> LBS8.putStrLn . A.encode =<< handle request
+          Left message  -> die message
```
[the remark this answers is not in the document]
palas: It seems the problem is that the last triple quote is not in its own line, which seems a requirement for parseChunks. Is that a requirement or the format?

marlowe-test/src/Spec/Marlowe/Service/Serialization.hs
```diff
@@ -0,0 +1,106 @@
...
+-- | Deserialize and then serialize a value.
+roundtripSerialization
+  :: String  -- ^ The key to the type.
+  -> A.Value  -- ^ The value.
+  -> SerializationResponse  -- ^ The result.
+roundtripSerialization typeSerialized valueSerialized =
+  if isKnownJson knownJsonTypes typeSerialized
+    then case roundTripJsonable knownJsonTypes typeSerialized valueSerialized of
+             Right valueReserialized -> SerializationSuccess{..}
+             Left serializationError      -> SerializationError{..}
```
palas, line 80:
```suggestion
             Left serializationError  -> SerializationError{..}
```
Just strange indentation

palas (approved):
Thanks, looks good 👍

marlowe-test/spec-client/Main.hs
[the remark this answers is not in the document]
hrajchert: The format doesnt set a  requirement of where the triple backslash should be

marlowe-test/src/Spec/Marlowe/Service/Serialization.hs
bwbush: Fixed in c4cbff6a21bc2da875aa867375658d8f8fe803a6.

marlowe/test/test-report.md
bwbush: Fixed in c4cbff6a21bc2da875aa867375658d8f8fe803a6.

marlowe-test/spec-client/Main.hs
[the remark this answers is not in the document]
bwbush: Fixed in c4cbff6a21bc2da875aa867375658d8f8fe803a6.

PR #419, opened 2023-01-11
Missing metadata in marlowe-runtime for discovered contracts

jhbertra:


yveshauser (approved):
Works as expected when testing manually

PR #436, opened 2023-01-25
SCP-4881 Fixed loss of redeemer bytes in chain index

bwbush:
The `Redemer` from the block is represented as `MemoizedBytes`, which includes an `originalBytes` function for retrieving the actual on-chain bytes. Deserializing and the serializing does not yield the same CBOR bytes, but this round-trip serialization was what was stored in the database record. This PR retains the original bytes.

Tested via:
- 100% match now for redeemer bytes between `marlowe-chain-indexer` versus `cardano-db-sync` for both `preprod` and `preview` networks. (Formerly there had been a one-byte shortage on the Marlowe side.)
- Ran two contracts on `preprod` to make sure `marlowe-tx` still worked.
- Successfully deserialized all Marlowe contracts on `preprod` using `marlowe-finder`.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


jhbertra (approved):
Did you check that the result deserializes properly?

bwbush:
> Did you check that the result deserializes properly?

Yes and no, @jhbertra, I've looked into this very interesting question. Here is what I've found:
1. The ledger specification does require that `plutus_data` be the CBOR serialization of `Data`.
2. There are datums on the ledger that have extra bytes beyond the CBOR serialization, and this doesn't seem to be simple padding. Some CBOR deserializers will report that there are extra trailing bytes. However, those extra bytes were hashed when originally computing the datum, so they cannot be safely discarded.
3. All of the Marlowe datums have proper CBOR with no extra trailing bytes.
4. My conjecture is that some folks are using non-Haskell CBOR libraries to serialize their datums, and these trailing bytes are an artifact. We'd have to run these third-party Plutus scripts against the datum and redeemer on chain to verify that the node is ignoring the trailing bytes etc. when calls the untyped validator as `BuiltinData -> BuiltinData -> BuiltinData -> ()`

For the chain index, I think it is safest to take the bytes and hash as they appear on the ledger and not to check their conformance and CBOR deserialization: otherwise, we'd be second-guessing the ledger. This problem is irrelevant for all Marlowe contracts, but someday we might be running Marlowe and Plutus in the same tx, so I think we should respect the on-chain data, since that is what the Plutus scripts are expecting.

PR #438, opened 2023-01-27
SCP-4874 MarloweSync Server

jhbertra:
As a user of `marlowe-sync`, I want it to support the `MarloweSync` protocol so that existing clients of `marlowe-history` can easily switch to using `marlowe-sync`.

### DIFF
--- a/marlowe-runtime/sync/Language/Marlowe/Runtime/Sync/Database/PostgreSQL/GetIntersectionForContract.hs
+++ b/marlowe-runtime/sync/Language/Marlowe/Runtime/Sync/Database/PostgreSQL/GetIntersectionForContract.hs
@@ -1,9 +1,76 @@
 module Language.Marlowe.Runtime.Sync.Database.PostgreSQL.GetIntersectionForContract
   where
 
+import Data.ByteString (ByteString)
+import Data.Int (Int64)
+import Data.Maybe (listToMaybe)
+import qualified Data.Vector as V
+import Hasql.TH (vectorStatement)
 import qualified Hasql.Transaction as T
-import Language.Marlowe.Runtime.ChainSync.Api (BlockHeader, ChainPoint)
-import Language.Marlowe.Runtime.Core.Api (ContractId, MarloweVersion)
+import Language.Marlowe.Runtime.ChainSync.Api
+  (BlockHeader(..), BlockHeaderHash(..), ChainPoint, TxId(..), TxOutRef(..), WithGenesis(..))
+import Language.Marlowe.Runtime.Core.Api (ContractId(ContractId))
 
-getIntersectionForContract :: ContractId -> MarloweVersion v -> [BlockHeader] -> T.Transaction ChainPoint
-getIntersectionForContract = undefined
+getIntersectionForContract :: ContractId -> [BlockHeader] -> T.Transaction ChainPoint
+getIntersectionForContract _ [] = pure Genesis
+getIntersectionForContract (ContractId TxOutRef{..}) (b : bs) = do
+  serverBlocks <- fmap decodeBlock . V.toList <$> T.statement params
+    [vectorStatement|
+      WITH contractId (txId, txIx) AS
+        ( SELECT $1 :: bytea, $2 :: smallint
+        )
+      SELECT
+        block.slotNo :: bigint,
+        block.id :: bytea,
+        block.blockNo :: bigint
+      FROM marlowe.createTxOut
+      JOIN marlowe.block
+        ON block.id = createTxOut.blockId
+      JOIN contractId USING (txId, txIx)
+      WHERE block.rollbackToSlot IS NULL
+      UNION
+      SELECT DISTINCT
+        block.slotNo :: bigint,
+        block.id :: bytea,
+        block.blockNo :: bigint
+      FROM marlowe.applyTx
+      JOIN marlowe.block
+        ON block.id = applyTx.blockId
+      JOIN contractId
+        ON contractId.txId = applyTx.createTxId
+        AND contractId.txIx = applyTx.createTxIx
+      WHERE block.rollbackToSlot IS NULL
+      UNION
+      SELECT DISTINCT
+        block.slotNo :: bigint,
+        block.id :: bytea,
+        block.blockNo :: bigint
+      FROM marlowe.withdrawalTxIn
+      JOIN marlowe.block
+        ON block.id = withdrawalTxIn.blockId
+      JOIN marlowe.payoutTxOut
+        ON payoutTxOut.txId = withdrawalTxIn.payoutTxId
+        AND payoutTxOut.txIx = withdrawalTxIn.payoutTxIx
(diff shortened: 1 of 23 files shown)


marlowe-runtime/sync/Language/Marlowe/Runtime/Sync/Database/PostgreSQL/GetIntersectionForContract.hs
```diff
@@ -0,0 +1,77 @@
...
+        block.slotNo :: bigint,
+        block.id :: bytea,
+        block.blockNo :: bigint
+      FROM marlowe.createTxOut
+      JOIN marlowe.block
+        ON block.id = createTxOut.blockId
+      JOIN contractId USING (txId, txIx)
+      WHERE block.rollbackToSlot IS NULL
+      UNION
+      SELECT DISTINCT
```
bwbush, line 33: Actually, isn't `DISTINCT` redundant because `UNION` eliminates duplicates?

bwbush: Also, it might be worthwhile adding a comment that the purpose of the unions is to find creation, applications, and withdrawals involving the contract.

marlowe-runtime/sync/Language/Marlowe/Runtime/Sync/Database/PostgreSQL/GetNextSteps.hs
```diff
@@ -0,0 +1,430 @@
...
+  [maybeStatement|
+    WITH params (createTxId, createTxIx, afterSlot) AS
+      ( SELECT $1 :: bytea, $2 :: smallint, $3 :: bigint
+      )
+    , nextApplyTxIds (slotNo, blockHeaderHash, blockNo, txIds) AS
+      ( SELECT
+          block.slotNo,
+          (ARRAY_AGG(block.id))[1],
+          (ARRAY_AGG(block.blockNo))[1],
+          ARRAY_AGG(applyTx.txId)
```
bwbush, line 107: I'm not sure if it's better to use `ARRAY_AGG` or just to add to the `GROUP BY`, given the `LIMIT 1`. I guess that either way is fine.

marlowe-runtime/sync/Language/Marlowe/Runtime/Sync/Database/PostgreSQL/GetNextSteps.hs
```diff
@@ -0,0 +1,430 @@
...
+    SELECT
+      COALESCE (nextApplyTxIds.slotNo, nextWithdrawalTxIds.slotNo) :: bigint,
+      COALESCE (nextApplyTxIds.blockHeaderHash, nextWithdrawalTxIds.blockHeaderHash) :: bytea,
+      COALESCE (nextApplyTxIds.blockNo, nextWithdrawalTxIds.blockNo) :: bigint,
+      COALESCE (nextApplyTxIds.txIds, '{}') :: bytea[],
+      COALESCE (nextWithdrawalTxIds.txIds, '{}') :: bytea[]
+    FROM nextApplyTxIds
+    FULL OUTER JOIN nextWithdrawalTxIds
+    ON nextApplyTxIds.slotNo = nextWithdrawalTxIds.slotNo
+    ORDER BY COALESCE (nextApplyTxIds.slotNo, nextWithdrawalTxIds.slotNo)
```
bwbush, line 148: Would this read more cleanly as a `UNION` instead of a `FULL OUTER JOIN` with all of the `COASLESCE`s? I guess the point is that applications are preferred over withdrawals if both occur in the same slot. In principle, one would like to look at the position of the transaction in the block and order by that, but we don't have such data.

Also, it would be interesting to measure the performance of this on the example contract that involved ~400 transactions.

marlowe-runtime/sync/Language/Marlowe/Runtime/Sync/Database/PostgreSQL/GetTipForContract.hs
```diff
@@ -0,0 +1,73 @@
...
+module Language.Marlowe.Runtime.Sync.Database.PostgreSQL.GetTipForContract
+  where
+
+import Hasql.TH (maybeStatement)
+import qualified Hasql.Transaction as T
+import Language.Marlowe.Runtime.ChainSync.Api
+  (BlockHeader(..), BlockHeaderHash(..), ChainPoint, TxId(..), TxOutRef(..), WithGenesis(..))
+import Language.Marlowe.Runtime.Core.Api (ContractId(..))
+
+getTipForContract :: ContractId -> T.Transaction ChainPoint
```
bwbush, line 12: Thanks! ❤️

bwbush:
@ladamesny, our test cases for `marlowe-sync` should involve lengthy contracts that interleave various combinations of withdrawals and input applications, including ones that occur in the same block.

marlowe-runtime/sync/Language/Marlowe/Runtime/Sync/Database/PostgreSQL/GetIntersectionForContract.hs
jhbertra: I can add that comment

jhbertra: Does `UNION` eliminate duplicates within the sub queries, or only between the two queries?

marlowe-runtime/sync/Language/Marlowe/Runtime/Sync/Database/PostgreSQL/GetNextSteps.hs
jhbertra: I tried both - `(ARRAY_AGG(expr))[1]` gives better performance than multiple `GROUP BY` keys.

marlowe-runtime/sync/Language/Marlowe/Runtime/Sync/Database/PostgreSQL/GetNextSteps.hs
jhbertra: It would not be correct if it was a `UNION` - the purpose of this is to join the two sub queries by block ID, and take the minimum block which has apply transactions, withdraw transactions, or both.

jhbertra:
> Questions:
> 
>     1. Should we soon formally market the history and discovery client modules with the `DEPRECATED` pragma?
> 
>     2. Will `MarloweHeaderSyncClient` be retired, or is it scheduled to be implemented in sync?
> 
>     3. Would it be useful for me to put the 400-transaction example Marlowe contract on `preprod` and `preview`, so we can verify performance?

1. `marlowe-history` can already be deprecated when this is merged, `marlowe-discovery` can be deprecated when its APIs are replicated here.
2. `MarloweHeaderSyncClient` will be implemented in sync
3. Yes, that would be quite useful!

PR #440, opened 2023-01-27
Scp 4677 QC test for balanceTx

dino-iog:
I figured out what was wrong with my test TxBodyContent (it wasn't empty enough to begin with). We can discuss if this test seems like it's doing enough.

### DIFF
--- a/marlowe-runtime/marlowe-runtime.cabal
+++ b/marlowe-runtime/marlowe-runtime.cabal
@@ -53,10 +53,6 @@ common lang
     StandaloneDeriving
     TypeApplications
     TupleSections
-  ghc-options:
-    -Wall -Wnoncanonical-monad-instances -Wunused-packages
-    -Wincomplete-uni-patterns -Wincomplete-record-updates
-    -Wredundant-constraints -Widentities
   if flag(defer-plugin-errors)
     ghc-options: -fplugin-opt PlutusTx.Plugin:defer-errors
 
@@ -704,6 +700,7 @@ test-suite marlowe-runtime-test
     , bytestring
     , cardano-api
     , cardano-api:gen
+    , cardano-slotting
     , containers
     , errors
     , hedgehog-quickcheck
@@ -731,6 +728,7 @@ test-suite marlowe-runtime-test
     , stm
     , time
     , transformers
+    , tx-api
     , typed-protocols
   build-tool-depends: hspec-discover:hspec-discover
   ghc-options: -threaded
--- a/marlowe-runtime/marlowe-runtime.cabal
+++ b/marlowe-runtime/marlowe-runtime.cabal
@@ -53,6 +53,10 @@ common lang
     StandaloneDeriving
     TypeApplications
     TupleSections
+  ghc-options:
+    -Wall -Wnoncanonical-monad-instances -Wunused-packages
+    -Wincomplete-uni-patterns -Wincomplete-record-updates
+    -Wredundant-constraints -Widentities
   if flag(defer-plugin-errors)
     ghc-options: -fplugin-opt PlutusTx.Plugin:defer-errors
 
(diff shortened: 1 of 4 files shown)


marlowe-runtime/marlowe-runtime.cabal
```diff
@@ -53,10 +53,6 @@ common lang
     StandaloneDeriving
     TypeApplications
     TupleSections
-  ghc-options:
```
dino-iog, line 56: Oops! I accidentally left this from development. Fixing now.

bwbush, line 56: Why are we removing all of the warnings?

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/ConstraintsSpec.hs
```diff
@@ -486,6 +492,109 @@ spec = do
              $ ensureMinUtxo protocolTestnet (inAddress, inValue)
          pure $ selectLovelace outValue `shouldBe` maximum [selectLovelace inValue, selectLovelace expected]
 
+  describe "balanceTx" do
+    prop "tx should balance if coin selection succeeded and no Plutus failure occurred" \(SomeTxConstraints marloweVersion constraints) -> do
+      marloweContext <- genSimpleMarloweContext marloweVersion constraints
+
+      -- We MUST dictate the distribution of wallet context assets, default
+      -- generation only tests with empty wallets!
+      maxLovelace <- choose (0, 40_000_000)
```
bwbush, line 501: Do we really want to generate a wallet with a tiny amount of Lovelace (< ~1,000,000)? No real wallet could contain that little. I suppose this is okay because the test should be conditional on the transaction being balanceable.

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/ConstraintsSpec.hs
```diff
@@ -486,6 +492,109 @@ spec = do
...
+
+        -- We need to make a TxBodyContent that would have come from executing
+        -- selectCoins, containing the tx information in the WalletContext we
+        -- will also be passing to balanceTx. To do so, we'll use the
+        -- walletContext.
+
+        addBuilder :: TxIn -> (TxIn, BuildTxWith BuildTx (Witness WitCtxTxIn BabbageEra))
+        addBuilder = (, BuildTxWith (KeyWitness KeyWitnessForSpending))
+
+        txBodyContent = emptyTxBodyContent
```
bwbush, line 559: This `txBodyContent` just spends the whole wallet. It ignores the `marloweContext` and it doesn't create a transaction that matches the description of this property, "tx should balance if coin selection succeeded and no Plutus failure occurred", since coin selection was never called and there is no possibility of the Marlowe validator being run. Perhaps the test should be renamed "tx should balance for non-Plutus transactions where the wallet has sufficient funds"?

bwbush (commented):
This doesn't test what the property "tx should balance if coin selection succeeded and no Plutus failure occurred" states it is testing.

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/ConstraintsSpec.hs
dino-iog: It's generating a small amount of wallets that result in unbalanceable transactions

```  balanceTx
    tx should balance if coin selection succeeded and no Plutus failure occurred [v]     
      +++ OK, passed 100 tests:
      95% balancing succeeded
       5% non-balanceable test cases
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
But we don't have to leave it that way if we don't like it.

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/ConstraintsSpec.hs
dino-iog: I was using `selectCoins` initially but reworked it to not because @jhbertra was concerned about potentially buggy code being involved in this test, which is valid.
I don't understand how to make a wallet + marlowe context + tx body that looks close to what comes out of successful coin selection.

jhbertra: This is partly my fault - I discouraged @dino-iog from calling `selectCoins` in the test, but you are right that the test title indicates that it should be.

bwbush: Perhaps we should just rename the test?

PR #442, opened 2023-01-30
SCP-4875 MarloweHeaderSync Server

jhbertra:
As a user of `marlowe-sync`, I want it to support the `MarloweHeaderSync` protocol so that existing clients of `marlowe-discovery` can easily switch to using `marlowe-sync`.


bwbush (approved):
Very nice. I look forward to using this drop-in replacement for `marlowe-discovery`!

PR #444, opened 2023-01-31
Scp 4992 get contract headers

jhbertra:
- [x] Adds new table columns and indexes for improved query performance
- [x] Adds new `MarloweQuery` protocol for queries supported by `marlowe-sync`
- [x] Implements `GetContractHeaders` query.

### DIFF
--- a/marlowe-runtime/indexer/Language/Marlowe/Runtime/Indexer/Database/PostgreSQL/GetMarloweUTxO.hs
+++ b/marlowe-runtime/indexer/Language/Marlowe/Runtime/Indexer/Database/PostgreSQL/GetMarloweUTxO.hs
@@ -24,27 +24,21 @@ getMarloweUTxO block = runMaybeT do
   slotNo <- MaybeT $ H.statement (prepareBlockQueryParams block) [maybeStatement|
     SELECT slotNo :: bigint
       FROM marlowe.block
-     WHERE id = $1 :: bytea AND slotNo = $2 :: bigint AND blockNo = $3 :: bigint AND rollbackToSlot IS NULL AND rollbackToBlock IS NULL
+     WHERE id = $1 :: bytea AND slotNo = $2 :: bigint AND blockNo = $3 :: bigint
   |]
 
-
   unspentContractOutputs <- lift $ Map.fromDistinctAscList . V.toList . fmap decodeContractOutputRow <$> H.statement slotNo [vectorStatement|
-    WITH block (blockId) AS
-      ( SELECT id :: bytea
-          FROM marlowe.block
-         WHERE slotNo <= $1 :: bigint AND rollbackToSlot IS NULL AND rollbackToBlock IS NULL
-      )
-    , contractOut (createTxId, createTxIx, txId, txIx, address, payoutScriptHash) AS
+    WITH contractOut (createTxId, createTxIx, txId, txIx, address, payoutScriptHash) AS
       ( SELECT createTxOut.txId
              , createTxOut.txIx
              , txOut.txId
              , txOut.txIx
              , txOut.address
              , contractTxOut.payoutScriptHash
-          FROM block
-          JOIN marlowe.txOut AS txOut USING (blockId)
-          JOIN marlowe.contractTxOut AS contractTxOut USING (txId, txIx)
-          JOIN marlowe.createTxOut AS createTxOut USING (txId, txIx)
+          FROM marlowe.txOut
+          JOIN marlowe.contractTxOut USING (txId, txIx)
+          JOIN marlowe.createTxOut USING (txId, txIx)
+          WHERE createTxOut.slotNo <= $1 :: bigint
          UNION
         SELECT applyTx.createTxId
              , applyTx.createTxIx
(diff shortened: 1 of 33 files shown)


marlowe-runtime/sync/Language/Marlowe/Runtime/Sync/Database/PostgreSQL/GetTipForContract.hs
```diff
@@ -15,53 +15,32 @@ getTipForContract (ContractId TxOutRef{..}) = T.statement params $ decodePoint <
...
-        block.id :: bytea,
-        block.blockNo :: bigint
-      FROM
-        ( SELECT * FROM createBlock
-          UNION SELECT * FROM applyBlock
-          UNION SELECT * FROM withdrawalBlock
-        ) AS block
-      ORDER BY block.slotNo DESC
-      LIMIT 1
+    SELECT
```
bwbush, line 18: Nice!

bwbush (approved):
Thanks for adding the slot and block numbers to the header. I think that will make client-side rollback handling much easier.

marlowe-runtime/indexer/Language/Marlowe/Runtime/Indexer/Database/PostgreSQL/GetMarloweUTxO.hs
```diff
@@ -24,49 +24,44 @@ getMarloweUTxO block = runMaybeT do
   slotNo <- MaybeT $ H.statement (prepareBlockQueryParams block) [maybeStatement|
     SELECT slotNo :: bigint
       FROM marlowe.block
-     WHERE id = $1 :: bytea AND slotNo = $2 :: bigint AND blockNo = $3 :: bigint AND rollbackToSlot IS NULL AND rollbackToBlock IS NULL
+     WHERE id = $1 :: bytea AND slotNo = $2 :: bigint AND blockNo = $3 :: bigint
   |]
 
-
   unspentContractOutputs <- lift $ Map.fromDistinctAscList . V.toList . fmap decodeContractOutputRow <$> H.statement slotNo [vectorStatement|
-    WITH block (blockId) AS
```
bwbush, line 32: I don't understand why this pattern was problematic, aside from some added complexity.

bwbush: This pattern worked for `marlowe-chain-indexer`. What is the fundamental difference for `marlowe-indexer`?

bwbush (approved):
The design looks safe to me, but I judged that the previous approach in this PR was safe, too, so I'm not very confident in my judgment. Both methods seemed equally safe, with the previous one maybe be more performant but complex, whereas this version might be less performant but cleaner.

I am concerned about the performance of the cascading delete because this is a blocking operation that I've seen be very troublesome for `cardano-db-sync`. I feel that we should do some stressful rollback testing on our rollback testnet. I'd like to put a lot of conflicting Marlowe transactions onto a forked testnet, have the network rejoin to reach consensus, and then verify the integrity of the tables in the `marlowe` schema.

jhbertra:
> The design looks safe to me, but I judged that the previous approach in this PR was safe, too, so I'm not very confident in my judgment. Both methods seemed equally safe, with the previous one maybe be more performant but complex, whereas this version might be less performant but cleaner.

This approach is definitely safer than the last one. With the previous approach, there was a high likelihood (as evidenced by the numerous bugs I fixed related to this problem) that we would end up serving the client rolled-back blocks because of neglecting to check for `rollbackToBlock IS NOT NULL`. In this schema, only non-rolled back blocks exist in the `block` table, so this mistake cannot happen.

> I am concerned about the performance of the cascading delete because this is a blocking operation that I've seen be very troublesome for `cardano-db-sync`. I feel that we should do some stressful rollback testing on our rollback testnet. I'd like to put a lot of conflicting Marlowe transactions onto a forked testnet, have the network rejoin to reach consensus, and then verify the integrity of the tables in the `marlowe` schema.

There isn't any difference from a performance perspective between this and what we were doing before (which is updating the `block` table and manually deleting any rows from the other tables). We were already doing a manual cascading delete, the only difference here is that now the DBMS manages this for us.

> Also, rollback forking is a tree-like behavior, but we are overwriting the tree-like behavior with a main branch and a "squashed" record of rollbacks. Could that cause trouble in finding intersections?

Yes, potentially, if the squashing logic is faulty. There is a performance and complexity tradeoff though. Essentially, we only care about what the "current" tip is, so from our POV the chain is linear. A safer approach for intersecting a sync client with the current chain would be to track the full history of blocks they have seen and intersect it with the current chain each time we service a request from them. This has performance and memory usage implications however.

marlowe-runtime/indexer/Language/Marlowe/Runtime/Indexer/Database/PostgreSQL/GetMarloweUTxO.hs
bwbush: Thanks for the explanation!

bwbush:
> neglecting to check for rollbackToBlock IS NOT NULL
Okay, so this isn't an intrinsic lack of safety, but just a danger in writing unsafe queries that omit the required `where` clause.

> squashing logic
The rollback table is hardly used, and not for anything critical, so I think the squashing is okay. I do think we should test some overlapping rollbacks, just in case.

marlowe-runtime/indexer/Language/Marlowe/Runtime/Indexer/Database/PostgreSQL/GetMarloweUTxO.hs
jhbertra: Thanks for clarifying that "this pattern = having rollback columns in the block table, instead of having a separate rollback table".

I already summarized the problems in my comment below - and as for `marlowe-sync` vs `marlowe-chain-sync`, this problem has also caused bugs in `marlowe-chain-sync`. However, here are the differences:

1. `marlowe-chain-sync` lacks foreign key constraints, so we get fewer guarantees and still need to carefully manage the cascading deletes of other tables when rolling back blocks
2. `marlowe-chain-sync` is currently more mature and mission-critical than `marlowe-sync`, so I am more reluctant to make such a drastic change there

I do however think that a similar change would be an improvement in `marlowe-chain-sync`, I'd just want to approach it a lot more carefully and in a separate PR.

PR #451, opened 2023-02-06
SCP-4817 Fix and unit test for timed-out validity interval computation.

bwbush:
Fix bug in computation of validity interval for a timed-out transaction, and added corresponding unit test.

Review tasks:
- [ ] Check revision to time logic for applying inputs.
- [ ] Check logic in unit test.
- [ ] Check inequalities in unit test.

Tested:
- 1,000,000 replications of property-based test.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
+++ b/marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
@@ -275,10 +275,8 @@ buildApplyInputsConstraintsSpec =
           :* K (oneSecondEraSummary 4) -- Alonzo lasted 1 second
           :* K (unboundedEraSummary 5) -- Babbage never ends
           :* Nil
-    Hspec.QuickCheck.prop "valid slot interval for timed-out contract" do
+    Hspec.QuickCheck.prop "valid slot interval for timed-out contract" \assets utxo -> do
       address <- arbitrary
-      assets <- arbitrary
-      utxo <- arbitrary
       -- The choice intervals for the tip, minimum time, and timeout overlap, so every ordering will occur.
       tipSlot' <- chooseInteger (9, 20) -- Make sure the tip is in the Babbage era.
       minTime <- oneof
--- a/marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
+++ b/marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
@@ -310,6 +310,8 @@ buildApplyInputsConstraintsSpec =
         contract' = Semantics.When [] (POSIXTime timeout) $ Semantics.When [] (POSIXTime timeout') Semantics.Close
       marloweContract <- elements [contract, contract'] -- This contract can only time out.
       let
+        -- Important note: these slot computations cannot be used generally, but are specificially tailored
+        -- to the contrived era history and system start used for this test case.
         toSlot = (`div` 1000)
         tipSlot = Chain.SlotNo $ fromInteger tipSlot'
         tipTime = 1000 * tipSlot'
--- a/marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
+++ b/marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
@@ -51,7 +51,19 @@ import Spec.Marlowe.Semantics.Arbitrary ()
 import Test.Hspec (Spec, shouldBe)
 import qualified Test.Hspec as Hspec
 import qualified Test.Hspec.QuickCheck as Hspec.QuickCheck
-import Test.QuickCheck (Arbitrary(..), Property, chooseInteger, discard, genericShrink, listOf1, oneof, suchThat, (===))
+import Test.QuickCheck
+  ( Arbitrary(..)
+  , Property
+  , chooseInteger
+  , counterexample
+  , discard
+  , elements
+  , genericShrink
+  , listOf1
+  , oneof
+  , suchThat
+  , (===)
+  )
 import qualified Test.QuickCheck as QuickCheck
 import Test.QuickCheck.Instances ()
 
@@ -240,6 +252,7 @@ withdrawSpec = Hspec.describe "buildWithdrawConstraints" do
 buildApplyInputsConstraintsSpec :: Spec
 buildApplyInputsConstraintsSpec =
   Hspec.describe "buildApplyInputsConstraints" do
+    -- TODO: Break up this preface to utility functions when additional properties are tested.
     let
       -- System start and era history are reusable across tests.
       systemStart = SystemStart $ posixSecondsToUTCTime 0  -- Without loss of generality.
@@ -275,27 +288,32 @@ buildApplyInputsConstraintsSpec =
           :* K (oneSecondEraSummary 4) -- Alonzo lasted 1 second
           :* K (unboundedEraSummary 5) -- Babbage never ends
           :* Nil
-    Hspec.QuickCheck.prop "valid slot interval for timed-out contract" \assets utxo -> do
-      address <- arbitrary
+      genTipSlot = chooseInteger (9, 20) -- Make sure the tip is in the Babbage era.
+      genMinTime = oneof
+                   [
+                     (1000 *) <$> chooseInteger (6, 20)  -- Even seconds, so there is a chance of collision.
+                   , chooseInteger (6000, 20000)         -- Milliseconds, so there is rounding off.
+                   ]
+      genTimeout = oneof
+                   [
+                     (1000 *) <$> chooseInteger (6, 20)  -- Even seconds, so there is a chance of collision.
+                   , chooseInteger (6000, 20000)         -- Milliseconds, so there is rounding off.
+                   ]
+    Hspec.QuickCheck.prop "valid slot interval for timed-out contract" \assets utxo address marloweParams-> do
       -- The choice intervals for the tip, minimum time, and timeout overlap, so every ordering will occur.
-      tipSlot' <- chooseInteger (9, 20) -- Make sure the tip is in the Babbage era.
-      minTime <- oneof
-                   [
-                     (1000 *) <$> chooseInteger (6, 20)  -- Even seconds, so there is a chance of collision.
-                   , chooseInteger (6000, 20000)         -- Milliseconds, so there is rounding off.
-                   ]
-      timeout <- oneof
-                   [
-                     (1000 *) <$> chooseInteger (6, 20)  -- Even seconds, so there is a chance of collision.
-                   , chooseInteger (6000, 20000)         -- Milliseconds, so there is rounding off.
-                   ]
-      marloweParams <- arbitrary
+      tipSlot' <- genTipSlot
+      minTime <- genMinTime
+      timeout <- genTimeout
+      timeout' <- genTimeout
+      let
+        contract = Semantics.When [] (POSIXTime timeout) Semantics.Close
+        contract' = Semantics.When [] (POSIXTime timeout) $ Semantics.When [] (POSIXTime timeout') Semantics.Close
+      marloweContract <- elements [contract, contract'] -- This contract can only time out.
       let
         toSlot = (`div` 1000)
         tipSlot = toEnum . fromEnum $ tipSlot'
         tipTime = 1000 * tipSlot'
         marloweState = Semantics.State AM.empty AM.empty AM.empty $ POSIXTime minTime
-        marloweContract = Semantics.When [] (POSIXTime timeout) Semantics.Close  -- This contract can only time out.
         datum = Semantics.MarloweData{..}
         marloweOutput = TransactionScriptOutput{..}
         result =
@@ -305,17 +323,24 @@ buildApplyInputsConstraintsSpec =
             tipSlot
             (Chain.TransactionMetadata mempty) Nothing Nothing mempty
       pure
-        case result of
-            -- A valid transaction will occur if the tip is not before the timeout.
-            Right _ -> toSlot timeout <= tipSlot'
-            -- A useless transaction will occur if the tip is before the timeout.
+        . counterexample ("tipTime = " <> show tipTime)
+        . counterexample ("minTime = " <> show minTime)
+        . counterexample ("contract = " <> show marloweContract)
+        . counterexample ("result = " <> show result)
+        $ case result of
+            Right _ ->
+              counterexample "A valid transaction will occur if tip is not before the first timeout."
+                $ toSlot timeout <= tipSlot'
             Left (ApplyInputsConstraintsBuildupFailed (MarloweComputeTransactionFailed "TEUselessTransaction")) ->
-              tipTime < timeout
-            -- If rounding off causes the timeout to fall at the tip, then there is an ambiguity.
-            Left (ApplyInputsConstraintsBuildupFailed (MarloweComputeTransactionFailed "TEAmbiguousTimeIntervalError")) ->
-              tipSlot' == toSlot timeout && minTime <= timeout
-            -- If the tip is in the past, then there should be an interval failure.
+              counterexample "A useless transaction will occur if the tip is before the timeout."
+                $ tipTime < timeout
             Left (ApplyInputsConstraintsBuildupFailed (MarloweComputeTransactionFailed message)) ->
-              "TEIntervalError (IntervalInPastError " `isPrefixOf` message == (tipTime < minTime)
-            -- Other failures should not occur.
-            Left _ -> False
+              if "TEIntervalError (IntervalInPastError " `isPrefixOf` message
+                then counterexample "If the tip is in the past, then the interval should be in the past."
+                  $ tipTime < minTime
+                else if "TEIntervalError (InvalidInterval " `isPrefixOf` message
+                  then counterexample "If rounding off causes the timeout to fall at the tip, then the interval should be invalid (effectively empty)."
+                    $ tipSlot' == toSlot timeout || marloweContract == contract' && tipSlot' == toSlot timeout'
+                else counterexample "Unexpected transaction failure" False
+            Left _ ->
+              counterexample "Unexpected transaction failure" False
--- a/marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
+++ b/marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
@@ -7,24 +7,30 @@ module Language.Marlowe.Runtime.Transaction.BuildConstraintsSpec
   ( spec
   ) where
 
+import Cardano.Api (ConsensusMode(..), EraHistory(EraHistory))
 import Data.ByteString (ByteString)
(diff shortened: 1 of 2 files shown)


marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
```diff
@@ -223,3 +236,88 @@ withdrawSpec = Hspec.describe "buildWithdrawConstraints" do
...
+            Right _ -> toSlot timeout <= tipSlot'
+            -- A useless transaction will occur if the tip is before the timeout.
+            Left (ApplyInputsConstraintsBuildupFailed (MarloweComputeTransactionFailed "TEUselessTransaction")) ->
+              tipTime < timeout
+            -- If rounding off causes the timeout to fall at the tip, then there is an ambiguity.
+            Left (ApplyInputsConstraintsBuildupFailed (MarloweComputeTransactionFailed "TEAmbiguousTimeIntervalError")) ->
+              tipSlot' == toSlot timeout && minTime <= timeout
+            -- If the tip is in the past, then there should be an interval failure.
+            Left (ApplyInputsConstraintsBuildupFailed (MarloweComputeTransactionFailed message)) ->
+              "TEIntervalError (IntervalInPastError " `isPrefixOf` message == (tipTime < minTime)
```
dino-iog, line 321: Could this end up being `False == False` as in `"TEIntervalError ..."` is NOT prefix of and `(tipTime >= minTime)` ? Just curious, seems like maybe an `&&` situation but I'm not sure.

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
```diff
@@ -223,3 +236,88 @@ withdrawSpec = Hspec.describe "buildWithdrawConstraints" do
...
+        datum = Semantics.MarloweData{..}
+        marloweOutput = TransactionScriptOutput{..}
+        result =
+          buildApplyInputsConstraints
+            systemStart eraHistory MarloweV1
+            marloweOutput
+            tipSlot
+            (Chain.TransactionMetadata mempty) Nothing Nothing mempty
+      pure
+        case result of
```
dino-iog, line 310: Not necessary but Pablo had shown me how to use `label` and `counterexample` in these checks so it produces little tables of the distribution of cases. Handy to see how well the generators are covering or if they're producing cases at all! You can see use of this in tests I wrote in `marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/ConstraintsSpec.hs`

dino-iog: They look like this

``` 
sufficient collateral is selected if possible [v]
  +++ OK, passed 100 tests:
  85% Wallet has funds, selection succeeded
  15% Wallet does not have funds, selection failed
```

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
bwbush: That's exactly what this is meant to do: the `(False == False) == True` will correctly succeed, meaning that if the transaction does not fail with a useless transaction or ambiguous time interval, then the test succeeds provided the transaction fails with an interval in the past error *if and only if* the tip is at least the minimum time. (This is quite a bit of nested logic.) The basic point is that a test that returns `(a :: Bool) == (b :: Bool)` encodes the logical operation "if and only if".

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
```diff
@@ -223,3 +236,88 @@ withdrawSpec = Hspec.describe "buildWithdrawConstraints" do
...
+          . mkInterpreter
+          . summaryWithExactly
+          $ Exactly
+          $  K (oneSecondEraSummary 0) -- Byron lasted 1 second
+          :* K (oneSecondEraSummary 1) -- Shelley lasted 1 second
+          :* K (oneSecondEraSummary 2) -- Allegra lasted 1 second
+          :* K (oneSecondEraSummary 3) -- Mary lasted 1 second
+          :* K (oneSecondEraSummary 4) -- Alonzo lasted 1 second
+          :* K (unboundedEraSummary 5) -- Babbage never ends
+          :* Nil
```
jhbertra, line 277: We need to move this to a helper module at some point

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
```diff
@@ -223,3 +236,88 @@ withdrawSpec = Hspec.describe "buildWithdrawConstraints" do
...
+    Hspec.QuickCheck.prop "valid slot interval for timed-out contract" do
+      address <- arbitrary
+      assets <- arbitrary
+      utxo <- arbitrary
```
jhbertra, line 281: The other generated values look fine because they just generate integers which don't really benefit from shrinking, but `assets` and `utxo` could definitely benefit from shrinking here. Can you rewrite this part of the test so that these will be shrunk if the test fails?

```suggestion
    Hspec.QuickCheck.prop "valid slot interval for timed-out contract" \assets utxo -> do
      address <- arbitrary
```

marlowe-runtime/tx/Language/Marlowe/Runtime/Transaction/BuildConstraints.hs
```diff
@@ -314,7 +314,7 @@ buildApplyInputsConstraintsV1 systemStart eraHistory marloweOutput tipSlot metad
     Nothing -> pure case nextMarloweTimeout contract of
       Nothing -> maxSafeSlot
       Just nextTimeout -> case utcTimeToSlotNo' nextTimeout of
-        Right slot -> slot
+        Right slot -> if slot > invalidBefore' then slot else maxSafeSlot
```
jhbertra, line 317: This isn't correct - we need to recursively find the next timeout in the continuation, not default to the max safe slot. I think the handling for this needs to be in `nextMarloweTimeout`, not here (we need to pass `invalidBefore'` to `nextMarloweTimeout`)

jhbertra (changes requested):
Thanks for adding the test, we really need to add a full test suite for this function. I think the updated logic is wrong though.

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
bwbush: Yes, and we don't even have Jira tickets yet for the comprehensive unit tests on this function.

marlowe-runtime/tx/Language/Marlowe/Runtime/Transaction/BuildConstraints.hs
bwbush: Thanks, I'll correct this and add a unit test targeting it.

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
jhbertra: https://input-output.atlassian.net/browse/SCP-5042

marlowe-runtime/tx/Language/Marlowe/Runtime/Transaction/BuildConstraints.hs
palas: you can find a verified calculation for a safe time interval (from the Marlowe perspective) in here: https://github.com/input-output-hk/marlowe/pull/140/files#diff-5103266280c0b8c38d09f6ed611b66de8bb9dc6c197fd67fbe0d9a2e18053427R695

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
```diff
@@ -223,3 +236,86 @@ withdrawSpec = Hspec.describe "buildWithdrawConstraints" do
...
+      pure
+        case result of
+            -- A valid transaction will occur if the tip is not before the timeout.
+            Right _ -> toSlot timeout <= tipSlot'
+            -- A useless transaction will occur if the tip is before the timeout.
+            Left (ApplyInputsConstraintsBuildupFailed (MarloweComputeTransactionFailed "TEUselessTransaction")) ->
+              tipTime < timeout
+            -- If rounding off causes the timeout to fall at the tip, then there is an ambiguity.
+            Left (ApplyInputsConstraintsBuildupFailed (MarloweComputeTransactionFailed "TEAmbiguousTimeIntervalError")) ->
+              tipSlot' == toSlot timeout && minTime <= timeout
```
palas, line 316: Wouldn't this also be the case if `tipSlot' >= timeout`? I guess it would fail if I was right, but I don't get it.

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
bwbush: I cleaned up the way this was coded, so it is easier to read.

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
bwbush: Thanks, I added the `counterexample`s.

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
bwbush: This is outdated. I rewrote the test logic so it is cleaner and more transparent.

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
```diff
@@ -223,3 +236,86 @@ withdrawSpec = Hspec.describe "buildWithdrawConstraints" do
...
+                   ]
+      timeout <- oneof
+                   [
+                     (1000 *) <$> chooseInteger (6, 20)  -- Even seconds, so there is a chance of collision.
+                   , chooseInteger (6000, 20000)         -- Milliseconds, so there is rounding off.
+                   ]
+      marloweParams <- arbitrary
+      let
+        toSlot = (`div` 1000)
+        tipSlot = toEnum . fromEnum $ tipSlot'
```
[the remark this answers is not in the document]
bwbush: Moving from integer to 64-bit word in the slot. I just committed a small change that makes this easier to read.

marlowe-runtime/tx/Language/Marlowe/Runtime/Transaction/BuildConstraints.hs
bwbush: Actually, I don't think we can use this reference code because of the slot <-> milliseconds roundoff and the fact that Marlowe uses closed intervals but Cardano uses half-open intervals.

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
```diff
@@ -223,3 +248,99 @@ withdrawSpec = Hspec.describe "buildWithdrawConstraints" do
...
+                   ]
+    Hspec.QuickCheck.prop "valid slot interval for timed-out contract" \assets utxo address marloweParams-> do
+      -- The choice intervals for the tip, minimum time, and timeout overlap, so every ordering will occur.
+      tipSlot' <- genTipSlot
+      minTime <- genMinTime
+      timeout <- genTimeout
+      timeout' <- genTimeout
+      let
+        contract = Semantics.When [] (POSIXTime timeout) Semantics.Close
+        contract' = Semantics.When [] (POSIXTime timeout) $ Semantics.When [] (POSIXTime timeout') Semantics.Close
```
bwbush, line 310: @palas and @jhbertra, should we add deeper test contracts here?
- *Pro:* it's a more comprehensive of timeout traversal.
- *Con:* this is more controlled that with a general arbitrary contract, where we'd more-or-less have to recode the "subsequent timeout" logic that we are testing. If we go with a general arbitrary contract, then I'd like to have a completely different timeout analysis in the test code, as compared to the code being tested.

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
```diff
@@ -223,3 +248,99 @@ withdrawSpec = Hspec.describe "buildWithdrawConstraints" do
...
+        $ case result of
+            Right _ ->
+              counterexample "A valid transaction will occur if tip is not before the first timeout."
+                $ toSlot timeout <= tipSlot'
+            Left (ApplyInputsConstraintsBuildupFailed (MarloweComputeTransactionFailed "TEUselessTransaction")) ->
+              counterexample "A useless transaction will occur if the tip is before the timeout."
+                $ tipTime < timeout
+            Left (ApplyInputsConstraintsBuildupFailed (MarloweComputeTransactionFailed message)) ->
+              if "TEIntervalError (IntervalInPastError " `isPrefixOf` message
+                then counterexample "If the tip is in the past, then the interval should be in the past."
```
palas, line 339: I think the statement in the counter example is not what we are testing but the converse, same for the next branch

bwbush: I believe it's an "if and only if", but the comment should read the same as the logic.

bwbush: I reworded so the statement matches the logic.

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
jhbertra: I don't think an arbitrary contract makes sense here - but we could potentially generate a contract using a custom generator (and `forallShrink`) which generates contracts appropriate for the property. This could also be tackled separately though.

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
```diff
@@ -223,3 +248,99 @@ withdrawSpec = Hspec.describe "buildWithdrawConstraints" do
...
+          buildApplyInputsConstraints
+            systemStart eraHistory MarloweV1
+            marloweOutput
+            tipSlot
+            (Chain.TransactionMetadata mempty) Nothing Nothing mempty
+      pure
+        . counterexample ("tipTime = " <> show tipTime)
+        . counterexample ("minTime = " <> show minTime)
+        . counterexample ("contract = " <> show marloweContract)
+        . counterexample ("result = " <> show result)
```
jhbertra, line 329: Nice :+1:

marlowe-runtime/tx/Language/Marlowe/Runtime/Transaction/BuildConstraints.hs
```diff
@@ -311,11 +311,19 @@ buildApplyInputsConstraintsV1 systemStart eraHistory marloweOutput tipSlot metad
...
+          else do                                                                   -- The next timeout is in the past.
+            invalidBefore'' <- slotStart invalidBefore'                             -- Find the start time of the validity range.
+            pure case nextMarloweTimeoutAfter invalidBefore'' contract of           -- Find the subsequent timeout, if any.
+              Nothing -> maxSafeSlot                                                -- There is no subsequent future timeout.
+              Just subsequentTimeout -> case utcTimeToSlotNo' subsequentTimeout of  -- There is a subsequent future timeout.
+                Right slot' -> slot'                                                -- The subsequent timeout is before the safe horizon, but note that this might be an empty interval.
+                _ ->  maxSafeSlot                                                   -- The subsequent timeout is beyond the safe horizon.
+        _ -> pure maxSafeSlot                                                       -- The next timeout is in the future, but beyond the safe horizon.
```
jhbertra, line 326: Is it not equivalent to eliminate the first `nextMarloweTimeout` check like so?

```suggestion
    Nothing -> do                                                                   -- The next timeout is in the past.
      invalidBefore'' <- slotStart invalidBefore'                             -- Find the start time of the validity range.
      pure case nextMarloweTimeoutAfter invalidBefore'' contract of           -- Find the subsequent timeout, if any.
        Nothing -> maxSafeSlot                                                -- There is no subsequent future timeout.
        Just subsequentTimeout -> case utcTimeToSlotNo' subsequentTimeout of  -- There is a subsequent future timeout.
          Right slot' -> slot'                                                -- The subsequent timeout is before the safe horizon, but note that this might be an empty interval.
          _ ->  maxSafeSlot                                                   -- The subsequent timeout is beyond the safe horizon.
```

In which case, instead of adding `nextMarloweTimeoutAfter` as an extra function, you could just add the `invalidBefore :: UTCTime` parameter to `nextMarloweTimeout`.

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
```diff
@@ -223,3 +248,99 @@ withdrawSpec = Hspec.describe "buildWithdrawConstraints" do
...
+      tipSlot' <- genTipSlot
+      minTime <- genMinTime
+      timeout <- genTimeout
+      timeout' <- genTimeout
+      let
+        contract = Semantics.When [] (POSIXTime timeout) Semantics.Close
+        contract' = Semantics.When [] (POSIXTime timeout) $ Semantics.When [] (POSIXTime timeout') Semantics.Close
+      marloweContract <- elements [contract, contract'] -- This contract can only time out.
+      let
+        toSlot = (`div` 1000)
```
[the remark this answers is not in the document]
bwbush: Fixed in 3939ac3b8c6ed306e789d930aefde6615a0efbfe

marlowe-runtime/tx/Language/Marlowe/Runtime/Transaction/BuildConstraints.hs
bwbush: Fixed. Thanks for your insight!

marlowe-runtime/test/Language/Marlowe/Runtime/Transaction/BuildConstraintsSpec.hs
palas: My thinking was that the case and nested if basically checks that the tip is in the past only if there is an error. So the tip could be in the past and, if there isn't an error, it wouldn't be checked that it isn't in the past, I think

palas: Maybe I am misunderstanding

jhbertra (approved):
LGTM, thanks!

PR #454, opened 2023-02-07
SCP-5043 Reviewing criteria for PRs.

bwbush:
This PR adds reviewing criteria to the template for pull requests.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,7 +3,15 @@ Here are some checklists you may like to use. Use your judgement.
 
 This is just a checklist, all the normative suggestions are covered in more detail in CONTRIBUTING.
 -->
+
 Pre-submit checklist:
+- Review required
+    - [ ] Substantial changes to code, test, or documentation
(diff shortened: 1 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -3,7 +3,15 @@ Here are some checklists you may like to use. Use your judgement.
 
 This is just a checklist, all the normative suggestions are covered in more detail in CONTRIBUTING.
 -->
+
 Pre-submit checklist:
+- Review required
+    - [ ] Substantial changes to code, test, or documentation
+    - [ ] Change made to Marlowe validator (@bwbush and @palas must be included as reviewers)
+- Review not required
+    - [ ] Pair programming or equivalent review outside of PR comments
```
jhbertra, line 12: I actually think we should still require reviews on the results of pair programming from someone with fresh eyes. My experience has been that while it does result in a lower bus factor and better design, it changes the way you think, and can sometimes result in odd decisions you wouldn't make on your own. I always double-check anything I've written while pair-programming for awkward things that came about as a result of incoherent verbal thought processes.

PR #457, opened 2023-02-07
SCP-4878 GetContractState query

jhbertra:


marlowe-runtime/sync-api/Language/Marlowe/Protocol/Query/Types.hs
```diff
@@ -111,6 +125,53 @@ data Page a b = Page
...
+
+data ContractState v = ContractState
+  { contractId :: ContractId
+  , roleTokenMintingPolicyId :: PolicyId
+  , metadata :: TransactionMetadata
+  , initialBlock :: BlockHeader
+  , initialOutput :: TransactionScriptOutput v
+  , latestBlock :: BlockHeader
+  , latestOutput :: Maybe (TransactionScriptOutput v)
+  , availablePayouts :: Map TxOutRef (Payout v)
```
bwbush, line 167: Is `availablePayouts` the best terminology? Would `unclaimedPayouts` or `unspentPayouts` be better?

marlowe-runtime/sync/Language/Marlowe/Runtime/Sync/Database/PostgreSQL/GetContractState.hs
```diff
@@ -0,0 +1,292 @@
...
+      JOIN marlowe.applyTx USING (txId)
+      LEFT JOIN marlowe.withdrawalTxIn
+        ON withdrawalTxIn.payoutTxId = payoutTxOut.txId
+        AND withdrawalTxIn.payoutTxIx = payoutTxOut.txIx
+      WHERE applyTx.createTxId = $1 :: bytea
+        AND applyTx.createTxIx = $2 :: smallint
+        AND withdrawalTxIn.txId IS NULL
+      GROUP BY payoutTxOut.txId, payoutTxOut.txIx
+      ORDER BY payoutTxOut.txId, payoutTxOut.txIx
+    |]
```
bwbush, line 129: We actually have a choice here between two behaviors, in situations where several contracts use the same policy ID for role tokens.
1. This query only returns the available payouts resulting from this particular contract, which is what happens here.
2. This query returns all payouts from the contract's policy ID.

Eventually, we might want a separate query that provides the second result, since that would be convenient for wallet-centric use cases.

marlowe-runtime/sync/Language/Marlowe/Runtime/Sync/Database/PostgreSQL/GetContractState.hs
```diff
@@ -0,0 +1,292 @@
...
+        ARRAY_REMOVE(ARRAY_AGG(txOutAsset.quantity), NULL) :: bigint[]
+      FROM marlowe.createTxOut
+      JOIN marlowe.contractTxOut USING (txId, txIx)
+      JOIN marlowe.txOut USING (txId, txIx)
+      LEFT JOIN marlowe.txOutAsset USING (txId, txIx)
+      WHERE createTxOut.txId = $1 :: bytea
+        AND createTxOut.txIx = $2 :: smallint
+      GROUP BY createTxOut.txId, createTxOut.txIx
+    |]
+  (latestBlock, latestOutput) <- lift $ maybe (initialBlock, Just initialOutput) decodeLatestResults <$> T.statement params [maybeStatement|
```
bwbush, line 71: So latest = initial for the creation tranaction, which is what we want.

marlowe-runtime/sync-api/Language/Marlowe/Protocol/Query/Types.hs
jhbertra: Let's decide to adopt the terminology of `claimed` vs `unclaimed`.

marlowe-runtime/sync/Language/Marlowe/Runtime/Sync/Database/PostgreSQL/GetContractState.hs
jhbertra: Exactly

PR #458, opened 2023-02-07
SCP-5047 Remove `marlowe ls` command

jhbertra:
Also removes `add` and `rm`
Not updated: jupyter notebooks


bwbush (approved):
Looks good.

@joseph-fajen and @ladamesny, please note that the `ls`, `add`, and `rm` commands have been removed from the `marlowe` executable. This affects the documentation and the test planning.

PR #460, opened 2023-02-08
SCP-4880 GetTransaction Query

jhbertra:


bwbush (approved):
The documentation needs to make it crystal clear that `GetTransaction` only looks up Marlowe-related transactions, not any transaction on the chain, and that one needs to use `FindTx` for general query.

PR #461, opened 2023-02-08
Add handshakes to all protocols.

jhbertra:
- [x] Adds a new `Handshake` higher-order protocol.
- [x] Updates all executables to use the handshake-wrapped protocol.

This will ensure that if we accidentally connect two peers that speak incompatible protocols, we will get a failure with an easy to understand message instead of unpredictable behaviour.


marlowe-chain-sync/src/Language/Marlowe/Runtime/ChainSync/Api.hs
```diff
@@ -84,6 +86,9 @@ data WithGenesis a = Genesis | At a
   deriving stock (Show, Eq, Ord, Functor, Generic)
   deriving anyclass (Binary, ToJSON)
 
+instance HasSignature a => HasSignature (WithGenesis a) where
+  signature _ = T.intercalate " " ["WithGenesis", signature $ Proxy @a]
```
bwbush, line 90: Nice!

bwbush (approved):
It's great to have this handshake capability.

Do we need documentation to make sure that the `HasSignature` instances are updated when future capabilities are added, or existing ones changed? It looks like type system will make it hard to forget, so I'm not sure.

jhbertra:
I think these will be fairly hands-off. We might want to extend them in the future to add some kind of message binary schema, but that is a much bigger change.

PR #463, opened 2023-02-09
SCP-4993 GetTransactions

jhbertra:


marlowe-runtime/sync-api/Language/Marlowe/Protocol/Query/Server.hs
```diff
@@ -32,6 +33,7 @@ marloweQueryServer getContractHeaders getContractState getTransaction = go
     serviceRequest :: Request a -> m a
     serviceRequest = \case
       ReqContractHeaders range -> getContractHeaders range
-      ReqContractState range -> getContractState range
-      ReqTransaction range -> getTransaction range
+      ReqContractState contractId -> getContractState contractId
```
bwbush, line 36: So the API changed. I agree it is cleaner this way.

marlowe-runtime/sync/Language/Marlowe/Runtime/Sync/Database/PostgreSQL/GetTransactions.hs
```diff
@@ -0,0 +1,136 @@
...
+        AND txOut.txIx = applyTx.outputTxIx
+      LEFT JOIN marlowe.txOutAsset
+        ON txOutAsset.txId = applyTx.txId
+        AND txOutAsset.txIx = applyTx.outputTxIx
+      WHERE applyTx.createTxId = $1 :: bytea
+        AND applyTx.createTxId = $2 :: smallint
+      GROUP BY applyTx.txId
+      ORDER BY (ARRAY_AGG(applyTx.inputTxId))[1], (ARRAY_AGG(applyTx.inputTxIx))[1]
+    |]
+
```
bwbush, line 84: I think we'll need performance testing on `mainnet`. I've noticed that some asset-related queries perform poorly there, and this query looks superficially similar to some queries that I've had difficulty with.

bwbush (approved):
Looks great. I'll make sure we have a Jira ticket to test performance on the large public networks.

marlowe-runtime/sync-api/Language/Marlowe/Protocol/Query/Server.hs
jhbertra: No, just some misnamed variables

marlowe-runtime/sync/Language/Marlowe/Runtime/Sync/Database/PostgreSQL/GetTransactions.hs
jhbertra: Yes for sure - FYI I always analyze the queries with `EXPLAIN` to make sure the indexes are being used properly, there are no expensive sequential scans or sorts happening, etc. All the queries here have very good asymptotic complexity. But you're right, we'll want to check on a large network.

marlowe-runtime/sync-api/Language/Marlowe/Protocol/Query/Server.hs
bwbush: Ah

palas (approved):
Looks good to me :+1:

PR #466, opened 2023-02-09
Improve Integration Test Stability

jhbertra:
- [x] Add a new option to `marlowe-tx` for waiting a number of blocks before confirming a transaction
- [x] Set the option to `2` for the integration tests.

### DIFF
--- a/marlowe-integration/src/Test/Integration/Marlowe/Local.hs
+++ b/marlowe-integration/src/Test/Integration/Marlowe/Local.hs
@@ -285,7 +285,7 @@ withLocalMarloweRuntime' MarloweRuntimeOptions{..} test = withRunInIO \runInIO -
     webPort <- liftIO $ randomRIO (4000, 5000)
     manager <- liftIO $ newManager defaultManagerSettings
 
-    let submitConfirmationBlocks = 2
+    let submitConfirmationBlocks = 5
     let mkSubmitJob = Submit.mkSubmitJob SubmitJobDependencies{..}
     let baseUrl = BaseUrl Http "localhost" webPort ""
     let clientEnv = mkClientEnv manager baseUrl
--- a/marlowe-integration/src/Test/Integration/Marlowe/Local.hs
+++ b/marlowe-integration/src/Test/Integration/Marlowe/Local.hs
@@ -285,6 +285,7 @@ withLocalMarloweRuntime' MarloweRuntimeOptions{..} test = withRunInIO \runInIO -
     webPort <- liftIO $ randomRIO (4000, 5000)
     manager <- liftIO $ newManager defaultManagerSettings
 
+    let submitConfirmationBlocks = 2
     let mkSubmitJob = Submit.mkSubmitJob SubmitJobDependencies{..}
     let baseUrl = BaseUrl Http "localhost" webPort ""
     let clientEnv = mkClientEnv manager baseUrl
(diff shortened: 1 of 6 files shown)


marlowe-integration/src/Test/Integration/Marlowe/Local.hs
```diff
@@ -285,6 +285,7 @@ withLocalMarloweRuntime' MarloweRuntimeOptions{..} test = withRunInIO \runInIO -
     webPort <- liftIO $ randomRIO (4000, 5000)
     manager <- liftIO $ newManager defaultManagerSettings
 
+    let submitConfirmationBlocks = 2
```
bwbush, line 288: If we still have troubles with the integration tests on the fast network, we can increase this further.

marlowe-runtime/marlowe-tx/Main.hs
```diff
@@ -244,6 +247,14 @@ getOptions = execParser $ info (helper <*> parser) infoMod
       , help "The logging configuration JSON file."
       ]
 
+    submitConfirmationBlocksParser = option (BlockNo <$> auto) $ mconcat
+      [ long "submit-confirmation-blocks"
```
bwbush, line 251: Nice!

marlowe-integration/src/Test/Integration/Marlowe/Local.hs
jhbertra: Yeah, this is our knob. This is an "optimistic" initial value

marlowe-runtime/marlowe-tx/Main.hs
```diff
@@ -244,6 +247,14 @@ getOptions = execParser $ info (helper <*> parser) infoMod
...
+      , value 0
```
bwbush, line 252: Yes, the default should be zero because some clients will want to tune this themselves.

jhbertra: It also preserves the current behaviour by default

jhbertra:
Reopen to try and create the jobset on Hydra

marlowe-runtime/tx/Language/Marlowe/Runtime/Transaction/Submit.hs
```diff
@@ -61,18 +63,38 @@ doSubmit SubmitJobDependencies{..} tellStatus era tx= do
...
-          , recvMsgRollForward = \_ point _ -> case point of
+          , recvMsgRollBackward = \_ _ -> pure clientIdleAwaitConfirmation
+          , recvMsgRollForward = \_ point tip -> case (point, tip) of
+              (At block, At tipBlock) -> pure $ clientIdleAwaitMaturity block tipBlock
+              _ -> error "chainseekd rolled forward to genesis"
+          , recvMsgWait = threadDelay 100_000 $> SendMsgPoll clientNextAwaitConfirmation
+          }
+
+        clientIdleAwaitMaturity block tipBlock
+          | blockNo tipBlock - blockNo block >= submitConfirmationBlocks = SendMsgDone block
```
bwbush, line 86: Should `WaitForTx` allow the global default confirmation to be overridden by clients?

jhbertra: Perhaps, we could add this as an option to the `Submit` command

bwbush: Let's hold off, since we might deprecate `Submit`.

bwbush (approved):

PR #467, opened 2023-02-09
SCP-4532 Harmonized program names and command-line options.

bwbush:
1. `--...--port-number` becomes `--...--port`.
2. `chainseekd` becomes `marlowe-chain-sync`.
3. References to "chain seek" in environment variables, documentation, and comments become references to "chain sync".
4. No changes to any Haskell type or function names are made.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


bwbush:
@jhbertra , I'm concerned that it's too late to change `chainseekd` to `marlowe-chain-sync` and that this might create chaos. Maybe we should just harmonize the port number flags? However, if we're going to do this, then we shouldn't delay.

palas:
It seems there are still some references to chainseek under the `deploy` folder, I would go all in. But yes @Pacman99, could you confirm that we are not going to break the deployments by changing it? Or if we do where do we need to change it too? 
These other 3 files that also still have references to chainseek:
- `marlowe-integration/src/Test/Integration/Marlowe/Local.hs`
- `marlowe-runtime/tx/Language/Marlowe/Runtime/Transaction/Submit.hs`
- `scripts/setup-dev-connection`

bwbush:
> It seems there are still some references to chainseek . . .

@palas, thanks. I 'll do another grep through the code. In avoiding changing any function or type identifiers, I might have missed so other text.

bwbush:
Hmm, our main branch is changing too fast: each rebase triggers a full re-review of the changes needed.

marlowe-chain-sync/docs/marlowe-chain-sync.1
```diff
@@ -1,24 +1,24 @@
-.TH chainseekd 1 "18 July, 2022" "version 0.0.0.0"
+.TH marlowe-chain-sync 1 "18 July, 2022" "version 0.0.0.0"
```
jhbertra, line 1: Go ahead and delete all these manpage documents. The have not been maintained and are not used.

palas (approved):
Thanks, looks good :+1:

marlowe-chain-sync/docs/marlowe-chain-sync.1
bwbush: Fixed in 95482d61461f42d5f7abe4850836f26e246b94b3. It looks like we only had one file of this type, but I remember that we used to have more.

Pacman99 (approved):

PR #475, opened 2023-02-13
Reload EraHistory in `marlowe-indexer`

jhbertra:
We previously loaded the `EraHistory` once on startup in `marlowe-indexer`, but that was a mistake because the era history changes every epoch as the safe zone is extended. Now it always loads a fresh copy from `chainseekd` when it needs it.


bwbush (approved):
Nice. I'm glad it was straightforward to remedy this.

PR #489, opened 2023-02-22
Delete rollback blocks when saving blocks

jhbertra:
Here is an issue that impacts `marlowe-indexer` when the node rolls back to genesis. Old blocks get marked as rolled back to Genesis, but these are not removed when adding the blocks back later.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
        - Review required
    - [x] Reviewer requested


bwbush (approved):
Since `marlowe-chain-sync` handles rollbacks differently in the schema, I assume that this same problem doesn't exist there, too.

jhbertra:
> Since `marlowe-chain-sync` handles rollbacks differently in the schema, I assume that this same problem doesn't exist there, too.

Correct

PR #494, opened 2023-02-23
Refactor connection code

jhbertra:
This PR includes several changes to how the protocol peers are connected to channels:

- [x] Remove explicit `Codecs` in favour of `BinaryMessage` typeclass (makes composition for higher-order protocols (e.g. `Handshake` much easier)
- [x] Remove `throwImpl` provided to `mkDriver` and use `throwIO` directly.
- [x] Introduce composable primatives for defining connection sources and replace fixed-functionality connection functions
  - `Connection`: A peer with a channel and a close function, ready to be run.
  - `Connector`: A function that receives a peer and creates a `Connection`
  - `ConnectionSource`: An STM action that yields `Connectors` (used to implement servers).

Additionally, because `ConnectionSource`s can be merged to create a new `ConnectionSource`, this allows servers that listen on multiple channels. This is used in the integration tests where the protocols can be connected to via STM channels, as well as sockets (which is necessary for CLI testing). The integration tests now bind to TCP ports and pass the port numbers to the integration test for use in e.g. calling the CLI.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
        - Review required
    - [x] Reviewer requested


marlowe-runtime/sync-api/Language/Marlowe/Protocol/Query/Codec.hs
```diff
@@ -1,66 +1,2 @@
-{-# LANGUAGE GADTs #-}
 module Language.Marlowe.Protocol.Query.Codec
```
bwbush, line 1: Why not remove this whole module?

marlowe-protocols/src/Network/Protocol/Connection.hs
```diff
@@ -0,0 +1,207 @@
...
+  forall ps. BinaryMessage ps => SomeConnectionSource (ConnectionSource ps server m)
+
+data ConnectorSelector ps f where
+  Session :: ConnectorSelector ps Void
+  ConnectionSelector :: ConnectionSelector ps f -> ConnectorSelector ps f
+
+getConnectorSelectorConfig
+  :: MessageToJSON ps
+  => Bool
+  -> Bool
```
bwbush, line 63: The `Bool`s are a bit opaque when these functions are called. Maybe it is worthwhile to add a comment in the two places where these are called with `False`?

marlowe-runtime/sync-api/Language/Marlowe/Protocol/Query/Codec.hs
jhbertra: Done

bwbush (approved):
Nice. This is quite a bit cleaner and more functional.

I tested this on `preview`:
- 20 clients running 10 four-step contracts sequentially: 100% success, with no retries.
- `marlowe-finder` for the full history of contracts: no problems

Note that these files contain some lint:
```bash
hlint marlowe-runtime/ marlowe-protocols/
```

PR #501, opened 2023-02-27
SCP-5098 Use node chain-sync client for submissions and queries

bwbush:
Instead of connecting to a new node socket for each submission or query, `malrowe-chain-sync` now keeps a connection open to the node and uses submit and query clients to submit, where each of those two clients reads from a separate STM channel.

This was tested by running 500 clients in parallel against a single Marlowe Runtime deployment, with `marlowe-scaling` settings so that the backend was hit with requests at the maximum rate and so that the node's memory pool and the chain's blocks were full. No failures occurred aside from a rollback that the `marlowe-scaling` client did not handle properly.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
        - Review required
        - Review not required
    - [x] Reviewer requested


marlowe-chain-sync/libchainsync/Language/Marlowe/Runtime/ChainSync/NodeClient.hs
```diff
@@ -0,0 +1,272 @@
...
+                    -- the node would close the connection in about three seconds,
+                    -- making it impossible to submit and query.
+                    localChainSyncClient = LocalChainSyncClient chainSyncClient
+                  , localTxSubmissionClient = Just $ submitClient submitChannel
+                  , localTxMonitoringClient = Nothing
+                  , localStateQueryClient = Just $ queryClient queryChannel
+                  }
+            case result of
+              Left ex -> failEvent ev ex *> throw ex
+              Right _ -> pure ()
```
jhbertra, line 128: This will keep the `Connect` event active for the entire run of the app, are you sure this is what we want? If it is, you can use `withEvent`
```suggestion
        runNodeClient = withEvent eventBackend Connect \_ -> connectToLocalNode LocalNodeClientProtocols
          {
            -- We don't use the block and tip information, but the chain
            -- sync client keeps the connection open. Without this client,
            -- the node would close the connection in about three seconds,
            -- making it impossible to submit and query.
            localChainSyncClient = LocalChainSyncClient chainSyncClient
          , localTxSubmissionClient = Just $ submitClient submitChannel
          , localTxMonitoringClient = Nothing
          , localStateQueryClient = Just $ queryClient queryChannel
          }
```

marlowe-chain-sync/libchainsync/Language/Marlowe/Runtime/ChainSync/NodeClient.hs
```diff
@@ -0,0 +1,272 @@
...
+type QueryChannel = TChan (Q.Some QueryJob)
+
+
+data QueryJob result =
+  QueryJob
+  {
+    point :: Maybe ChainPoint
+  , query :: QueryInMode CardanoMode result
+  , queryResultTMVar :: TMVar (Either Q.AcquireFailure result)
+  }
```
jhbertra, line 192: Rather than adding the additional `Some` wrapper, you could also existentially quantify `result` within the `QueryJob` constructor

```suggestion
data QueryJob = forall result.
  QueryJob
  {
    point :: Maybe ChainPoint
  , query :: QueryInMode CardanoMode result
  , queryResultTMVar :: TMVar (Either Q.AcquireFailure result)
  }
```

jhbertra (approved):
Nice, just a couple small suggestions on my end. We may want to consider changing this single connection for a pool if it proves an unnecessary bottleneck, but this is a good place to start!

marlowe-chain-sync/libchainsync/Language/Marlowe/Runtime/ChainSync/NodeClient.hs
bwbush: I'll make this change. I actually just adopted the strategy that `marlowe-chain-indexer` was using: https://github.com/input-output-hk/marlowe-cardano/blob/1e778adb7f1b8e5087bdddabb034208626136ced/marlowe-chain-sync/chain-indexer/Language/Marlowe/Runtime/ChainIndexer/NodeClient.hs#L152.

marlowe-chain-sync/libchainsync/Language/Marlowe/Runtime/ChainSync/NodeClient.hs
bwbush: Nice! Thanks, fixed in 847d17a911fd63d51230b1c439eb1ceb0d966f5c.

marlowe-chain-sync/libchainsync/Language/Marlowe/Runtime/ChainSync/NodeClient.hs
bwbush: Fixed in c642758800a2cd902b0e2a324ffb8ded6211c738.

bwbush:
> We may want to consider changing this single connection for a pool if it proves an unnecessary bottleneck

There's no evidence of a bottleneck, even at 500 simultaneous requests. I think it would be better to throttle at the proxy level, and only if the OS doesn't have sufficient memory, sockets, or handles available.

PR #507, opened 2023-02-28
SCP-5126 MList annotations and tests

bwbush:
1. Discussion and justification for using `PlutusTx.AssocMap` instead of Isabelle's `MList` were added as comments to all usages in Marlowe's semantics implementation.
2. A paragraph was added to the Marlowe-Cardano specification about `MList`.
3. Property-based tests were added for all `AssocMap` functions used the Marlowe validator, to verify that if there are no duplicates entries and the `MList` is sorted, then the behavior matches `AssocMap`.
4. The change log was updated.

***Reviewers: Does this adequately address the following audit findings?***

image

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
        - Review required
        - [x] Change made to Marlowe validator (@bwbush and @palas must be included as reviewers)
        - Review not required
    - [x] Reviewer requested

### DIFF
--- a/marlowe-test/src/Spec/Marlowe/Plutus/MList.hs
+++ b/marlowe-test/src/Spec/Marlowe/Plutus/MList.hs
@@ -1,0 +1,298 @@
+-----------------------------------------------------------------------------
+--
+-- Module      :  $Headers
+-- License     :  Apache 2.0
+--
+-- Stability   :  Experimental
+-- Portability :  Portable
+--
+-- | Compare the `PlutusTx.AssocMap` to Isabelle's `MList`.
+--
+-----------------------------------------------------------------------------
+
+
+{-# ANN module "HLint: Use guards" #-}
+
+
+module Spec.Marlowe.Plutus.MList
+  ( -- * Testing
+    tests
+  ) where
+
+
+import Prelude hiding (lookup)
+import Data.Function (on)
+import Data.Maybe (isJust, fromMaybe)
+import Data.List (sortBy, nubBy)
+import Test.Tasty (TestTree, testGroup)
+import Test.Tasty.HUnit (Assertion, testCase, assertBool)
+import Test.Tasty.QuickCheck (Arbitrary(..), Gen, Property, elements, forAll, property, testProperty, shuffle, elements)
+
+import qualified PlutusTx.AssocMap as AM (Map, delete, empty, fromList, insert, lookup, member, null, singleton, toList, unionWith)
+
+
+-- | An association list in Isabelle.
+type MList a b = [(a, b)]
+
+
+-- | Empty Isabelle `MList`.
+--
+-- @
+-- definition empty :: "('a::linorder \<times> 'b) list" where
+--   "empty = Nil"
+-- @
+empty :: MList a b
+empty = []
+
+
+-- | Insert into Isabelle `MList`.
+--
+-- @
+-- fun insert :: "'a::linorder \<Rightarrow> 'b \<Rightarrow> ('a \<times> 'b) list \<Rightarrow> ('a \<times> 'b) list" where
+--   "insert a b Nil = Cons (a, b) Nil" |
+--   "insert a b (Cons (x, y) z) =
+--     (if a < x
+--      then (Cons (a, b) (Cons (x, y) z))
+--      else (if a > x
+--            then (Cons (x, y) (insert a b z))
+--            else (Cons (x, b) z)))"
+-- @
+insert :: Ord a => a -> b -> MList a b -> MList a b
+insert a b [] = (a, b) : []
+insert a b ((x, y) : z) =
+  if a < x
+    then (a, b) : (x, y) : z
+    else if a > x
+      then (x, y) : insert a b z
+      else (x, b) : z
+
+
+-- | Delete from Isabelle `MList`.
+--
+-- @
+-- fun delete :: "'a::linorder \<Rightarrow> ('a \<times> 'b) list \<Rightarrow> ('a \<times> 'b) list" where
+--   "delete a Nil = Nil" |
+--   "delete a (Cons (x, y) z) =
+--     (if a = x
+--      then z
+--      else (if a > x
+--            then (Cons (x, y) (delete a z))
+--            else (Cons (x, y) z)))"
+-- @
+delete :: Ord a => a -> MList a b -> MList a b
+delete _a [] = []
+delete a ((x, y) : z) =
+  if a == x
+    then z
+    else if a > x
+      then (x, y) : delete a z
+      else (x, y) : z
+
+
+-- | Lookup in Isabelle `MList`.
+--
+-- @
+-- fun lookup :: "'a::linorder \<Rightarrow> ('a \<times> 'b) list \<Rightarrow> 'b option" where
+--   "lookup a Nil = None" |
+--   "lookup a (Cons (x, y) z) =
+--     (if a = x
+--      then Some y
+--      else (if a > x
+--            then lookup a z
+--            else None))"
+-- @
+lookup :: Ord a => a -> MList a b -> Maybe b
+lookup _a [] = Nothing
+lookup a ((x, y) : z) =
+  if a == x
+    then Just y
+    else if a > x
+      then lookup a z
+      else Nothing
+
+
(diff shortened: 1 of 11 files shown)


marlowe-test/src/Spec/Marlowe/Plutus/MList.hs
```diff
@@ -0,0 +1,298 @@
...
+--
+-- Stability   :  Experimental
+-- Portability :  Portable
+--
+-- | Compare the `PlutusTx.AssocMap` to Isabelle's `MList`.
+--
+-----------------------------------------------------------------------------
+
+
+{-# ANN module "HLint: Use guards" #-}
```
bwbush, line 14: This breaks the stylish-Haskell check.

marlowe/src/Language/Marlowe/Core/V1/Semantics.hs
```diff
@@ -505,6 +520,10 @@ evalObservation env state obs = let
 refundOne :: Accounts -> Maybe ((Party, Token, Integer), Accounts)
 refundOne accounts = case Map.toList accounts of
     [] -> Nothing
+    -- SCP-5126: The return value of this function differs from
+    -- Isabelle semantics in that it returns the least-recently
+    -- added account-token combination rather than the first
+    -- lexicographically ordered one.
     ((accId, token), balance) : rest ->
         if balance > 0
         then Just ((accId, token, balance), Map.fromList rest)
```
palas, line 529: Should we say something about `Map.fromList . tail . Map.toList` preserving the invariant?

palas (approved):
Thanks, looks good, just a comment

marlowe-test/src/Spec/Marlowe/Plutus/MList.hs
```diff
@@ -0,0 +1,305 @@
...
+  let
+    gen = do
+      mlist <- arbitraryMList
+      assocmap <- AM.fromList <$> shuffle mlist
+      pure (mlist, assocmap)
+  forAll gen
+    $ \(mlist, assocmap) -> (== empty) mlist == AM.null assocmap
+
+
+-- | Compare `empty` for `MList` and `AssocMap`.
```
hrajchert, line 205: this comment is off

marlowe-test/src/Spec/Marlowe/Plutus/MList.hs
```diff
@@ -0,0 +1,305 @@
...
+  let
+    gen = do
+      a <- arbitrary :: Gen Integer
+      b <- arbitrary :: Gen [()]
+      pure (a, b)
+  forAll gen
+    $ \(a, b) -> [(a, b)] `equivalent` AM.singleton a b
+
+
+-- | Compare `empty` for `MList` and `AssocMap`.
```
hrajchert, line 217: This comment is off

marlowe-test/src/Spec/Marlowe/Plutus/MList.hs
```diff
@@ -0,0 +1,305 @@
...
+
+-- | Generate a sorted `MList` with no duplicates.
+arbitraryMList :: Gen (MList Integer [()])
+arbitraryMList = nubBy ((==) `on` fst) . sortBy (compare `on` fst) <$> arbitrary
+
+
+-- | Compare an `MList` to an `AssocMap`.
+equivalent :: Ord a => Eq b => MList a b -> AM.Map a b -> Bool
+equivalent mlist assocmap = mlist == sortBy (compare `on` fst) (AM.toList assocmap)
+
```
hrajchert, line 186: instead of comparing by equality of the sorted AssocMap, shouldn't we check that for each entry in MList Assoc.lookup gives a result and vice-versa?

This way is not checking for duplicate entries in AssocMap (see my comment on insert)

marlowe-test/src/Spec/Marlowe/Plutus/MList.hs
```diff
@@ -0,0 +1,305 @@
...
+    $ \(a, b) -> [(a, b)] `equivalent` AM.singleton a b
+
+
+-- | Compare `empty` for `MList` and `AssocMap`.
+checkInsert :: Property
+checkInsert = property $ do
+  let
+    gen = do
+      mlist <- arbitraryMList
+      assocmap <- AM.fromList <$> shuffle mlist
```
hrajchert, line 223: I see that these tests start from MList and shuffle the AssocMap, shouldn't we do the opposite? Start by AssocMap, remove duplicates and sort for MList? The examples provided by the auditors are mostly preoccupied with duplicate entries or unsorted entries in MList, in these tests we are not covering that.

The Isabelle theorems `delete_valid` and `insert_valid` already prove that if the MList is valid before insert/delete it will be valid after

marlowe/specification/marlowe-cardano-specification.md
```diff
@@ -97,6 +97,17 @@ data Input =
...
+2.  If the elements of an `MList` are not in lexicographic order, then result of `insert`, `delete`, `member`, and `lookup` may differ from that in `AssocMap`.
+3.  The ordering of the result of `toList` may differ between `AssocMap` and `MList`.
+
+Provided that the initial state of the Marlowe contract does not contain duplicate entries in `accounts`, `choices`, or `boundValues` and also provided that the behavior of Marlowe's semantics validator is compared to Isabelle semantics with the Isabelle initial state for `accounts`, `choices`, and `boundValues` lexicographic ordered, the behavior of the Marlowe semantics validator does not differ from Isabelle semantics aside from the ordering of `accounts`, `choices`, and `boundValues` in Marlowe's state in the datum. Typically, Marlowe contracts start with an empty state, so this is not an issue, but one can in principle start a Marlowe contract with a non-empty state, even with one that contains duplicate entries. Internally to the validator, `computeTransaction` might order payments upon `Close` differently from Isabelle semantics, but that is undetectable at the validation or transaction level.
```
hrajchert, line 108: > Typically, Marlowe contracts start with an empty state, so this is not an issue, but one can in principle start a Marlowe contract with a non-empty state, even with one that contains duplicate entries

Can't we prevent this? Maybe we should add as a constraint in the validator that there are no duplicate entries in the state and the values are positive (for accounts). Even if the semantics are supposed to preserve that (if we start from a valid point)

marlowe/src/Language/Marlowe/Core/V1/Semantics.hs
bwbush: Thanks! Fixed in e2277677987f359dcd8754275eccd5b7c9a880a7.

marlowe-test/src/Spec/Marlowe/Plutus/MList.hs
bwbush: Thanks! Fixed in a95a616aca59bc7159d3ffa92487f8f021820bff.

marlowe-test/src/Spec/Marlowe/Plutus/MList.hs
bwbush: Fixed in a95a616aca59bc7159d3ffa92487f8f021820bff.

marlowe/specification/marlowe-cardano-specification.md
bwbush: > Maybe we should add as a constraint in the validator that there are no duplicate entries in the state and the values are positive (for accounts).

I plan to measure the impact of adding this check on validator size, but I'm not optimistic that the effect will be small.

marlowe-test/src/Spec/Marlowe/Plutus/MList.hs
bwbush: Fixed in b65ccfec3708484ff52514671e0e227c13f084f3 by including `lookup` tests, too.

Note that the testing is preconditioned on the `MList` being sorted and neither list containing duplicates. I'll noted that in a revised comment.

marlowe-test/src/Spec/Marlowe/Plutus/MList.hs
bwbush: Thanks. Fixed in 9e068b6a60103f2a40604a7eaea57c5f29b59180.

paluh (commented):
Looks great!

PR #516, opened 2023-03-02
Commit hie.yaml

jhbertra:
Adds this important file to the repo so that we all use the same cradle file to reduce discrepancies in development environments.


bwbush (approved):
I was under the impression that recent versions of HLS under Nix did not require this file at all.

hrajchert (approved):

jhbertra:
They can get confused - so it's better to provide an explicit one

PR #525, opened 2023-03-07
Upgrade eventuo11y to 0.9

shlevy:
Future work to include:

- Switching to `MonadEvent` instead of explicit backend passing
- Making it easier to derive `MonadWith` for newtypes (ideally fixing `deriving newtype`, if not perhaps some TH)


shlevy:
Trying to merge `main` locally now.

.github/workflows/test.yml
```diff
@@ -10,27 +10,6 @@ on:
       - main
   workflow_call:
 jobs:
-  check-materialization:
```
bwbush, line 13: I don't understand why this is removed. What does this have to do with `eventuo11y`?

dev-shell.nix
```diff
@@ -50,7 +46,7 @@ let
         # While nixpkgs-fmt does exclude patterns specified in `.ignore` this
         # does not appear to work inside the hook. For now we have to thus
         # maintain excludes here *and* in `./.ignore` and *keep them in sync*.
-        excludes = [ ".*nix/pkgs/haskell/materialized.*/.*" ".*/spago-packages.nix$" ".*/packages.nix$" ];
+        excludes = [ ".*/spago-packages.nix$" ".*/packages.nix$" ];
```
bwbush, line 49: FYI: We don't use spago in this repo any more.

marlowe-apps/marlowe-apps.cabal
```diff
@@ -30,9 +30,10 @@ library
                     , cardano-api
                     , containers
                     , data-default
```
bwbush, line 32: Thanks for cleaning up the logging usage in this package!

bwbush (approved):
Looks good.

I found it confusing that removing `updateMaterialized` was mixed into this PR.

.github/workflows/test.yml
jhbertra: Because apparently materialization does not actually speed up flake eval, which is the only reason we were doing it. So we're removing it so we have one fewer manual thing to manage.

shlevy:
Removed materialization after confirming that it wasn't winning anything because I kept having to go through materialization updates as I was trying to update the index state.

marlowe-runtime-web/server/Language/Marlowe/Runtime/Web/Server/Monad.hs
```diff
@@ -39,14 +42,25 @@ newtype AppM r a = AppM { runAppM :: ReaderT (AppEnv r) Handler a }
...
+      allocA' :: (forall x. ExceptT ServerError IO x -> ExceptT ServerError IO x) -> (ExceptT ServerError IO) (GeneralAllocated (ExceptT ServerError IO) (WithException (ExceptT ServerError IO)) releaseReturn b a)
+      allocA' restore = do
+        let
+          restore' :: forall x. AppM r x -> AppM r x
+          restore' mx = AppM . ReaderT $ coerce . restore . coerce . (runReaderT . runAppM) mx
+        GeneralAllocated a releaseA <- (coerce . runReaderT . runAppM) (allocA restore') r
+        let
+          releaseA' relTy = (coerce . runReaderT . runAppM) (releaseA relTy) r
+        pure $ GeneralAllocated a releaseA'
+    coerce $ stateThreadingGeneralWith (GeneralAllocate allocA') (flip (coerce . runReaderT . runAppM) r . go)
```
jhbertra, line 63: These `MonadWith` instances are pretty monstrous :fearful: Is there no way to wrap these newtype implementations up in reusable default functions?

jhbertra (approved):
LGTM, thanks!

marlowe-runtime-web/server/Language/Marlowe/Runtime/Web/Server/Monad.hs
shlevy: Yeah it's bad, it's on my list to fix. The deriving newtype/deriving via *should* work I think, there's something funky in GHC's unification.

If that's not fixable, it might be possible to have default implementations using a [ControlTransformation](https://hackage.haskell.org/package/eventuo11y-0.9.0.1/docs/Control-Natural-Control.html#t:ControlTransformation)

PR #532, opened 2023-03-10
Update plutus-apps.

shlevy:
These are the versions which will be released on CHaP.

hasql-pool upgrade needed due to version interactions causing cabal to select a broken version otherwise.


bwbush (approved):
Looks good.

I tested by running 10,000 Plutus executions, just to make sure that the change to the validator didn't adversely increase the execution cost.

PR #536, opened 2023-03-13
Plt 4216 Protocol Golden Tests

jhbertra:
Adds golden tests for all protocol messages. This is to prevent us accidentally releasing API-breaking changes in the binary encoding of the messages.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
        - Review required
    - [x] Reviewer requested


marlowe-chain-sync/.golden/ChainSeek/golden
```diff
@@ -0,0 +1,516 @@
+Show: MsgQueryNext (AdvanceBlocks 1)
```
bwbush, line 1: Should these really be in hidden folders? That might mystify users.

marlowe-runtime-web/server/Language/Marlowe/Runtime/Web/Server/REST/ApiError.hs
```diff
@@ -125,7 +125,6 @@ instance ToDTO (WithdrawError 'V1) where
     WithdrawConstraintError (BalancingError _) -> ApiError "Internal error" "InternalError" Null 500
     WithdrawLoadMarloweContextFailed LoadMarloweContextErrorNotFound -> ApiError "Marlowe contract not found" "MarloweContractNotFound" Null 404
     WithdrawLoadMarloweContextFailed (LoadMarloweContextErrorVersionMismatch _) -> ApiError "Marlowe contract version mismatch" "MarloweContractVersionMismatch" Null 400
-    WithdrawLoadMarloweContextFailed (HandshakeFailed _) -> ApiError "Internal error" "InternalError" Null 500
     WithdrawLoadMarloweContextFailed LoadMarloweContextToCardanoError -> ApiError "Internal error" "InternalError" Null 500
```
bwbush, line 128: It would be helpful if all of the error messages described the actual type instead of just `"InternalError"`. Some of these messages have that detail, but others don't.

bwbush (approved):
The PR description isn't complete: this does significant reorganization within the API and its implementation, including removing functionality such as `FindTxsTo`.

Also, I'm a bit concerned that we're making this reorganization before the e2e tests are all in place. I'll run the suite of manual tests.

Why was it necessary to remove the query `FindTxsTo`, which is useful for certain types of discovery?

marlowe-runtime-web/server/Language/Marlowe/Runtime/Web/Server/REST/ApiError.hs
jhbertra: Agreed - but definitely not in the scope of this PR

marlowe-chain-sync/.golden/ChainSeek/golden
jhbertra: This is where `hspec-golden` puts the golen test data files by default - though we could certainly change it.

PR #542, opened 2023-03-21
PLT-5286 Adjust default logging configuration for each component and document

jhbertra:
- [x] Adds `--print-log-config` option for each runtime service
- [x] Disables all `send` `recv` and `database` logs by default
- [x] Adds documentation for the log config.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
        - Review required
    - [x] Reviewer requested

### DIFF
--- a/deploy/operables.nix
+++ b/deploy/operables.nix
@@ -167,7 +167,7 @@ in
 
       ${wait-for-socket}/bin/wait-for-socket "$CARDANO_NODE_SOCKET_PATH"
 
-      ${packages.marlowe-chain-indexer}/bin/marlowe-chain-indexer $([ -z $LOG_CONFIG_FILE ] || echo --log-config-file "$LOG_CONFIG_FILE") \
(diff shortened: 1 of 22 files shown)


bwbush (approved):
Wonderful! I think these are appropriate defaults for production deployment. For development and testing, we can turn up the verbosity.

I suppose there is no type-safe way to ensure (via compiler warnings) that we remember to add the correct entries to things like `defaultRootSelectorLogConfig` if we later expand the logging?

jhbertra:
> I suppose there is no type-safe way to ensure (via compiler warnings) that we remember to add the correct entries to things like `defaultRootSelectorLogConfig` if we later expand the logging?

Alas, no. Our options for this are:

1. Generate this stuff with template Haskell
2. Use a row-based approach to specifying selectors instead of GADTs (i.e. extensible Variants and records for selectors and config, respectively).

PR #550, opened 2023-03-28
Plt 5269 create tests

jhbertra:
Test suite for the `Create` command. Also fixed a couple minor bugs found in the code.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
        - Review required
    - [x] Reviewer requested


marlowe-integration-tests/test/Language/Marlowe/Runtime/Integration/Common.hs
```diff
@@ -87,11 +117,118 @@ import UnliftIO.Environment (setEnv, unsetEnv)
...
+    pure (address, WitnessPaymentKey signingKey, collateralUtxos)
+  pure Wallet
+    { addresses = WalletAddresses (head addresses) (Set.fromList $ tail addresses) (fold collateralUtxos)
+    , signingKeys
+    }
+
+submitBuilder :: Wallet -> TxBuilder a -> Integration (BlockHeader, a)
+submitBuilder wallet builder = mdo
+  -- Note - the txId is not evaluated yet - we're referring to it lazily.
+  (a, txOuts) <- runStateT (runReaderT builder txId) []
```
bjornkihlberg, line 146: I wonder if we should use `\() -> expr` or something along those lines when relying on laziness. Otherwise laziness works like a side effect but it's even sneakier than they are in impure languages.

jhbertra: I don't think that adds many benefits - Haskell is lazy by nature; it's not a quirk of the language. Taking advantage of it is, therefore not a hack.

This particular paradigm (using `MonadFix` and `RecursiveDo`) is pretty common. Sometimes you just need to take special care not to force it to WHNF too early (i.e. by pattern matching on it).

jhbertra: In answer to your question, consider the following code:

```hs
mdo
  a <- foo case b of
    Nothing -> []
    Just b' -> [b]
  b <- mkB a
  pure a
```

With `RecursiveDo`, this will desugar as:

```hs
( mfix \(a, b) ->
    ( foo case b of
      Nothing -> []
      Just b -> [b]
    ) >>= (\a -> mkB a) >>= (\b -> pure (a, b))
) >>= (\(a. b) -> pure a)
```

The problem here is pattern matching on `b'` using a case expression before `b'` has been created. How this reduces depends on the definition of `>>=` and `mfix` for the particular `Monad` instance, but in particular, the base of our stack here is `IO`, which delegates its implementation of `mfix` to `fixIO`. In the documentation for `fixIO` there is a [warning](https://hackage.haskell.org/package/base-4.18.0.0/docs/src/System.IO.html#fixIO)

>The implementation of 'Control.Monad.Fix.mfix' for 'IO'. If the function passed to 'fixIO' inspects its argument, the resulting action will throw 'FixIOException'.

jhbertra: Also, consider if the `TxBodyContent` were to depend on the value of the `TxId` - the TxId is computed by hashing the tx body. So You'd end up in an infinite recursion loop (non-productive fixpoint).

PR #551, opened 2023-03-29
PLT-4221 Add draft release process document

jhbertra:
For the first release: no changelog exists yet, we should just do more thorough smoke testing.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -10,9 +10,10 @@ release (i.e. the most recent tag on `main`) is: a prerelease or a normal releas
 
 ### Process if the current release is a normal release
 
-Increment the version numbers in `marlowe-chain-sync.cabal`, `marlowe-runtime.cabal`,
-`marlowe-runtime-web.cabal`, `marlowe-client.cabal`, and `marlowe-runtime-cli.cabal`
-as prescribed by semantic versioning. Given a version number scheme `major.minor.patch`:
+For the `runtime` component, increment the version numbers in
+`marlowe-chain-sync.cabal`, `marlowe-runtime.cabal`, `marlowe-runtime-web.cabal`,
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -0,0 +1,114 @@
...
+## Step 1: Increment Version Numbers
+
+There are two processes to follow depending on what type of release the current
+release (i.e. the most recent tag on `main`) is: a prerelease or a normal release.
+
+### Process if the current release is a normal release
+
+Increment the version numbers in `marlowe-chain-sync.cabal`, `marlowe-runtime.cabal`,
+`marlowe-runtime-web.cabal`, `marlowe-client.cabal`, and `marlowe-runtime-cli.cabal`
+as prescribed by semantic versioning. Given a version number scheme `major.minor.patch`:
```
bwbush, line 15: We have other `.cabal` files, so this number process should include those, too?

RELEASE.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Step 5: Create a Release Tag
+
+Create and push a release tag *on the same commit as the prerelease tag*.
+A release tag is the same as a prerelease tag with the `-prerelease<n>` suffix
+dropped.
+
+## Step 6: Publish Haskell Libraries to CHaP
+
+TODO
```
bwbush, line 97: This should conform to https://github.com/input-output-hk/cardano-engineering-handbook/blob/5cc0176c81404e21bf55b33a6e2bed4101b00d68/policy/haskell/packaging/versioning.md.

bwbush (approved):

RELEASE.md
jhbertra: No - Those other packages are not part of the runtime. I should clarify that this is the set of packages to increment for a runtime release. This document has some strong evidence that the runtime should be a separate repo I think.

shlevy (approved):

PR #561, opened 2023-04-06
PLT-4173: E2E CLI tests

bjornkihlberg:
## Description

This ticket runs `JobClient` and the **Marlowe Runtime CLI** and compares their results to check that they are the same. The ticket is currently implementing a simple example based suite. Property testing might not work because the mechanisms tested are too slow (There might be an interesting opportunity to do this anyway in order to seek out opportunities for performance optimization).

## Unfinished tasks

- Currently the comparison between these two mechanisms are not comparing everything that it could. This is because there's a record called `TxBodyRaw` that has a field `_vldt` that is not deterministic between tests. I haven't yet gotten around to figuring out how to inspect this record as it seems to be a cleverly generated type.
- I am not comparing results of `advance`.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested
    - [x] Changelog


ladamesny (approved):
Looks good @bjornkihlberg !

PR #567, opened 2023-04-19
Refactor Logging

jhbertra:
Changes include:

- [x] Logging refactor to pass event backends around implicitly in a monad
  - [x]  also automatically creates parent / child relationships for nested events
  - [x] Uses a different approach than the one provided by `eventuo11y` so that we can have a standard monad kind `* -> *` instead of `* -> (* -> *) -> * -> *`
- [x] Removes all usage of `MonadBase` and `MonadBaseControl` in favour of `MonadIO` and `MonadUnliftIO` throughout.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
        - Review required
    - [x] Reviewer requested


bwbush (approved):
The code looks fine, but I'd like to better understand the motivation for this. Is it just to reduce the threading of `EventBackend` through function calls? Effectively, there is still residue of `EventBackend` in that `MonadInjectEvent` and and `InjectSelector` appear now where `EventBackend` used to appear, plus there is the added complexity of `UnliftIO`. I gather that the performance and reporting detail is unchanged.

jhbertra:
> The code looks fine, but I'd like to better understand the motivation for this. Is it just to reduce the threading of `EventBackend` through function calls? Effectively, there is still residue of `EventBackend` in that `MonadInjectEvent` and and `InjectSelector` appear now where `EventBackend` used to appear, plus there is the added complexity of `UnliftIO`. I gather that the performance and reporting detail is unchanged.

Basically, it's the same motivation for replacing `ReaderT` or explicit parameters with a `MonadReader` constraint. However, there is a significant improvement that this approach offers - automatic nesting of events. If an event is create in a `withEvent` block, that event will automatically be made the child of the outer event. This enables the automatic linking of sub-events without having to explicitly pass an event backend around, keeping it out of function signatures. In some cases, we don't get to decide what the signature of a function should be, such as in library code, so by passing it around in the monadic context, we can get around this limitation.

PR #568, opened 2023-04-19
PLT-4588 Library functions for safety checks of contracts

bwbush:
Adds library functions supporting safety checks in Marlowe Runtime. A separate pull request (for PLT-3564) will incorporate these checks into the Runtime web service.

Pre-submit checklist:
- Branch
    - [x] [Test report is updated](https://github.com/input-output-hk/marlowe-cardano/blob/main/marlowe/test/test-report.md) (if relevant)
- PR
    - [x] Self-reviewed the diff
        - Review required
        - Review not required
    - [x] Reviewer requested


marlowe/src/Language/Marlowe/Analysis/Safety/Ledger.hs
```diff
@@ -0,0 +1,318 @@
...
+  }
+
+
+-- | Compute a bound on the minimum UTxO that might be required for a Marlowe contract.
+worstMinimumUtxo
+  :: Integer  -- ^ The `utxoCostPerByte` protocol parameter.
+  -> Contract  -- ^ The contract.
+  -> Continuations  -- ^ The merkleized continuations.
+  -> Integer  -- ^ A worst-case bound on the minimum UTxO lovelace for the contract.
+worstMinimumUtxo utxoCostPerByte contract continuations =
```
jhbertra, line 215: Does this take into account any payments made from the deposit account holder during the contract? For example, if I use a 2 ADA minLovelace deposit but I subsequently pay 2 ADA from that account to a party, the min ADA deposit is no longer there.

marlowe-runtime/tx/Language/Marlowe/Runtime/Transaction/Safety.hs
```diff
@@ -0,0 +1,389 @@
...
+    <$> protocolParamUTxOCostPerByte
+
+
+-- | Check a contract for design errors and ledger violations.
+checkContract
+  :: RoleTokensConfig
+  -> MarloweVersion v
+  -> Contract v
+  -> Continuations v
+  -> [SafetyError]
```
jhbertra, line 128: Could this return the full `SafetyReport`? Then we can make the user-provided min ADA deposit optional and use that directly.

marlowe/src/Language/Marlowe/Analysis/Safety/Ledger.hs
bwbush: No, this is caught by the all-transactions analysis. I suppose that we could analyze the contract to see if it ever might withdraw the ada from the deposit account.

bwbush:
FYI: It looks like the CI is failing due to a CI-unfriendly relative path. I'll investigate options for fixing this.

marlowe-runtime/tx/Language/Marlowe/Runtime/Transaction/Safety.hs
```diff
@@ -0,0 +1,389 @@
...
+    . (\utxoCostPerByte -> worstMinimumUtxo (toInteger utxoCostPerByte) contract $ remapContinuations continuations)
+    <$> protocolParamUTxOCostPerByte
+
+
+-- | Check a contract for design errors and ledger violations.
+checkContract
+  :: RoleTokensConfig
+  -> MarloweVersion v
+  -> Contract v
+  -> Continuations v
```
jhbertra, line 127: Where does this error out if a continuation isn't found?

bwbush: This doesn't walk the continuations, so it won't error out if one is missing. I suppose we could add a check that all continuation hashes present in the contract are also keys in the map.

marlowe-runtime/tx/Language/Marlowe/Runtime/Transaction/Safety.hs
bwbush: That's certainly possible--it sort of depends upon the use case for these function. We may want to consider designing the REST API and then altering this library to conveniently enable that use case.

It is possible that `checkContract` may not be needed because `checkTransactions` will catch everything. However, `checkContract` is far faster than `checkTransactions`, so it might be worth calling the former first, just to catch low-hanging fruit before spending resources on a pointless `checkTransactions`.

marlowe-runtime/tx/Language/Marlowe/Runtime/Transaction/Safety.hs
jhbertra: I think that is crucial - or else we aren't doing a full analysis of the contract. The continuations could contain problems and the user would never know.

bwbush:
> FYI: It looks like the CI is failing due to a CI-unfriendly relative path. I'll investigate options for fixing this.

Fixed in 5658e77346e4d2cdbd01399c56adba32023af82d.

marlowe-runtime/tx/Language/Marlowe/Runtime/Transaction/Safety.hs
bwbush: Fixed in 7139176d82a3167d7c7ecbac9cf421457da675c7.

bwbush: Just to clarify, prior to 7139176d82a3167d7c7ecbac9cf421457da675c7 the checks were done on both the contract and its continuations, but there was not check to see that the map of continuations was complete.

marlowe-runtime/tx/Language/Marlowe/Runtime/Transaction/Safety.hs
bwbush: I think we should wait until we're further along on the REST sides of this because this relates to where we want to perform the protocol-dependent checks, where the min-UTxO is chosen (and how the user might override that), etc.

There are really three phases to the safety checking:
1. Select the protocol-dependent min-UTxO for the contract and the account into which that is placed.
2. Check the protocol-independent aspects of the contract.
3. Run the transactions for all possible execution paths, which is expensive.

PR #574, opened 2023-05-05
Plt 5846 marlowe load

jhbertra:
Adds a new protocol for incremental loading and merkleization of contracts.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
        - Review required
    - [x] Reviewer requested


jhbertra:
My justification for adding the `No Changelog Required` label is that this doesn't add any user-facing functionality, as it's all dead code for now.

marlowe-runtime/load-api/Language/Marlowe/Protocol/Load/Client.hs
```diff
@@ -0,0 +1,302 @@
...
+  peerComplete ClientStComplete{..} = AwaitTraced (ServerAgency TokComplete) \case
+    MsgComplete hash -> Closed TokDone $ recvMsgComplete hash
+
+-- | Load a contract into the runtime in a space-efficient way. Note: this
+-- relies on the input contract being lazily evaluated and garbage collected as
+-- it is processed. This means that the caller of this function should avoid holding
+-- onto a reference to the input contract (e.g. in a let-binding). Preferably,
+-- the return value of a function that generates the contract should be passed
+-- directly to this function.
+pushContract
```
bwbush, line 224: I think that the protocol should include a form of `pushContract` that merkleizes each sub-contract within the contract and returns `Map DatumHash Contract`. This supports the second of the two use cases that we have for merkleization:
1. The client incrementally builds the merkleized contract through repeated pushes as they are building it.
2. The client creates a fully unmerkleized contract but needs the protocol to deeply merkelize it so it fits on the chain.

The first use case occurs when, from necessity, the contract is too large to ever represent (in memory or on disk) as unmerkleized. The second use case occurs when the contract is small enough to create an manipulate in memory, but it is too big to fit in a datum.

marlowe-runtime/test/Language/Marlowe/Protocol/LoadSpec.hs
```diff
@@ -0,0 +1,189 @@
...
+import Test.Hspec (Spec, describe)
+import Test.Hspec.QuickCheck (prop)
+import Test.QuickCheck hiding (shrinkMap)
+
+spec :: Spec
+spec = do
+  describe "MarloweLoad protocol" do
+    prop "Has a lawful codec" $ checkPropCodec @(Handshake MarloweLoad)
+    codecGoldenTests @MarloweLoad "MarloweLoad"
+    prop "Merkleizes the  the correct contract hash" \contract ->
```
bwbush, line 40:
```suggestion
    prop "Merkleizes to the correct contract hash" \contract ->
```

bwbush (approved):
I think we need an additional, deeper form of `pushContract`, as I noted in a separate comment.

Thanks for including tests!

marlowe-runtime/load-api/Language/Marlowe/Protocol/Load/Client.hs
jhbertra: This is a great idea, as it would be far more performant for contracts that can fit in memory.

jhbertra:
I've been working on and thinking a bit more about the need for `MsgPushContract`, and I'm not so sure it's a good idea for two reasons:

1. As mentioned, this is a potential attack vector as a client could send the server a potentially huge contract which could cause an OOM error on the server (actually our current Create command has this vulnerability).
2. It allows clients to push already merkleized contracts, which we might miss if we forget to check that the contract is not merkleized already. One of the design goals for `marlowe-load` is a guarantee that if a contract is in the store, it can be fully demerkleized, and allowing clients to push merkleized contracts would break that assumption.

Both of these issues can be addressed by adding more input validation, but in the first case, that would be a major nuisance, as we'd have to do it at the codec level, and in the second it would be a minor nuisance. When you consider that the existing API adequately solves the case of pushing a full contract in one message, albeit in a potentially less efficient way, but with a much more succinct and clean API, I don't think adding this is worth the hassle.

jhbertra:
As for returning a `Map DatumHash Contract`, this can be easily implemented as a client of https://input-output.atlassian.net/browse/PLT-5851

bwbush:
> As for returning a Map DatumHash Contract, this can be easily implemented as a client

I like the idea of pushing this functionality to the client, since that reduces the attack surface and simplifies the API.

PR #585, opened 2023-05-18
Plt 5848 get contract

jhbertra:
Changes in this PR:

- [x] Updated and unified query protocol so chain-sync, marlowe-sync, and marlowe-contract all use the same query protocol.
- [x] Added new contract query API to `marlowe-contract`
- [x] Added contract query sub-protocol to `marlowe-proxy`
- [x] Added contract store query - get contract by hash
- [x] Added Runtime CLI query commands

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
        - Review required
        - Review not required
    - [x] Reviewer requested


marlowe-integration/src/Test/Integration/Marlowe/Local.hs
```diff
@@ -316,6 +315,7 @@ withLocalMarloweRuntime' MarloweRuntimeOptions{..} test = withRunInIO \runInIO -
     contractStore <- createContractStore ContractStoreOptions
       { contractStoreDirectory = resolveWorkspacePath workspace "contract-store"
       , contractStoreStagingDirectory = resolveWorkspacePath workspace "contract-staging-area"
+      , lockingSleepBetweenRetries = 100_000
```
bwbush, line 318: Presumably this is microseconds. I wonder if `lockingMicrosecondsBetweenRetries` might be more expressive and self documenting?

marlowe-protocols/src/Network/Protocol/Query/Types.hs
```diff
@@ -1,401 +1,386 @@
...
-
-class IsQuery query => QueryVariations query where
-  tags :: NonEmpty (SomeTag query)
-  queryVariations :: Tag query delimiter err results -> NonEmpty (query delimiter err results)
-  delimiterVariations :: Tag query delimiter err results -> [delimiter]
-  errVariations :: Tag query delimiter err results -> [err]
-  resultsVariations :: Tag query delimiter err results -> NonEmpty results
-
-instance ArbitraryQuery query => ArbitraryMessage (Query query) where
+class Request req => ArbitraryRequest req where
```
bwbush, line 246: Why isn't this located in a separate package for test generators. Are these arbitrary instances needed by the protocol?

bwbush (approved):
Will the format/organization of the file store for contracts always be forward compatible, or do we have to version it? For some users, it might be the only record they have of the continuations, so they'll want it forever usable in new versions of Marlowe Runtime.

jhbertra:
> Will the format/organization of the file store for contracts always be forward compatible, or do we have to version it? For some users, it might be the only record they have of the continuations, so they'll want it forever usable in new versions of Marlowe Runtime.

Good question. As long as the `.contract` files are always forward-compatible, we should be fine. The `.closure` and `.adjacency` files are basically just precomputed data from the `.contract` files. I don't see a reason why the `.contract` files would ever need to change.

PR #591, opened 2023-05-26
Improve semantics arbitrarty intances

jhbertra:
Did a sweep of `Spec.Marlowe.Semantics.Arbitrary` to improve the instances. Changes mostly focused on:

1. Simplifying overly complex generation logic
2. Making proper use of the `size` parameter
3. Avoiding the use of `suchThat` where possible
4. Fixing incorrectly-written `shrink` implementations (can make a huge difference in performance)

I have annotated my changes with their rationale in the PR.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
        - Review required
    - [x] Reviewer requested

### DIFF
--- a/marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
+++ b/marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
@@ -945,6 +945,8 @@ instance SemiArbitrary Contract where
         , (2, If <$> semiArbitrary ctx <*> resize (size `quot` 2) (semiArbitrary ctx) <*> resize (size `quot` 2) (semiArbitrary ctx))
         , ( 3
           , do
+            -- Since the size of a `When` is O(c*n) where `c` is the number of cases and `n` is the size of sub
+            -- contracts, we need to use `c ~ sqrt size` and `n = size / c` to create a contract that is an appropriate size.
             let maxCases = floor $ sqrt @Double $ fromIntegral size
             numCases <- chooseInt (0, maxCases)
             let numSubContracts = succ numCases
--- a/marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
+++ b/marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
@@ -301,7 +301,7 @@ randomPubKeyHashes =
   ]
 
 instance Arbitrary PubKeyHash where
-  arbitrary = elements randomPubKeyHashes
+  arbitrary = arbitraryFibonacci randomPubKeyHashes
 
 
 -- | Some currency symbols.
@@ -325,7 +325,7 @@ randomCurrencySymbols =
  ]
 
 instance Arbitrary CurrencySymbol where
-  arbitrary = elements randomCurrencySymbols
+  arbitrary = arbitraryFibonacci randomCurrencySymbols
 
 
 -- | Some token names.
@@ -350,7 +350,7 @@ randomTokenNames =
   ]
 
 instance Arbitrary TokenName where
-  arbitrary = elements randomTokenNames
+  arbitrary = arbitraryFibonacci randomTokenNames
   shrink = shrinkByteString (\(TokenName x) -> x) randomTokenNames
 
 instance Arbitrary Token where
@@ -390,7 +390,7 @@ randomRoleNames =
 instance Arbitrary Party where
   arbitrary = frequency
     [ (1, Address <$> arbitrary <*> arbitrary)
-    , (4, Role <$> elements randomRoleNames)
+    , (4, Role <$> arbitraryFibonacci randomRoleNames)
     ]
   shrink (Address _ _) = []
   shrink (Role x)      = Role <$> shrinkByteString unTokenName randomRoleNames x
@@ -415,7 +415,7 @@ instance Arbitrary StakingCredential where
 
 
 instance Arbitrary ValidatorHash where
-  arbitrary = elements randomValidatorHashes
+  arbitrary = arbitraryFibonacci randomValidatorHashes
 
 
 -- | Some validator hashes.
@@ -952,11 +952,18 @@ assertContractWeights = (0, 0, 0, 0, 0, 1)
 instance SemiArbitrary Contract where
   semiArbitrary ctx = sized \size ->
     if size <= 0
-       then pure Close
+       then oneof -- Why not simply close? Because there are tests that effectively do this: `arbitrary `suchThat` (/= Close), which
+        [ Pay <$> semiArbitrary ctx <*> semiArbitrary ctx <*> semiArbitrary ctx <*> semiArbitrary ctx <*> pure Close
+        , If <$> semiArbitrary ctx <*> resize (size `quot` 2) (semiArbitrary ctx) <*> pure Close
+        , When [Case (Notify TrueObs) Close] <$> semiArbitrary ctx <*> semiArbitrary ctx
+        , Let <$> semiArbitrary ctx <*> semiArbitrary ctx <*> pure Close
+        , Assert <$> semiArbitrary ctx <*> pure Close
+        , pure Close
+        ]
        else frequency
-        [ (4 * size, Pay <$> semiArbitrary ctx <*> semiArbitrary ctx <*> semiArbitrary ctx <*> semiArbitrary ctx <*> resize (pred size) (semiArbitrary ctx))
-        , (2 * size, If <$> semiArbitrary ctx <*> resize (size `quot` 2) (semiArbitrary ctx) <*> resize (size `quot` 2) (semiArbitrary ctx))
-        , ( 3 * size
+        [ (4, Pay <$> semiArbitrary ctx <*> semiArbitrary ctx <*> semiArbitrary ctx <*> semiArbitrary ctx <*> resize (pred size) (semiArbitrary ctx))
+        , (2, If <$> semiArbitrary ctx <*> resize (size `quot` 2) (semiArbitrary ctx) <*> resize (size `quot` 2) (semiArbitrary ctx))
+        , ( 3
           , do
             -- Since the size of a `When` is O(c*n) where `c` is the number of cases and `n` is the size of sub
             -- contracts, we need to use `c ~ sqrt size` and `n = size / c` to create a contract that is an appropriate size.
@@ -969,8 +976,8 @@ instance SemiArbitrary Contract where
               <*> semiArbitrary ctx
               <*> resize subContractSize (semiArbitrary ctx)
           )
-        , (4 * size, Let <$> semiArbitrary ctx <*> semiArbitrary ctx <*> resize (pred size) (semiArbitrary ctx))
-        , (size, Assert <$> semiArbitrary ctx <*> resize (pred size) (semiArbitrary ctx))
+        , (4, Let <$> semiArbitrary ctx <*> semiArbitrary ctx <*> resize (pred size) (semiArbitrary ctx))
+        , (1, Assert <$> semiArbitrary ctx <*> resize (pred size) (semiArbitrary ctx))
         , (1, pure Close)
         ]
 
--- a/marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
+++ b/marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
@@ -11,13 +11,16 @@
 -----------------------------------------------------------------------------
 
 
+{-# LANGUAGE ApplicativeDo #-}
+{-# LANGUAGE BlockArguments #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE TypeApplications #-}
 
 {-# OPTIONS_GHC -fno-warn-orphans #-}
-{-# LANGUAGE BlockArguments #-}
 
 
 module Spec.Marlowe.Semantics.Arbitrary
@@ -52,7 +55,6 @@ module Spec.Marlowe.Semantics.Arbitrary
   ) where
 
 
-import Control.Monad (replicateM)
 import Data.Function (on)
 import Data.List (nub, nubBy)
 import Language.Marlowe.Core.V1.Semantics
@@ -104,9 +106,23 @@ import Plutus.V2.Ledger.Api
 import PlutusTx.Builtins (BuiltinByteString, appendByteString, lengthOfByteString, sliceByteString)
 import Spec.Marlowe.Semantics.Golden (GoldenTransaction, goldenContracts, goldenTransactions)
 import Spec.Marlowe.Semantics.Merkle (merkleizeInputs, shallowMerkleize)
-import Test.Tasty.QuickCheck
-  (Arbitrary(..), Gen, chooseInteger, elements, frequency, listOf, oneof, shrinkList, sized, suchThat, vectorOf)
+import Test.QuickCheck
+  ( Arbitrary(..)
+  , Gen
+  , chooseInt
+  , chooseInteger
+  , elements
+  , frequency
+  , listOf
+  , oneof
+  , resize
+  , shrinkList
+  , sized
+  , suchThat
+  , vectorOf
+  )
 
+import Data.Functor ((<&>))
 import qualified Plutus.V2.Ledger.Api as Ledger (Address(..))
 import qualified PlutusTx.AssocMap as AM (Map, delete, empty, fromList, keys, toList)
 import qualified PlutusTx.Eq as P (Eq)
@@ -117,12 +133,12 @@ fibonaccis :: Num a => [a]
 fibonaccis = [2, 3, 5, 8, 13, 21, 34, 55, 89, 144, 233, 377, 610, 987, 1597, 2584]
 
 
--- | Inverse-Fibanoncci frequencies.
+-- | Inverse-Fibonacci frequencies.
 fibonacciFrequencies :: Integral a => [a]
 fibonacciFrequencies = (1000000 `div`) <$> fibonaccis
 
 
--- | Select an element of a list with propability proportional to inverse-Fibonacci weights.
+-- | Select an element of a list with probability proportional to inverse-Fibonacci weights.
 arbitraryFibonacci :: [a] -> Gen a
 arbitraryFibonacci = frequency . zip fibonacciFrequencies . fmap pure
 
@@ -243,34 +259,25 @@ instance IsValid Context where
 
 -- | An arbitrary positive integer, mostly small.
 arbitraryPositiveInteger :: Gen Integer
-arbitraryPositiveInteger =
-  frequency
-    [
-      (60, arbitrary `suchThat` (> 0))
-    , (30, arbitraryFibonacci fibonaccis)
-    ]
+arbitraryPositiveInteger = arbitraryNonnegativeInteger <&> \case
+  0 -> 1
+  n -> n
 
 
 -- | An arbitrary non-negative integer, mostly small.
 arbitraryNonnegativeInteger :: Gen Integer
-arbitraryNonnegativeInteger =
-  frequency
-    [
-      (60, arbitrary `suchThat` (>= 0))
-    , (30, arbitraryFibonacci fibonaccis)
-    ]
+arbitraryNonnegativeInteger = frequency
+  [ (1, abs <$> arbitrary)
+  , (6, floor . sqrt @Double . abs <$> arbitrary)
+  ]
 
 
 -- | An arbitrary integer, mostly small.
 arbitraryInteger :: Gen Integer
-arbitraryInteger =
-  frequency
-    [
-      ( 5, arbitrary `suchThat` (< 0))
-    , (30, arbitrary `suchThat` (> 0))
-    , ( 5, pure 0)
-    , (60, arbitraryFibonacci fibonaccis)
-    ]
+arbitraryInteger = frequency
+  [ (1, negate <$> arbitraryPositiveInteger)
+  , (6, arbitraryNonnegativeInteger)
+  ]
 
 
 -- | Some public key hashes.
@@ -291,13 +298,10 @@ randomPubKeyHashes =
   , "a2bd7dd7f41c2781d1d11c7f4994fac750525705f9c259f97cb27d0e"
   , "c5b4c543a0d0d181ec387ad8250b18617bb18bcf2eccc0f27fe7aa23"
   , "d877b83ece77d785fee4a52bd7226949fa64e111aa0e20cd4a1c471b"
-  , "e14025a93f867851b9bb3c48601d1845bcbe9e2e1856c16cfc0522"      -- NB: Too short for ledger.
-  , "e3351d289f3eaa66e500f17b91a74e492193f4485c32e5ad606da83542"  -- NB: Too long for ledger.
   ]
 
 instance Arbitrary PubKeyHash where
-  arbitrary = arbitraryFibonacci randomPubKeyHashes `suchThat` (\(PubKeyHash h) -> lengthOfByteString h == 28)
-  shrink x = filter (< x) randomPubKeyHashes
+  arbitrary = elements randomPubKeyHashes
 
 
 -- | Some currency symbols.
@@ -318,13 +322,10 @@ randomCurrencySymbols =
   , "9f92753881b398a247e53b6cad08eab0e158cf1ef5df84c7f5766041"
   , "c1f46ec0147542f9bc155805993497ed44150687a41d0a63af3be466"
   , "cc2189d7adde0ed26355fd03e134feb508e5924959b07a53557f285e"
-  , "df97bf95b2d21327731329d94173344ff4db5ac16f92250d9cab00"      -- NB: Too short for ledger.
-  , "ead659651c55f5481dbc7038a7c096fd7616d2f86471bd9d46de742ea0"  -- NB: Too long for ledger.
  ]
 
 instance Arbitrary CurrencySymbol where
-  arbitrary = arbitraryFibonacci randomCurrencySymbols
-  shrink x = filter (< x) randomCurrencySymbols
+  arbitrary = elements randomCurrencySymbols
 
 
 -- | Some token names.
@@ -350,19 +351,17 @@ randomTokenNames =
   ]
 
 instance Arbitrary TokenName where
-  arbitrary = arbitraryFibonacci randomTokenNames
+  arbitrary = sized \size -> elements $ takeWhile ((<= size) . fromInteger . lengthOfByteString . unTokenName) randomTokenNames
   shrink = shrinkByteString (\(TokenName x) -> x) randomTokenNames
 
 instance Arbitrary Token where
-  arbitrary =
-     do
-       isAda <- arbitrary
-       if isAda
-         then pure $ Token adaSymbol adaToken
-         else Token <$> arbitrary <*> arbitrary
+  arbitrary = oneof
+    [ pure $ Token adaSymbol adaToken
+    , Token <$> arbitrary <*> arbitrary
+    ]
   shrink (Token c n)
     | c == adaSymbol && n == adaToken = []
-    | otherwise                       = Token adaSymbol adaToken : [Token c' n' | c' <- shrink c, n' <- shrink n]
+    | otherwise                       = Token adaSymbol adaToken : (uncurry Token <$> shrink (c, n))
 
 instance SemiArbitrary Token where
   fromContext = tokens
@@ -387,18 +386,15 @@ randomRoleNames =
   , "Nonso Ernie Blanka"
   , "Umukoro Alexander Columb"
   , "Urbanus Roland Alison Ty Ryoichi"
-  , "Alcippe Alende Blanka Roland Dafne"  -- NB: Too long for ledger.
   ]
 
 instance Arbitrary Party where
-  arbitrary =
-    do
-       isPubKeyHash <- frequency [(2, pure True), (8, pure False)]
-       if isPubKeyHash
-         then Address <$> arbitrary <*> arbitrary
-         else Role <$> arbitraryFibonacci randomRoleNames
-  shrink (Address _ _) = Role <$> randomRoleNames
-  shrink (Role x)      = Role <$> shrinkByteString (\(TokenName y) -> y) randomRoleNames x
+  arbitrary = frequency
+    [ (1, Address <$> arbitrary <*> arbitrary)
+    , (4, sized \size -> Role <$> elements (takeWhile ((<= size) . fromInteger . lengthOfByteString . unTokenName) randomRoleNames))
+    ]
+  shrink (Address _ _) = []
+  shrink (Role x)      = Role <$> shrinkByteString unTokenName randomRoleNames x
 
 instance SemiArbitrary Party where
   fromContext = parties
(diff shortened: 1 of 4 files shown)


marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -243,34 +259,25 @@ instance IsValid Context where
...
 arbitraryPositiveInteger :: Gen Integer
-arbitraryPositiveInteger =
-  frequency
-    [
-      (60, arbitrary `suchThat` (> 0))
-    , (30, arbitraryFibonacci fibonaccis)
-    ]
+arbitraryPositiveInteger = arbitraryNonnegativeInteger <&> \case
+  0 -> 1
+  n -> n
```
jhbertra, line 264: Rationale: avoid `suchThat` in favour of generating only valid values.

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -243,34 +259,25 @@ instance IsValid Context where
...
-arbitraryNonnegativeInteger =
-  frequency
-    [
-      (60, arbitrary `suchThat` (>= 0))
-    , (30, arbitraryFibonacci fibonaccis)
-    ]
+arbitraryNonnegativeInteger = frequency
+  [ (1, abs <$> arbitrary)
+  , (6, floor . sqrt @Double . abs <$> arbitrary)
+  ]
```
jhbertra, line 272: Rationale:
- Generate only valid values instead of using `suchThat` to improve performance.
- To satisfy "mostly small", generate values on an inverse-quadratic curve instead of picking only Fibbonacci values (results in better coverage).

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -243,34 +259,25 @@ instance IsValid Context where
...
-    [
-      ( 5, arbitrary `suchThat` (< 0))
-    , (30, arbitrary `suchThat` (> 0))
-    , ( 5, pure 0)
-    , (60, arbitraryFibonacci fibonaccis)
-    ]
+arbitraryInteger = frequency
+  [ (1, negate <$> arbitraryPositiveInteger)
+  , (6, arbitraryNonnegativeInteger)
+  ]
```
jhbertra, line 280: Rationale:
- Generate only valid values instead of using suchThat to improve performance.
- Simplify unnecessarily complex generation logic

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -291,13 +298,10 @@ randomPubKeyHashes =
   , "a2bd7dd7f41c2781d1d11c7f4994fac750525705f9c259f97cb27d0e"
   , "c5b4c543a0d0d181ec387ad8250b18617bb18bcf2eccc0f27fe7aa23"
   , "d877b83ece77d785fee4a52bd7226949fa64e111aa0e20cd4a1c471b"
-  , "e14025a93f867851b9bb3c48601d1845bcbe9e2e1856c16cfc0522"      -- NB: Too short for ledger.
-  , "e3351d289f3eaa66e500f17b91a74e492193f4485c32e5ad606da83542"  -- NB: Too long for ledger.
```
jhbertra, line 295: Removed because we just filtered these out anyway.

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -291,13 +298,10 @@ randomPubKeyHashes =
...
   ]
 
 instance Arbitrary PubKeyHash where
-  arbitrary = arbitraryFibonacci randomPubKeyHashes `suchThat` (\(PubKeyHash h) -> lengthOfByteString h == 28)
-  shrink x = filter (< x) randomPubKeyHashes
+  arbitrary = elements randomPubKeyHashes
```
jhbertra, line 304: - simplified
- avoid use of `suchThat`

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -318,13 +322,10 @@ randomCurrencySymbols =
   , "9f92753881b398a247e53b6cad08eab0e158cf1ef5df84c7f5766041"
   , "c1f46ec0147542f9bc155805993497ed44150687a41d0a63af3be466"
   , "cc2189d7adde0ed26355fd03e134feb508e5924959b07a53557f285e"
-  , "df97bf95b2d21327731329d94173344ff4db5ac16f92250d9cab00"      -- NB: Too short for ledger.
-  , "ead659651c55f5481dbc7038a7c096fd7616d2f86471bd9d46de742ea0"  -- NB: Too long for ledger.
```
jhbertra, line 322: Removed - `Arbitrary` should not generate invalid values.

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -350,19 +351,17 @@ randomTokenNames =
   ]
 
 instance Arbitrary TokenName where
-  arbitrary = arbitraryFibonacci randomTokenNames
+  arbitrary = sized \size -> elements $ takeWhile ((<= size) . fromInteger . lengthOfByteString . unTokenName) randomTokenNames
```
jhbertra, line 354: Use size parameter to control length of chosen value rather than Fibonacci frequencies

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -350,19 +351,17 @@ randomTokenNames =
...
-  arbitrary =
-     do
-       isAda <- arbitrary
-       if isAda
-         then pure $ Token adaSymbol adaToken
-         else Token <$> arbitrary <*> arbitrary
+  arbitrary = oneof
+    [ pure $ Token adaSymbol adaToken
+    , Token <$> arbitrary <*> arbitrary
+    ]
```
jhbertra, line 361: Use oneof instead of generating a boolean.

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -387,18 +386,15 @@ randomRoleNames =
   , "Nonso Ernie Blanka"
   , "Umukoro Alexander Columb"
   , "Urbanus Roland Alison Ty Ryoichi"
-  , "Alcippe Alende Blanka Roland Dafne"  -- NB: Too long for ledger.
```
jhbertra, line 390: Removed: `Arbitrary` instances should not produce invalid values.

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -387,18 +386,15 @@ randomRoleNames =
...
-         then Address <$> arbitrary <*> arbitrary
-         else Role <$> arbitraryFibonacci randomRoleNames
-  shrink (Address _ _) = Role <$> randomRoleNames
-  shrink (Role x)      = Role <$> shrinkByteString (\(TokenName y) -> y) randomRoleNames x
+  arbitrary = frequency
+    [ (1, Address <$> arbitrary <*> arbitrary)
+    , (4, sized \size -> Role <$> elements (takeWhile ((<= size) . fromInteger . lengthOfByteString . unTokenName) randomRoleNames))
+    ]
+  shrink (Address _ _) = []
+  shrink (Role x)      = Role <$> shrinkByteString unTokenName randomRoleNames x
```
jhbertra, line 397: Fixed improper `shrink` implementation: all values of a `shrink` should be derived from the value being shrunk. Shrinking an address to a random role does not satisfy this.

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -420,14 +416,13 @@ instance Arbitrary StakingCredential where
 
 
 instance Arbitrary ValidatorHash where
-  arbitrary = arbitraryFibonacci randomValidatorHashes `suchThat` (\(ValidatorHash h) -> lengthOfByteString h == 28)
+  arbitrary = elements randomValidatorHashes
```
jhbertra, line 419: Simplified

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -441,14 +436,11 @@ randomValidatorHashes =
   , "a2bd7ddd1d11c7f4994fa7f41c2781c750525705f9c259f97cb27d0e"
   , "c5b4c54ec387ad8250b183a0d0d181617bb18bcf2eccc0f27fe7aa23"
   , "d877b83fee4a52bd72269ece77d78549fa64e111aa0e20cd4a1c471b"
-  , "e14025ab9bb3c48601d1893f86785145bcbe9e2e1856c16cfc0522"      -- NB: Too short for ledger.
-  , "e3351d2e500f17b91a74e89f3eaa66492193f4485c32e5ad606da83542"  -- NB: Too long for ledger.
```
jhbertra, line 445: Removed - we were filtering these out anyway, and arbitrary instances should not generate invalid values.

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -441,14 +436,11 @@ randomValidatorHashes =
...
   ]
 
 
 instance Arbitrary POSIXTime where
   arbitrary = POSIXTime <$> arbitraryInteger
-  shrink x = filter (< x) fibonaccis
```
jhbertra, line 451: It is unnecessary to shrink integer values

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -570,36 +559,34 @@ invalidValues lower upper = [x | x <- availableValues, x < lower  || x > upper]
...
--- | Generate a semi-random time interrval straddling a given time.
+-- | Generate a semi-random time interval straddling a given time.
 arbitraryTimeIntervalAround :: POSIXTime -> Gen TimeInterval
 arbitraryTimeIntervalAround (POSIXTime limit) =
   do
-    start <- arbitraryInteger `suchThat` (<= limit)
-    duration <- ((limit - start) +) <$> arbitraryNonnegativeInteger
-    pure (POSIXTime start, POSIXTime $ start + duration)
+    (dStart, dEnd) <- (,) <$> arbitraryNonnegativeInteger <*> arbitraryNonnegativeInteger
+    pure (POSIXTime $ limit - dStart, POSIXTime $ limit + dEnd)
```
jhbertra, line 575: Performance: generate valid values by construction instead of using `suchThat`

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -570,36 +559,34 @@ invalidValues lower upper = [x | x <- availableValues, x < lower  || x > upper]
...
 
 -- | Generate a semi-random time interval before a given time.
 arbitraryTimeIntervalBefore :: POSIXTime -> POSIXTime -> Gen TimeInterval
 arbitraryTimeIntervalBefore (POSIXTime lower) (POSIXTime upper) =
   do
-    start <- arbitraryInteger `suchThat` (<= lower)
-    duration <- chooseInteger (0, upper - start - 1)
-    pure (POSIXTime start, POSIXTime $ start + duration)
+    (dStart, dEnd) <- (,) <$> arbitraryNonnegativeInteger <*> arbitraryNonnegativeInteger
+    pure (POSIXTime $ lower - dStart, POSIXTime $ upper - dEnd)
```
jhbertra, line 582: Performance: generate valid values by construction instead of using suchThat

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -570,36 +559,34 @@ invalidValues lower upper = [x | x <- availableValues, x < lower  || x > upper]
...
 
 -- | Generate a semi-random time interval after a given time.
 arbitraryTimeIntervalAfter :: POSIXTime -> Gen TimeInterval
 arbitraryTimeIntervalAfter (POSIXTime lower) =
   do
-    start <- arbitraryInteger `suchThat` (>= lower)
-    duration <- arbitraryNonnegativeInteger
+    (dStart, duration)  <- (,) <$> arbitraryNonnegativeInteger <*> arbitraryNonnegativeInteger
+    let start = lower + dStart
```
jhbertra, line 589: Performance: generate valid values by construction instead of using suchThat

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -618,18 +605,10 @@ shrinkTimeInterval (start, end) =
...
 
-instance SemiArbitrary TimeInterval where
-  semiArbitrary context =
-    do
-      POSIXTime start <- semiArbitrary context
-      duration <- arbitraryNonnegativeInteger
-      pure (POSIXTime start, POSIXTime $ start + duration)
-  semiShrink = fmap shrinkTimeInterval
-
-
```
jhbertra, line 629: Moved

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -618,18 +605,10 @@ shrinkTimeInterval (start, end) =
...
 instance Arbitrary ChoiceId where
   arbitrary = ChoiceId <$> arbitraryChoiceName <*> arbitrary
-  shrink (ChoiceId n p) = [ChoiceId n' p' | n' <- shrinkChoiceName n, p' <- shrink p]
+  shrink (ChoiceId n p) = (ChoiceId <$> shrinkChoiceName n <*> pure p)
+                       ++ (ChoiceId n <$> shrink p)
```
jhbertra, line 611: Improper `shrink` implementation - shrink parts of products independently rather than taking cartesian product of shrinks.

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -695,24 +674,25 @@ instance SemiArbitrary Integer where
...
+        , (4, UseValue <$> arbitrary)
+        ]
+      else oneof -- size > 0 produces compound values
+        [ resize (pred size) $ NegValue <$> arbitrary
+        , resize (size `quot` 2) $ AddValue <$> arbitrary <*> arbitrary
+        , resize (size `quot` 2) $ SubValue <$> arbitrary <*> arbitrary
+        , resize (size `quot` 2) $ MulValue <$> arbitrary <*> arbitrary
+        , resize (size `quot` 2) $ DivValue <$> arbitrary <*> arbitrary
+        , resize (size `quot` 3) $ Cond <$> arbitrary <*> arbitrary <*> arbitrary
+        ]
```
jhbertra, line 694: Use `size` parameter.

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -695,24 +674,25 @@ instance SemiArbitrary Integer where
...
   shrink (AvailableMoney a t) = [AvailableMoney a' t | a' <- shrink a] ++ [AvailableMoney a t' | t' <- shrink t]
-  shrink (Constant x)         = Constant <$> shrink x
```
jhbertra, line 715: Shrinking an integer is unnecessary.

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -721,43 +701,49 @@ instance Arbitrary (Value Observation) where
   shrink (ChoiceValue c)      = ChoiceValue <$> shrink c
   shrink (UseValue v)         = UseValue <$> shrink v
   shrink (Cond o x y)         = [Cond o' x y | o' <- shrink o] ++ [Cond o x' y | x' <- shrink x] ++ [Cond o x y' | y' <- shrink y]
-  shrink _                    = []
+  shrink (Constant _)         = []
+  shrink TimeIntervalStart    = []
+  shrink TimeIntervalEnd      = []
```
jhbertra, line 706: Expand wildcard case to explicit cases (so new cases must be handled when added).

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -721,43 +701,49 @@ instance Arbitrary (Value Observation) where
...
+        , (4, UseValue <$> semiArbitrary context)
+        ]
+      else oneof -- size > 0 produces compound values
+        [ resize (pred size) $ NegValue <$> semiArbitrary context
+        , resize (size `quot` 2) $ AddValue <$> semiArbitrary context <*> semiArbitrary context
+        , resize (size `quot` 2) $ SubValue <$> semiArbitrary context <*> semiArbitrary context
+        , resize (size `quot` 2) $ MulValue <$> semiArbitrary context <*> semiArbitrary context
+        , resize (size `quot` 2) $ DivValue <$> semiArbitrary context <*> semiArbitrary context
+        , resize (size `quot` 3) $ Cond <$> semiArbitrary context <*> semiArbitrary context <*> semiArbitrary context
+        ]
```
jhbertra, line 726: Use `size` parameter

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -767,25 +753,27 @@ instance Arbitrary Observation where
   shrink (ValueLT x y)      = [ValueLT x' y | x' <- shrink x] ++ [ValueLT x y' | y' <- shrink y]
   shrink (ValueLE x y)      = [ValueLE x' y | x' <- shrink x] ++ [ValueLE x y' | y' <- shrink y]
   shrink (ValueEQ x y)      = [ValueEQ x' y | x' <- shrink x] ++ [ValueEQ x y' | y' <- shrink y]
-  shrink _                  = []
+  shrink TrueObs            = []
+  shrink FalseObs           = []
```
jhbertra, line 757: Avoid using `_` pattern (hides unhandled cases)

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -767,25 +753,27 @@ instance Arbitrary Observation where
...
+      else oneof -- size > 0 produces compound observations.
+        [ resize (size `quot` 2) $ AndObs <$> semiArbitrary context <*> semiArbitrary context
+        , resize (size `quot` 2) $ OrObs <$> semiArbitrary context <*> semiArbitrary context
+        , resize (pred size) $ NotObs <$> semiArbitrary context
+        , resize (size `quot` 2) $ ValueGE <$> semiArbitrary context <*> semiArbitrary context
+        , resize (size `quot` 2) $ ValueGT <$> semiArbitrary context <*> semiArbitrary context
+        , resize (size `quot` 2) $ ValueLT <$> semiArbitrary context <*> semiArbitrary context
+        , resize (size `quot` 2) $ ValueLE <$> semiArbitrary context <*> semiArbitrary context
+        , resize (size `quot` 2) $ ValueEQ <$> semiArbitrary context <*> semiArbitrary context
+        ]
```
jhbertra, line 776: Use `size` parameter

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -815,16 +803,16 @@ instance SemiArbitrary Bound where
         pure $ Bound lower (lower + extent)
 
 
-instance SemiArbitrary [Bound] where
-  semiArbitrary context = listOf $ semiArbitrary context
+instance SemiArbitrary a => SemiArbitrary [a] where
+  semiArbitrary = listOf . semiArbitrary
```
jhbertra, line 807: Generalize instance

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -815,16 +803,16 @@ instance SemiArbitrary Bound where
...
 
 
 instance Arbitrary Action where
   arbitrary =
     frequency
       [
         (3, Deposit <$> arbitrary <*> arbitrary <*> arbitrary <*> arbitrary)
-      , (6, Choice <$> arbitrary <*> arbitrary `suchThat` ((< 5) . length))
+      , (6, Choice <$> arbitrary <*> (take 4 <$> arbitrary))
       , (1, Notify <$> arbitrary)
```
jhbertra, line 816: Avoid `suchThat`

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -899,11 +881,11 @@ instance Arbitrary Contract where
   shrink (When a t c) = [c] ++ [When a' t c | a' <- shrink a] ++ [When a t' c | t' <- shrink t] ++ [When a t c' | c' <- shrink c]
   shrink (Let v x c) = [c] ++ [Let v' x c | v' <- shrink v] ++ [Let v x' c | x' <- shrink x] ++ [Let v x c' | c' <- shrink c]
   shrink (Assert o c) = [c] ++ [Assert o' c | o' <- shrink o] ++ [Assert o c' | c' <- shrink c]
-  shrink _ = []
+  shrink Close = []
```
jhbertra, line 884: Don't use `_` case

bwbush (approved):
Thanks for including the rationales for the changes!

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -912,7 +894,7 @@ arbitraryContractWeighted ((wClose, wPay, wIf, wWhen, wLet, wAssert) : w) contex
       (wClose , pure Close)
     , (wPay   , Pay <$> semiArbitrary context <*> semiArbitrary context <*> semiArbitrary context <*> semiArbitrary context <*> arbitraryContractWeighted w context)
     , (wIf    , If <$> semiArbitrary context <*> arbitraryContractWeighted w context <*> arbitraryContractWeighted w context)
-    , (wWhen  , When <$> listOf (arbitraryCaseWeighted w context) `suchThat` ((<= length w) . length) <*> semiArbitrary context <*> arbitraryContractWeighted w context)
+    , (wWhen  , When <$> (take (length w) <$> listOf (arbitraryCaseWeighted w context)) <*> semiArbitrary context <*> arbitraryContractWeighted w context)
```
jhbertra, line 897: Don't use `suchThat` when it can be avoided. This is another example of the same problem and is almost certainly the reason why contract generation performance was affected by the size parameter to the degree it was.

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -954,25 +936,33 @@ assertContractWeights :: (Int, Int, Int, Int, Int, Int)
...
+            let numSubContracts = succ numCases
+            let subContractSize = size `quot` numSubContracts
+            When
+              <$> (vectorOf numCases $ resize subContractSize $ semiArbitrary ctx)
+              <*> semiArbitrary ctx
+              <*> resize subContractSize (semiArbitrary ctx)
+          )
+        , (4, Let <$> semiArbitrary ctx <*> semiArbitrary ctx <*> resize (pred size) (semiArbitrary ctx))
+        , (1, Assert <$> semiArbitrary ctx <*> resize (pred size) (semiArbitrary ctx))
+        ]
```
jhbertra, line 959: Make proper use of size parameter.

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -954,25 +936,35 @@ assertContractWeights :: (Int, Int, Int, Int, Int, Int)
...
```
jhbertra, line 961: Make use of `size` parameter

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -954,25 +936,35 @@ assertContractWeights :: (Int, Int, Int, Int, Int, Int)
...
 -- | Generate a random association map.
 arbitraryAssocMap :: Eq k => Gen k -> Gen v -> Gen (AM.Map k v)
-arbitraryAssocMap arbitraryKey arbitraryValue =
-  do
-    entries <- arbitraryFibonacci [0..]
-    fmap (AM.fromList . nubBy ((==) `on` fst))
-      . replicateM entries
-      $ (,) <$> arbitraryKey <*> arbitraryValue
+arbitraryAssocMap arbitraryKey arbitraryValue = AM.fromList . nubBy ((==) `on` fst)
+  <$> listOf ((,) <$> arbitraryKey <*> arbitraryValue)
```
jhbertra, line 967: Use `size` to determine the length of the map instead of Fibonacci numbers.

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -990,34 +982,34 @@ instance Arbitrary Accounts where
   shrink = shrinkAssocMap
 
 instance SemiArbitrary Accounts where
+  semiArbitrary context = arbitraryAssocMap (semiArbitrary context) (semiArbitrary context `suchThat` (> 0))
```
jhbertra, line 985: Use size to determine the length of the map instead of Fibonacci numbers.

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -1086,11 +1078,7 @@ instance Arbitrary TransactionInput where
     <> [TransactionInput txInterval  txInputs' | txInputs'   <- shrink txInputs  ]
 
 instance SemiArbitrary TransactionInput where
-  semiArbitrary context =
-    do
-      Environment txInterval <- semiArbitrary context
-      n <- arbitraryFibonacci [1, 1, 1, 1, 0, 2]  -- TODO: Review.
```
jhbertra, line 1092: Use `size` to control length of inputs instead of Fibonacci numbers.

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
palas: That also removes the giving fibonaccis more frequency, not sure whether that is good or bad

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
palas: This removes shrinking, I think it was useful to have it as it was, because, then you get counterexamples with fewer different addresses (if the actual address doesn't matter), easier to read

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -318,13 +322,10 @@ randomCurrencySymbols =
...
   , "c1f46ec0147542f9bc155805993497ed44150687a41d0a63af3be466"
   , "cc2189d7adde0ed26355fd03e134feb508e5924959b07a53557f285e"
-  , "df97bf95b2d21327731329d94173344ff4db5ac16f92250d9cab00"      -- NB: Too short for ledger.
-  , "ead659651c55f5481dbc7038a7c096fd7616d2f86471bd9d46de742ea0"  -- NB: Too long for ledger.
  ]
 
 instance Arbitrary CurrencySymbol where
-  arbitrary = arbitraryFibonacci randomCurrencySymbols
-  shrink x = filter (< x) randomCurrencySymbols
+  arbitrary = elements randomCurrencySymbols
```
palas, line 328: Same here

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
palas: We do need negative tests though. Do we still have those? How do we go about them?

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
palas: Why?

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
```diff
@@ -473,7 +465,6 @@ randomChoiceNames =
   , "attend unknown animals"
   , "position increated radiation"
   , "proclaim endless sordid figments"
-  , "understand weigh originate envisage"  -- NB: Too long for ledger.
```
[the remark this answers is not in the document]
palas: That makes sense to me

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
palas: I think it would be good to generate leaves sometimes even if size isn't 0, cause size isn't random

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
palas: It can point out where the error is, if the error is because of the number, then it may be the one that doesn't shrink

palas (approved):
I love having the comments on the changes, really helpful. I do have some suggestions, not all of them are crucial, but I think some will help the quality of the generated tests

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
jhbertra: There isn't anything special about the Fibonaccis, so I don't think it matters

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
jhbertra: That may be, but it assumes the context of the test (which arbitrary instances shouldn't do). Shrinking should produce structurally smaller values that can be derived from the value being shrunk, and picking a different element from the "universe" of hashes A) isn't structurally smaller and B) isn't creating a "smaller version" of the value being shrunk - it's just returning a new, unrelated value.

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
jhbertra: I don't think we explicitly had them in the first place - because the arbitrary instances sometimes gave invalid data. That is sometimes what you want, sometimes not. For testing handling of invalid data, we should explicitly test those cases and use custom `Gen`s that produce invalid data on purpose.

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
jhbertra: Because one integer value is not structurally smaller than another (though I see QuickCheck does define rather elaborate shrinking logic for integral values, I can restore this.

marlowe-test/src/Spec/Marlowe/Semantics/Arbitrary.hs
jhbertra: I will restore the shrinking of integers

PR #597, opened 2023-06-02
Log uncaught exceptions

jhbertra:
This logs all uncaught exceptions in async components, even in the server workers which are subsequently ignored.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
        - Review required
    - [x] Reviewer requested


bwbush (approved):
Excellent! Thanks. :smile:

bwbush:
Would it be possible to have the *full* stack trace printed when the thread dies? Reports like the following could be more informative:
```console
Component chain-seek-worker crashed with exception:
Data.Binary.Get.runGet at position 0: not enough bytes
CallStack (from HasCallStack):
  error, called at src/Data/Binary/Get.hs:351:5 in binary-0.8.8.0-HpmhsQDJ8BC4iODE4wIoDZ:Data.Binary.Get
```

jhbertra:
> Would it be possible to have the _full_ stack trace printed when the thread dies? Reports like the following could be more informative:
> 
> ```
> Component chain-seek-worker crashed with exception:
> Data.Binary.Get.runGet at position 0: not enough bytes
> CallStack (from HasCallStack):
>   error, called at src/Data/Binary/Get.hs:351:5 in binary-0.8.8.0-HpmhsQDJ8BC4iODE4wIoDZ:Data.Binary.Get
> ```

From what I can gather, there's no easy way to make this happen. I think this requires more investigation.

PR #602, opened 2023-06-06
PLT-3146 - GET /contracts/:contractId/next

nhenin:
- **JIRA** : [PLT-3146](https://input-output.atlassian.net/browse/PLT-6075?atlOrigin=eyJpIjoiODY5Zjg3ZDExNzNjNGE0NDk2OWZhNzQ3ZDFlNTgwNzgiLCJwIjoiaiJ9)
- **Value** : Giving the capabilities for Rest API users to know what next inputs can be applied for a given contract 
- **Next Semantics**
   - **Reducibility**
     - Can Reduce when contract is reducible
     - Can't Reduce when the contract provided is irreducible
   - **Applicability**
     - "Close" is not applicable
     - **Notify**
        - "Notify" is not applicable when evaluated falsified        
        - Shadowing : Following Notifies evaluated to True are not applicable.
     - **Deposit**
        - "CanDeposit" is a "Deposit" with its quantity evaluated and its "Case" index preserved (when no shadowing involved
        - Shadowing : Following Identical Evaluated Deposits are not applicable     
     - **Choice**
        - "Choice" with empty bounds is not applicables
        - **Shadowing** : Bounds are not applicable when previously covered by Choices with the same Id
          - Applicable [Indexed CanChoose]'s bounds on the same choiceId don't overlap                                    
          - "CanChoose" is isomorphic to "Choice" and its "Case" index preserved when no shadowing involved               
          - "[Indexed CanChoose]" and [Choice] on the same id have the same merged Bounds                             

- **Runtime REST** 

Adding new endpoint `GET contracts/:contractId/next?validityStart=:validityStart&validityEnd=:validityEnd`

```json
{
  "applicable_inputs": {
    "choices": [
      {
        "can_choose_between": [
          {
            "from": 0,
            "to": 0
          }
        ],
        "case_index": 0,
        "for_choice": {
          "choice_name": "string"
        }
      }
    ],
    "deposits": [
      {
        "can_deposit": 0,
        "case_index": 1,
        "of_token": {
          "currency_symbol": "string",
          "token_name": "string"
        }
      }
    ],
    "notify": {
      "case_index": 2,
      "is_merkleized_continuation": true
    }
  },
  "can_reduced": true
}
```

**Done** 
- [x] next semantics
- [x] next endpoint

**Todo** 
- [x] Review Next Semantic
- [x] Review Next Endpoint and Resource Encoding
- [x] Testing Semantics
- [x] Testing Endpoint

[PLT-3146]: https://input-output.atlassian.net/browse/PLT-3146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hie.yaml
```diff
@@ -1,2 +1,376 @@
 cradle:
   cabal:
+    - path: "async-components/src"
```
jhbertra, line 3: Can you please revert these changes?

marlowe/src/Language/Marlowe/Core/V1/Semantics/Next.hs
```diff
@@ -0,0 +1,175 @@
...
+next  _ _ Close         = Right . CanAdvance $ []
+next  _ _ (When [] _ _) = Right CanNotAdvance   -- Contract Suspended in a time out
+next environment state (When (x:xs) _ _)
+  = Right
+    . maybe
+          CanNotAdvance -- Contract Suspended (only falsified Notifies evaluated in the When)
+        ( CanAdvance . keepFirstNotifyTrueAndRemoveFollowings  )
+    . nextGeneralizedInputsMaybe environment state
+    $ x :| xs
+next environment state  contract  = uncurry (next environment) =<< reduceContract environment state contract
```
jhbertra, line 88: All non-when cases could be treated the same, and should all set `canAdvance`:

```suggestion
next environment state contract = case reduceContractUntilQuiescent environment state contract of
  RRAmbiguousTimeIntervalError -> Left AmbiguousIntervalProvided
  ContractQuiescent reduced _ _ newState newContract -> Right case newContract of
    When (x : xs) _ _ ->
      let
        indexedCases = (CaseIndex 0, x) :| zip (CaseIndex <$> [1..]) xs
      in
        Next
          (if reduced then CanAdvance else CannotAdvance)
          (canNotify environment state indexedCases)
          (canDeposit environment state indexedCases)
          (canChoose environment state indexedCases)
    _ -> Next CanAdvance CannotNotify [] []
```

definitions for `canNotify`, `canDeposit`, `canChoose` omitted for brevity

marlowe/src/Language/Marlowe/Core/V1/Semantics/Next.hs
```diff
@@ -0,0 +1,175 @@
...
+    deriving anyclass (Pretty)
+    deriving (FromJSON,ToJSON)
+
+-- | Next Ranges of Inputs that can be applied to a given contract
+data NextGeneralizedInput
+  = CanDeposit CaseIndex Party AccountId Token Integer -- 1 possibility > 1 action
+  | CanChoose  CaseIndex ChoiceId [Bound] -- 1 possibility > many different actions
+  | CanNotify  CaseIndex -- 1 possibility > 1 action
+  deriving stock (Haskell.Show,Haskell.Eq,Haskell.Ord,Generic)
+  deriving anyclass (Pretty)
```
jhbertra, line 75: I want to suggest a couple of changes to this schema:

1. It is possible for a contract to be advanced or have inputs applied, so these should not be mutually exclusive cases.
2. It would be more useful if the input types were separated rather than combined in a sum type
   - It would be an error for more than one `CanNotify` to appear in the list
   - It makes it easier to ask questions like "can I make a deposit" or "can I make a choice" / "what deposits are available?", "what choices are available?"

```suggestion
data Next = Next
  { nextCanAdvance :: CanAdvance
  , nextCanNotify :: CanNotify
  , nextCanDeposit :: [CanDeposit]
  , nextCanChoose :: [CanChoose]
  }
  deriving stock (Haskell.Show, Haskell.Ord, Haskell.Eq, Generic)
  
data CanAdvance
  = CannotAdvance
  | CanAdvance
  deriving stock (Haskell.Show, Haskell.Ord, Haskell.Eq, Generic)
  deriving anyclass (Pretty)

newtype CaseIndex = CaseIndex {unIndex ::Integer}
    deriving stock (Haskell.Show,Haskell.Eq,Haskell.Ord,Generic)
    deriving anyclass (Pretty)
    deriving (FromJSON,ToJSON)

data CanNotify
  = CannotNotify
  | CanNotify CaseIndex (Maybe BuiltinByteString) -- Hash of the required merkleized continuation, if applicable
  deriving stock (Haskell.Show, Haskell.Ord, Haskell.Eq, Generic)
  deriving anyclass (Pretty)
  
data CanDeposist = CanDeposit CaseIndex Party AccountId Token Integer (Maybe BuiltinByteString)
  deriving stock (Haskell.Show,Haskell.Eq,Haskell.Ord,Generic)
  deriving anyclass (Pretty)

data CanChoose = CanChoose CaseIndex ChoiceId [Bound] (Maybe BuiltinByteString)
  deriving stock (Haskell.Show,Haskell.Eq,Haskell.Ord,Generic)
  deriving anyclass (Pretty)

data AmbiguousIntervalProvided = AmbiguousIntervalProvided
```

scripts/local-docker.env
```diff
@@ -10,4 +10,6 @@ export MARLOWE_RT_TRACED_PORT=$(docker-compose port marlowe-proxy 3701 | sed -e
 export MARLOWE_RT_REST_PORT=$(docker-compose port web 8080 | sed -e 's/.*://')
 export MARLOWE_RT_REST_URI=http://localhost:$MARLOWE_RT_REST_PORT
 
+export MARLOWE_RT_REST_OPEN_API_URI=$MARLOWE_RT_REST_URI/openapi.json
+
```
jhbertra, line 14: This seems to be of dubious value to add in here

jhbertra (changes requested):

marlowe/src/Language/Marlowe/Core/V1/Semantics/Next.hs
palas: The Advance bit has another issue, it is not possible to tell whether empty inputs is allowed or not (that is what I was understanding by "CanAdvance"). You can find this out from the result of `reduceContractUntilQuiescent`, the first parameter is true if "empty inputs" is allowed, and false if not.

But that is independent of the rest of potential inputs, you can have the four cases:
- Empty inputs allowed, no other inputs allowed
- Empty inputs allowed, other inputs allowed
- Empty inputs disallowed, no other inputs allowed
- Empty inputs disallowed, other inputs allowed
These definitions are only considering 3 of them, I think they are missing the last one

marlowe/src/Language/Marlowe/Core/V1/Semantics/Next.hs
palas: Exactly, but need to also consider the "empty inputs allowed" and "empty inputs disallowed" cases, which we get from the first parameter of `ContractQuiescent`, and we probably should throw an error `500` if we get anything but `When` or `Close` out of `reduceContractUntilQuiescent`

marlowe/src/Language/Marlowe/Core/V1/Semantics/Next.hs
```diff
@@ -0,0 +1,175 @@
...
+  = Right
+    . maybe
+          CanNotAdvance -- Contract Suspended (only falsified Notifies evaluated in the When)
+        ( CanAdvance . keepFirstNotifyTrueAndRemoveFollowings  )
+    . nextGeneralizedInputsMaybe environment state
+    $ x :| xs
+next environment state  contract  = uncurry (next environment) =<< reduceContract environment state contract
+
+
+keepFirstNotifyTrueAndRemoveFollowings ::  NonEmpty NextGeneralizedInput -> [NextGeneralizedInput]
```
palas:
```suggestion
keepFirstNotifyTrueAndRemoveFollowings ::  NonEmpty NextGeneralizedInput -> [NextGeneralizedInput]
```
Why consider empty list as a special case and then we have a `NonEmpty` list, if we could just have a normal list that considers both cases and is simpler and more widely understood?

In other words, instead of:
```haskell
  CanApplyInputs (NonEmpty NextGeneralizedInput)
| Suspended
``` 
Why not just write `[NextGeneralizedInput]`? But because we need to consider apply with no inputs, we actually need something like `(Bool, [NextGeneralizedInput])`, or something along those lines

marlowe/src/Language/Marlowe/Core/V1/Semantics/Next.hs
```diff
@@ -0,0 +1,175 @@
...
+keepFirstNotifyTrueAndRemoveFollowings
+  = keepFirstNotifyTrueAndRemoveFollowings' False . NEL.toList
+ where
+  keepFirstNotifyTrueAndRemoveFollowings' :: Bool -> [NextGeneralizedInput] -> [NextGeneralizedInput]
+  keepFirstNotifyTrueAndRemoveFollowings' _ [] =  []
+  keepFirstNotifyTrueAndRemoveFollowings' False (x@CanNotify {} : xs) =  keepFirstNotifyTrueAndRemoveFollowings' True (x:xs)
+  keepFirstNotifyTrueAndRemoveFollowings' True (CanNotify {} : xs) =  keepFirstNotifyTrueAndRemoveFollowings' True xs
+  keepFirstNotifyTrueAndRemoveFollowings' _ xs =  keepFirstNotifyTrueAndRemoveFollowings' True xs
```
palas: You could use `nubBy` for this. But actually, I realised you would need to do something similar for `Choices` (and for deposits, but that you could do with `nubBy`). For example, and this is pseudo-haskell:
```haskell
[ Notify FalseObs cont1
, Notify TrueObs cont2
, Deposit 153 cont3
, Choice [(2,5)] cont4
, Choice [(0,7)] cont5
, Deposit 153 cont6
, Deposit 155 cont7
, Notify TrueObs cont8
]
```
Should return
```haskell
[(2, Notify), (3, Deposit 153), (4, Choice [(2,5)]), (5, Choice [(0,1), (6, 7)]), (7, Deposit 155)]
```

It is easier if you don't return the branch position, but still needs some work

marlowe/src/Language/Marlowe/Core/V1/Semantics/Next.hs
```diff
@@ -0,0 +1,175 @@
...
+                token
+                (evalValue environment state value )
+      (Choice choiceId bounds) -> Just $ CanChoose caseIndex choiceId bounds
+      (Notify observation) | evalObservation environment state observation -> Just $ CanNotify caseIndex
+      (Notify _) -> Nothing
+  . \case
+      (Case action _)           -> action
+      (MerkleizedCase action _) -> action
+
+caseIndexed :: NonEmpty a -> [(CaseIndex,a)]
```
palas, line 158: Huh, so you had the case position separate initially, if you don't combine it with `NextGeneralizedInput` the whole `nextGeneralizedInputMaybe` gets simpler

marlowe/src/Language/Marlowe/Core/V1/Semantics/Types.hs
```diff
@@ -307,6 +314,13 @@ data State = State { accounts    :: Accounts
 newtype Environment = Environment { timeInterval :: TimeInterval }
   deriving stock (Haskell.Show,Haskell.Eq,Haskell.Ord)
 
+environment :: UTCTime -> UTCTime -> Environment
+environment start end = Environment (utcTimeToPOSIXTime start, utcTimeToPOSIXTime end)
+
+isTimedOut :: POSIXTime -> Environment -> Bool
+isTimedOut t Environment {timeInterval = (_,b)} | t  > b = True
+isTimedOut _ _ = False
```
palas, line 323:
```suggestion
```
This doesn't seem to be used, and also I think it is reversed, if it is the timeout, then it should be before the time interval right?

palas (changes requested):
It is looking better, but there are still some corner cases that are not considered

marlowe/src/Language/Marlowe/Core/V1/Semantics/Next.hs
nhenin: used a fold in the new version

marlowe-runtime-web/marlowe-runtime-web.cabal
```diff
@@ -1,25 +1,24 @@
...
-category: Language
+name:          marlowe-runtime-web
+version:       0.0.1
+synopsis:      Web server for Marlowe Runtime on Cardano.
+bug-reports:   https://github.com/input-output-hk/marlowe-cardano/issues
+license:       Apache-2.0
+author:        Jamie Bertram
+maintainer:    jamie.bertram@iohk.io
+stability:     experimental
+category:      Language
```
jhbertra, line 10: Can you please revert the formatting changes?

marlowe-runtime-web/server/Language/Marlowe/Runtime/Web/Server/REST/Contracts/Next.hs
```diff
@@ -0,0 +1,53 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GADTs #-}
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
+{-# HLINT ignore "Use <&>" #-}
```
jhbertra, line 4: Why?

marlowe/marlowe-cardano.cabal
```diff
@@ -1,88 +1,98 @@
-cabal-version: 2.2
-name: marlowe-cardano
-version: 0.1.0.3
-license: Apache-2.0
+cabal-version:   2.2
+name:            marlowe-cardano
+version:         0.1.0.3
+license:         Apache-2.0
```
jhbertra, line 4: Can you please revert the formatting changes?

jhbertra (changes requested):
Looks much better - the deposit and choice cases also need the merkleized continuation hash though, not just notify. Also still needs tests for `next`.

marlowe-runtime-web/marlowe-runtime-web.cabal
nhenin: Done.

marlowe-runtime-web/server/Language/Marlowe/Runtime/Web/Server/REST/Contracts/Next.hs
nhenin: sometimes, when I'm implementing functions, hlint proposes reductions that not readable anymore.. but usually this is temporary, and in that case it doesn't apply anymore.. I'll remove it...

marlowe-runtime-web/marlowe-runtime-web.cabal
nhenin: yeah I don't know how this has been formatted...

marlowe/src/Language/Marlowe/Core/V1/Semantics/Next.hs
```diff
@@ -0,0 +1,290 @@
...
+  = Indexed CaseIndex a
+    deriving stock (Haskell.Show,Haskell.Eq,Haskell.Ord,Generic)
+    deriving anyclass (Pretty)
+
+toCaseIndex :: Indexed a -> CaseIndex
+toCaseIndex (Indexed c _) = c
+
+instance Haskell.Semigroup ApplicableGeneralizedInputs where
+  a <> b  = ApplicableGeneralizedInputs
+              (getFirst $ (coerce . canNotifyMaybe $ a) <> (coerce . canNotifyMaybe $ b)) -- First Monoid
```
palas, line 114: This is really unreadable and complicated and hard to understand. Why not use `firstJust` or something like that?
```suggestion
              (firstJust (canNotifyMaybe a) (canNotifyMaybe b))
```

marlowe/src/Language/Marlowe/Core/V1/Semantics/Next.hs
```diff
@@ -0,0 +1,290 @@
...
+              (deposits a   <> deposits b) -- (++) monoid
+              (choices a    <> choices b ) -- (++) monoid
```
palas, line 116: I really doubt the default implementation of Monoid for `CanChoose` does the right thing

palas (changes requested):
Hey Nicolas,
Thanks, it is looking much better. I have written like 80% linting and styling advice. There are a couple of things that I think need to be addressed, like the combination of choices, I haven't tried, but by looking at the code I doubt it works as it should, because it needs combining choice intervals. And the exception of Close, I think we should remove for accuracy, even though it is a bit of an issue that it is the way it is.

marlowe/src/Language/Marlowe/Core/V1/Semantics/Next.hs
nhenin: CanChoose or CanNotify ? , I think he does because It is tested ...  
it's ~ to the monoid (First Maybe) => 
`Just y <> Just x = Just y`
`Nothing <> Just x = Just x`
`Nothing <> Nothing = Nothing`

marlowe/src/Language/Marlowe/Core/V1/Semantics/Next.hs
```diff
@@ -0,0 +1,290 @@
...
+    (canReduce,reducedState,reducedContract) <- reduceContract environment state contract
+    Right
+      $ Next
+        canReduce
+        (applicables environment reducedState reducedContract)
+
+-- | Describe for a given contract which inputs can be applied
+applicables :: Environment  -> State  ->  Contract -> ApplicableGeneralizedInputs
+applicables  _  _ Close = mempty
+applicables environment  state (When xs _ _) = applicableGeneralizedInputs' environment state xs
```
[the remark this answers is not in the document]
nhenin: That kind of formatting, I'm relying on the formatter trigger in the pre-commit stage... I don't know which one you use but I'll prefer we use the same one within the team

marlowe/src/Language/Marlowe/Core/V1/Semantics/Types.hs
```diff
@@ -315,6 +334,8 @@ instance ToJSON Environment where
   toJSON Environment{..} = object
     [ "timeInterval" .= posixIntervalToJSON timeInterval]
 
+
+
```
palas, line 338: Could you revert this file, please? This file has been audited, we shouldn't be making changes to it, and I think none of these functions are used. You can put them in `/marlowe/src/Language/Marlowe/Utils.hs`, also I don't think the whole `Next` folder should be under the `Semantics` folder, for the same reason. Also making changes to the semantics requires updating the formalisation and spec in `marlowe` repo to match. I see why we may want them under `V1`, but under `Semantics`, I am not sure

palas (changes requested):
Looking good! Just a few minor issues. I suspect tests could be improved a lot, but I haven't looked much at them and they can probably wait for another ticket

marlowe-runtime-web/marlowe-runtime-web.cabal
```diff
@@ -54,31 +54,32 @@ library
   exposed-modules:
```
jhbertra, line 54: Can you please run `cabal-fmt --inplace --no-tabular **/*.cabal && git commit -am "format cabal files" --no-verify`?

marlowe-runtime-web/src/Language/Marlowe/Runtime/Web/Orphans.hs
```diff
@@ -1,15 +1,34 @@
 {-# LANGUAGE OverloadedLists #-}
 
 {-# OPTIONS_GHC -Wno-orphans #-}
+{-# LANGUAGE FlexibleInstances #-}
```
jhbertra, line 4:
```suggestion
```

There aren't any code changes in here, so I assume this is not needed

marlowe-test/src/Spec/Marlowe/Semantics/Next/Contract/Generator.hs
```diff
@@ -0,0 +1,111 @@
...
+import Data.List (nub)
+import Spec.Marlowe.Semantics.Next.Common.Tuple (uncurry3)
+import Spec.Marlowe.Semantics.Next.Contract
+  (hasValidEnvironement, isEmptyWhenNonTimedOut, isIrreducible, isNotClose, isReducible, isReducibleToClose)
+import Spec.Marlowe.Semantics.Next.Contract.When.Deposit
+import Spec.Marlowe.Semantics.Next.Contract.When.Notify (areOnlyFalsifiedNotifies, atLeastOneNotifyTrue)
+
+
+anyReducibleContract :: Gen (Environment,State,Contract)
+anyReducibleContract
```
jhbertra, line 36: These generators will all be very slow because of the pervasive use of `suchThat`

marlowe/src/Language/Marlowe/Core/V1/Plate.hs
```diff
@@ -22,15 +22,15 @@ import Data.Generics.Multiplate (Multiplate(..), foldFor, mChildren, preorderFol
...
 import Language.Marlowe.Core.V1.Semantics.Types.Address (Network)
-import Plutus.V1.Ledger.Api (BuiltinByteString, TokenName)
 
 import qualified Data.Functor.Constant as F (Constant(..))
 import qualified Data.Map.Strict as M (foldr)
 import qualified Data.Set as S (Set, empty, fromList, map, singleton, toList, union)
 import qualified Plutus.V2.Ledger.Api as Ledger (Address)
 import qualified PlutusTx.AssocMap as AM (keys)
-
+import PlutusTx.Prelude (BuiltinByteString)
```
jhbertra, line 33:
```suggestion

```

marlowe/src/Language/Marlowe/Core/V1/Plate.hs
```diff
@@ -22,15 +22,15 @@ import Data.Generics.Multiplate (Multiplate(..), foldFor, mChildren, preorderFol
 import Data.Maybe (mapMaybe)
```
jhbertra, line 22: Revert no-op changes?

```
git ch main -- marlowe/src/Language/Marlowe/Core/V1/Plate.hs
```

jhbertra (approved):
LGTM

PR #607, opened 2023-06-13
PLT-6201 Allow tracing to become a true noop

jhbertra:
- [x] Make tracing a no-op if the `OTEL_EXPORTER_OTLP_ENDPOINT` env var is not set (this is not standard, but is convenient in our case).
  - No-op means that the global tracer provider is not initialized and that a `noopEventBackend` is used, so selectors and fields are never held in memory past creation and are not processed into spans and attributes.
- [x] Refactor: consolidate duplicated monad stack into a single module (used in all runtime components, web server, and integration tests).
- [x] Refactor: consolidate tracer initialization to the same place. 

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
        - Review required
    - [x] Reviewer requested

### DIFF
--- a/async-components/async-components.cabal
+++ b/async-components/async-components.cabal
@@ -49,11 +49,21 @@ library
   exposed-modules:
     Control.Concurrent.Component
     Control.Concurrent.Component.Probes
+    Control.Concurrent.Component.Run
   build-depends:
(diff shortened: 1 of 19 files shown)


bwbush (approved):
Nice!

PR #616, opened 2023-06-16
Fix static cross-building with musl.

shlevy:
This involved updating `iogx` to use the latest `iohk-nix` with its fixed crypto overlays, and removing the old ad hoc configuration to make libsodium available to needed builds (which selected the host package when cross-compiling).

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -781,6 +781,23 @@
         "type": "github"
       }
     },
+    "blst": {
+      "flake": false,
+      "locked": {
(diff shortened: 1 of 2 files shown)


jhbertra:
There doesn't seem to be much substance to these changes - could you add a pull request description that explains what has changed? Was it just a flake input that needed bumping?

shlevy:
@jhbertra updated

zeme-wana:
they are gone for good now

PR #632, opened 2023-06-25
Bump IOGX

zeme-wana:
- Updates to the nix interface files as per the lastest IOGX version.
- x-builds on windows still a WIP


jhbertra:
What changes does this include?

zeme-wana:
> What changes does this include?

I've updated the description

PR #651, opened 2023-07-05
Bump IOGX

zeme-wana:


cabal.project
```diff
@@ -165,14 +165,12 @@ source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-config
   tag: 1646e9167fab36c0bff82317743b96efa2d3adaa
-  --sha256: 11kf65x38laqhwspsl28j2x5a4rky8mfr6356w0li5g53sfykmjc
```
shlevy, line 168: I strongly prefer the inline comments as the sha256map is in a different file and the cabal.project is no longer standalone for h.nix. Is there a reason to change this?

zeme-wana: It was a somewhat desperate attempt to invalidate the hydra caches and fix the aborted builds on Darwin. 
I will revert it

PR #661, opened 2023-07-17
Add custom headers to exposed list on the server side

paluh:
Quick fix which adds headers to `Access-Control-Expose-Headers` list. It probably requires a fix release. Currently POST endpoints for instances on demeter.run are unusable from any web browser client. 

We don't catch CORS errors in the test suite unfortunately.


paluh:
I've to retest if this change make sens.

bwbush (approved):

paluh:
I'm closing this for now because I wasn't able to reproduce CORS error when I tested my local runner against the staging servers.

PR #674, opened 2023-08-02
Monadic Contract Bundler

jhbertra:
- [x] Adds a `Bundler` module containing a `BundlerT` monad transformer (and associated `MonadBundler` typeclass (bad typeclass with no laws but the one method is a bit of a black hole purely from the point of view of the class so :shrug:)
- [x] Adds tests for `BuntlerT` to make sure it satisfies typeclass laws etc...

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
        - Review not required
        - [x] Minor changes to non-critical code, documentation, nix derivations, configuration files, or scripts
    - [x] Reviewer requested

### DIFF
--- a/marlowe-object/src/Language/Marlowe/Object/Bundler.hs
+++ b/marlowe-object/src/Language/Marlowe/Object/Bundler.hs
@@ -1,0 +1,192 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+-- | A monad for conveniently defining Marlowe object bundles. It allows one to define a Marlowe contract
+--   with named abstractions for commonly-used sub-contracts, and under the hood it converts it to an
+--   object definition and replaces usages with a reference.
+module Language.Marlowe.Object.Bundler where
+
+import Control.Applicative (Applicative (liftA2))
+import Control.Monad.Cont (ContT)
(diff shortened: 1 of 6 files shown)


marlowe-object/src/Language/Marlowe/Object/Bundler.hs
```diff
@@ -0,0 +1,231 @@
...
+  catchError b f = go b
+    where
+      go = \case
+        Pure a -> Pure a
+        M m -> M $ (go <$> m) `catchError` (pure . f)
+        Define obj b' -> Define obj $ go b'
+
+-- | Run a bundler action, returning the resulting object bundle and the result of the action.
+runBundlerT :: (Monad m) => BundlerT m a -> m (ObjectBundle, a)
+runBundlerT = (fmap . first) ObjectBundle . toListM' . runBundlerTIncremental
```
yveshauser, line 177: To use of `toListM'` here hints to reconsider having the `[]` in `ObjectBundle`. Could be replace by a `Producer`, couldn't it?

yveshauser (approved):
Thanks! We successfully tested the bundle builder for the `Future` contract.

PS: There have been discussions of using the `pipes` library also for defining `BundlerT` directly as a `Producer`: 
```haskell
type BundlerT m a = Producer LabelledObject m a
```
In order to have clear API boundaries this has been discarded for now

marlowe-object/src/Language/Marlowe/Object/Bundler.hs
jhbertra: I suppose it could, but that would be a pretty major redesign of `ObjectBundle` and its interface at this point

PR #683, opened 2023-08-07
Bump IOGX V3

zeme-wana:
Changes to the interface:

1. `./nix/iogx-config.nix` renamed to -> `./nix/haskell.nix`
    - `iogx-config.nix:haskellCompilers` -> `./nix/haskell.nix:supportedCompilers`
    - `iogx-config.nix:defaultHaskellCompiler` -> `./nix/haskell.nix:defaultCompiler`
    - `iogx-config.nix:shouldCrossCompile` -> `./nix/haskell.nix:enableCrossCompilation`

2. `./nix/haskell-project.nix` renamed to -> `./nix/cabal-project.nix`

3. `./nix/pre-commit-check.nix` renamed to -> `./nix/formatters.nix`

4. `./nix/hydra-jobs.nix` renamed to -> `./nix/ci.nix`

5. `inputs` and `inputs'` have been swapped.
  `inputs` (no prime `'`) is now the original `inputs`, non-desystemised
  `inputs'` (with prime `'`) is now the de-systemized `inputs`
Reason behind this is that `inputs` is universally understood to be the original, unmodified flake inputs.
While `inputs'` is "non-standard" and thus warrants a different name.
This is the same notation used by [flake.parts](https://flake.parts/module-arguments#inputs)

6. The flake exports aliases for your haskell *executables* only in `outputs.apps`.
   For example, `outputs.apps.marlowe-apps-exe-marlowe-finder-ghc8107` also has an 
   alias `outputs.apps.marlowe-finder` (the cabal package name and the `exe` infix have been stripped)
   These can be access as usual with `inputs'.self.apps.marlowe-finder`

7. There is a new parameters to the interface files named `nix`, which can be used to reference files in your `./nix` folder by name using attribute sets, instead of using the import keyword. See this for documentation https://github.com/input-output-hk/iogx#301-nix


.github/workflows/test.yml
```diff
@@ -67,7 +67,7 @@ jobs:
           authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
       - name: Run tests
         run: |
-          nix run .#marlowe-integration-tests --show-trace -- --strict
```
zeme-wana, line 70: The reason why I renamed this is that the flake output would shadow the testsuite named `marlowe-integration-tests` in the `marlowe-integration-tests/marlowe-integration-tests.cabal`

jhbertra (approved):
LGTM

nix/marlowe-cardano/integration-tests.nix
```diff
@@ -1,13 +1,15 @@
-{ inputs, pkgs, projects }:
+{ inputs', pkgs, l, ... }:
+
+{ projects }:
 
 let
-  cardano-cli = inputs.cardano-world.cardano.packages.cardano-cli;
-  cardano-node = inputs.cardano-world.cardano.packages.cardano-node;
-  marlowe-runtime-cli = projects.ghc8107.hsPkgs.marlowe-runtime-cli.components.exes.marlowe-runtime-cli;
```
zeme-wana, line 6: You can support multiple ghc compilers in your project.
See `./nix/cabal-project.nix:supportedCompilers`, and so `projects` will contain an attribute for each compiler name.
However you can now use the `default` project which will use the `./nix/cabal-project.nix:defaultCompiler` (which is set the first compiler in `supportedCompilers` by default)

zeme-wana:
The flake output names have changed again (for the last time I promise).

Because your project only supports one compiler (see `./nix/cabal-project.nix:supportedCompilers`) there is no need to disambiguate between compiler names, and so the compiler name `"-ghc8107"` is no longer prefixed to the `packages|checks|apps` flake outputs.
```
inputs.'self.packages.marlowe-runtime-web-exe-marlowe-web-server-ghc8107 ->
inputs.'self.packages.marlowe-runtime-web-exe-marlowe-web-server
```

In addition, aliases for your cabal *executables* only (this excludes benchmarks and test suites) are *added* to your `packages` and `apps` flake outputs (the original, longer names are still kept).
"Alias" means just the executable name as found in your cabal file, without the `$CabalPackageName-exe-` prefix.
```
inputs.'self.packages.marlowe-runtime-web-exe-marlowe-web-server ->
inputs.'self.packages.marlowe-web-server

inputs.'self.apps.marlowe-runtime-web-exe-marlowe-web-server ->
inputs.'self.apps.marlowe-web-server
```

If you cabal project has multiple executables with the same name across your cabal files, a warning will be generated when running `list-flake-outputs` and no aliases will be created.

@bwbush 
Last time the flake output names were changed, things broke.
Please point out what other repositories call into the `marlowe-cardano` flake, and I'll go and update them myself.
Thank you for your patience.

PR #687, opened 2023-08-11
PLT-6750: Parsing pretty printed Marlowe contracts

yveshauser:
Pre-submit checklist:
- Branch
    - [ ] Formatting, PNG optimization, etc. are updated
- PR
    - [ ] Self-reviewed the diff
        - Review required
        - [ ] Substantial changes to code, test, or documentation
        - Review not required
    - [ ] Reviewer requested


marlowe-cli/command/Language/Marlowe/CLI/Command/Format.hs
```diff
@@ -0,0 +1,122 @@
...
+      O.optional . O.strOption $
+        mconcat
+          [ O.long "out-file"
+          , O.metavar "MARLOWE_FILE"
+          , O.help "The Marlowe file containing the output contract. Stdout if omitted."
+          ]
+    inFormatParser =
+      O.optional . O.option formatReader $
+        mconcat
+          [ O.long "in"
```
palas, line 99:
```suggestion
          [ O.long "in-format"
```
Even thought it would be more verbose, it may be good to call this `in-format`, because `in` is a bit ambiguous. Other alternatives would be:
- Having `--to-json`, `--to-yaml`, `--to-marlowe` as a flag instead, have `<in_file>` and `<out_file>` be required, and then infer <in_file> from the content
- Having combined arguments `input.json:json` `output.marlowe:marlowe`

palas (changes requested):
Nicely done! I am quite convinced the code is correct, and the test you wrote should give good coverage. I have written a couple of suggestions for improved usability and a couple of other minor things

marlowe-cli/command/Language/Marlowe/CLI/Command/Format.hs
palas: Same for `out`

yveshauser: Agree, changed to `in-format` and `out-format`. 

I prefer to have those arguments, otherwise also the flags `--from-json`, `--from-yaml` and `--from-marlowe` would be needed in order to read the contract from stdin. That's probably too many flags.

Combined arguments allowing as well `stdin` and `stdout`  instead of the file name would work.

marlowe-cli/command/Language/Marlowe/CLI/Command/Format.hs
```diff
@@ -0,0 +1,143 @@
...
+  :: (MonadError CliError m)
+  => (MonadIO m)
+  => FormatCommand
+  -- ^ The command.
+  -> m ()
+  -- ^ Action for runninng the command.
+runFormatCommand Format{..} =
+  do
+    contract <- case inFormat of
+      Just Json -> readContractJson inputFile
```
bwbush, line 64: JSON is a subset of YAML, so in principle we'd just need to read YAML. This might confuse folks, however.

marlowe-cli/marlowe-cli.cabal
```diff
@@ -107,10 +108,12 @@ library
     , errors
     , extra
     , marlowe-cardano
+    , megaparsec
```
bwbush, line 111: I think that `attoparsec` performs better, but that's unlikely to matter for contract files.

marlowe-cli/src/Language/Marlowe/CLI/Format.hs
```diff
@@ -0,0 +1,439 @@
...
+import Language.Marlowe.Pretty
+import Plutus.V1.Ledger.Time (POSIXTime (..))
+import PlutusTx.Builtins (BuiltinByteString)
+import PlutusTx.Builtins.Class (ToBuiltin (..))
+import Text.Megaparsec
+import Text.Megaparsec.Char as P
+import Text.Megaparsec.Char.Lexer qualified as L
+
+-- | Read a Marlowe Contract as JSON from a file.
+readContractJson
```
bwbush, line 66: This function is called for both JSON and YAML input. Does `Data.Aeson.eitherDecodeStrict` handle YAML in addition to JSON, or do we need to use `Data.Yaml.decodeFileEither`?

marlowe-cli/src/Language/Marlowe/CLI/Format.hs
```diff
@@ -0,0 +1,439 @@
...
+  -> m ()
+maybeWritePretty Nothing =
+  liftIO . print . pretty
+maybeWritePretty (Just outputFile) =
+  liftIO . writeFile outputFile . show . pretty
+
+-- | Parser for the pretty printed Marlowe Contract data type.
+contractParser
+  :: ParsecT Void Text m Contract
+contractParser = contractParser' <|> parens contractParser
```
bwbush, line 108: Nice!

marlowe-cli/src/Language/Marlowe/CLI/Format.hs
```diff
@@ -0,0 +1,439 @@
...
+-- | Read a Marlowe Contract as JSON from a file.
+readContractJson
+  :: (MonadError CliError m)
+  => (MonadIO m)
+  => Maybe FilePath
+  -> m Contract
+readContractJson (Just inputFile) =
+  decodeFileStrict inputFile
+readContractJson Nothing =
+  (liftEither . first (CliError . show)) . eitherDecodeStrict
```
bwbush, line 74: This function is called for both JSON and YAML input. Does `Data.Aeson.eitherDecodeStrict` handle YAML in addition to JSON, or do we need to use `Data.Yaml.decodeEither`?

bwbush (approved):
Looks good, aside from the one concern I noted in a comment. Thanks!

marlowe-cli/command/Language/Marlowe/CLI/Command/Format.hs
palas: Yes, for `stdin` and `stdout` we can either have `-` (dash) mean that (that is pretty standard), or just omit `--in-file` or `--out-file`, but no need to implement that for now

marlowe-cli/command/Language/Marlowe/CLI/Command/Format.hs
```diff
@@ -0,0 +1,144 @@
...
+    <*> outFile
+    <*> inFormatParser
+    <*> outFormatParser
+  where
+    inFile =
+      O.optional . O.strOption $
+        mconcat
+          [ O.long "in-file"
+          , O.metavar "MARLOWE_FILE"
+          , O.help "The Marlowe file containing the contract. If omitted, the Marlowe contract is read from stdin."
```
palas, line 105: Oh, I see you already do that

palas (approved):
Nice :+1:

PR #703, opened 2023-09-04
[Work In Progress] Reduce closure size of `nix develop` shell 

zeme-wana:
The `test.yml` workflow has been failing for a while with an out-of-disk error.

This is because the current closure size of the `nix develop` shell is `13.9G`, and the default disk size of the default GH runners is `14G`.

These are heaviest dependencies account for `8.8G`:
```
 270M /nix/store/ldgckm4sq2faarr596za470jz2s31zf4-ghc-lib-parser-lib-ghc-lib-parser-8.10.7.20220219
 400M /nix/store/rkvj58krwc7adi5bmf90v80s9x7rx4n5-ghc-lib-ghc-8.10.7                               
 550M /nix/store/c0xs3ady8sfvmvi7rhr9qp168w2pk2a6-ghc-8.10.7-doc                                   
 550M /nix/store/5a2xz9hqya7x2a4l13j4ssxarxrjind9-ghc-8.10.7-doc                                   
 670M /nix/store/mm077pca7jzy2xp6jgvqi7izldjbrp2l-ghc-9.2.8-doc                                    
 680M /nix/store/l1121l59mmgdg1caksn457yhrh5rsp04-ghc-9.2.7-doc                                    
1420M /nix/store/4x5vlfyzqlfdybm8f367lkl7vvx1lxgz-ghc-8.10.7                                       
1420M /nix/store/kn68p74mj82czdf44c3db63kdzq0c8hz-ghc-8.10.7                                       
1420M /nix/store/bsxzmawic6irvr6xla2rsdaxnqznmfa3-ghc-9.2.7                                        
1440M /nix/store/45b58qc554p4vz3qny9v5cs7nsvzrq1n-ghc-9.2.8                                                                            
```

Note that there are 4 GHCs, including two `8.10.7`s, and they all come with a `-doc`.

1. `ghc-9.2.7` comes from:
```nix
# shell.nix
packages = [
  pkgs.haskellPackages.hspec-golden
  ...
];
```
Which uses the ghc coming from nixpkgs, so that's easy to fix.
Can we remove `hspec-golden`? @jhbertra @bwbush 
This line accounts for `2.1G`

2. `cardano-world` comes from:
```nix
# shell.nix
packages = [
  inputs.cardano-world.cardano.packages.cardano-address
  inputs.cardano-world.cardano.packages.cardano-node
  inputs.cardano-world.cardano.packages.cardano-cli
];
```
We can't really get rid of these. 
However `cardano-world` is using a different ghc8107, and that explains why we end up with 2.
These lines accounts for `2.7G`

3. Finally `ghc-9.2.8` comes from `fourmolu`:
```nix 
# formatters.nix
{
  fourmolu.enable = true;
}
```
And defined [here](https://github.com/input-output-hk/iogx/blob/v4/src/ext/fourmolu.nix) 
This line accounts for `2.2G`


.github/workflows/test.yml
```diff
@@ -18,22 +16,17 @@ jobs:
...
-          nix_path: nixpkgs=channel:nixos-unstable
-          install_url: https://releases.nixos.org/nix/nix-2.10.3/install
+          nix_path: nixpkgs=channel:nixos-2305
+          install_url: https://releases.nixos.org/nix/nix-2.16.0/install
           extra_nix_config: |
-            allowed-uris = ${{ env.ALLOWED_URIS }}
             trusted-public-keys = ${{ env.TRUSTED_PUBLIC_KEYS }}
             substituters = ${{ env.SUBSTITUTERS }}
             experimental-features = nix-command flakes
-      - name: Setup Cachix
```
zeme-wana, line 28: I really think we can do without cachix. 
Everything is cached on hydra anyway.
Having cachix might actually be slowing things down due to upload overheads.
I'm testing this workflow on my fork and it runs quick/same-time https://github.com/zeme-wana/marlowe-cardano/actions

nix/shell.nix
```diff
@@ -19,11 +19,15 @@ in
     inputs'.cardano-world.cardano.packages.cardano-address
     inputs'.cardano-world.cardano.packages.cardano-node
     inputs'.cardano-world.cardano.packages.cardano-cli
+    # ^^^^^ 2.7G
 
     pkgs.z3
     pkgs.sqitchPg
     pkgs.postgresql
+
     pkgs.haskellPackages.hspec-golden
```
bwbush, line 28: Is there some way to pin `hspec-golden` so that is uses the same GHC as the rest of the build?

bwbush (approved):
Would there be a way to run the test in an alternative, stripped-down `nix-shell`, instead of the one used by developers? That stripped-down version could omit to unneeded tooling and just contain what is needed for the cabal build and test.

angerman:
```
1420M /nix/store/4x5vlfyzqlfdybm8f367lkl7vvx1lxgz-ghc-8.10.7                                       
1420M /nix/store/kn68p74mj82czdf44c3db63kdzq0c8hz-ghc-8.10.7                                       
1420M /nix/store/bsxzmawic6irvr6xla2rsdaxnqznmfa3-ghc-9.2.7                                        
1440M /nix/store/45b58qc554p4vz3qny9v5cs7nsvzrq1n-ghc-9.2.8
```
[`cardano-parts` recently switched to using `fetchClosure`](https://github.com/input-output-hk/cardano-parts/pull/8), maybe that helps a bit.

It would certainly be good to check if we can ensure both ghc-8.10 are the same.

- can you build `fourmolu` just statically? That should reduce the closure size quite a bit?
- I'm not sure you can build `hspec-golden`, also statically (maybe it depends on some dynamism or needs to call out to ghc); but pulling that from the same set as `fourmolu` would make sure both come from the same ghc.

zeme-wana:
@bwbush I've replaced `hspec-golden` like you suggested.
Note: the `hspec-golden` package provides one executable named `hgold`, which does *not* end up in the shell if using `pkgs.haskellPackages.hspec-golden`. Reasonably this means that `hspec-golden` never needed to be added to the shell in the first place. But it costs us nothing to have it now.

`fourmolu` comes from IOGX so I'll have to patch it upstream.

As for using `cardano-parts`, it looks like a non-trivial change so it probably warrants its own PR.

As for making a new minimal shell, it can certainly be done but it would add maintenance costs, and I suspect it wouldn't reduce the closure size to a point that makes it worth the effort. But that can only be proved empirically.
After fixing `fourmolu` and using `fetchClousure` we will most likely end up with a very light shell anyway.

The closure size is now 11.8G (down from 14G). 
This will be enough to fix the out-of-disk error.

So if you are ok with this we can merge this PR.
Then I'll add `fourmolu` and `fetchClousure` to my backlog.

PR #713, opened 2023-09-13
PLT-7288 Added handling of redemptions in contract streaming.

bwbush:
Changes in the underlying Runtime behavior resulted in a previously unreachable pattern match now being executed when payouts occur in a contract. Explicit handling on payout steps is added to gracefully deal with this case.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
        - Review required
        - Review not required
    - [ ] Reviewer requested

### DIFF
--- a/marlowe-apps/src/Language/Marlowe/Runtime/App/Channel.hs
+++ b/marlowe-apps/src/Language/Marlowe/Runtime/App/Channel.hs
@@ -185,6 +185,8 @@ runContractAction selectorName eventBackend runInput (RequeueFrequency requeueFr
               (Just _, Nothing, Just _, ContractStreamContinued{csContractStep = ApplyTransaction{}}) -> pure $ M.delete contractId lastSeen
               -- A payout was redeemed from the contract, so there is no need to update the map of most-recent information for contracts still open.
               (Just _, _, Just _, ContractStreamContinued{csContractStep = RedeemPayout{}}) -> pure lastSeen
+              -- A payout was redeemed from the contract after the contract closed, so there is no need to update the map of most-recent information for contracts still open.
+              (Nothing, _, Just _, ContractStreamContinued{csContractStep = RedeemPayout{}}) -> pure lastSeen
               -- FIXME: This should be impossible because a contract must either be created, continuing, closing or redeeming, but diagnose and remedy if this ever occurs.
               (seen, _, _, _) -> do
                 addField event $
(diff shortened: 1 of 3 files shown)


marlowe-apps/src/Language/Marlowe/Runtime/App/Channel.hs
```diff
@@ -171,12 +171,13 @@ runContractAction selectorName eventBackend runInput (RequeueFrequency requeueFr
...
-              (Just seen, Just contract, Just txId) -> pure $ M.insert contractId (seen{lastContract = contract, lastTxId = txId}) lastSeen
-              (Just _, Nothing, Just _) -> pure $ M.delete contractId lastSeen
-              (seen, _, _) -> do
-                -- FIXME: Diagnose and remedy situations if this ever occurs.
+         in case (contractId `M.lookup` lastSeen, contractFromStream cs, transactionIdFromStream cs, cs) of
+              (Nothing, Just contract, Just txId, ContractStreamStart{}) -> pure $ M.insert contractId (LastSeen contractId mempty contract txId mempty) lastSeen
+              (Just seen, Just contract, Just txId, ContractStreamContinued{csContractStep = ApplyTransaction{}}) -> pure $ M.insert contractId (seen{lastContract = contract, lastTxId = txId}) lastSeen
+              (Just _, Nothing, Just _, ContractStreamContinued{csContractStep = ApplyTransaction{}}) -> pure $ M.delete contractId lastSeen
+              (Just _, Nothing, Just _, ContractStreamContinued{csContractStep = RedeemPayout{}}) -> pure lastSeen
```
palas, line 178: Sorry, I don't really understand the context of this. It would help to have a comment that explains what each case represents.

But it looks like 177 would delete the contract from the lastSeen map if it gets closed by ApplyTransaction, and then RedeemPayout wouldn't match if the contract is already closed? But I am not understanding why contracts get deleted from `lastSeen` at all

bwbush: Thanks. I've added thorough comments in 11b49963cd178ace14f20112f25d8fad3ee4065e.

>and then RedeemPayout wouldn't match if the contract is already closed

True, I've fixed this in e00e2b1a1230eb5c71a48c577163dafaf9844125. The fact that the revised `MarloweHeaderSync` protocol never emits such redemptions is a bug, so I'll create a Jira ticket for that.

palas: I was thinking it would be the first element of the tuple from where it would disappear (the result of the dictionary lookup). But as I say, I don't understand the context too well. The comments you added are very useful

bwbush: I added a comment and match on this in 05e9a5cd28c95b8e258f9ad00b529c6188411efe, but the API for `runContractAction` doesn't notify the client of redemptions (either before or after the contract closes), so this is a bit moot.

PR #739, opened 2023-11-01
chore: disable qa instances and preview/mainnet networks

renebarbosafl:
This PR will disable temporarily the instances from QA, mainnet and preview networks.
The goal is to have only applications from demo-preprod running.


shlevy (approved):
Please confirm with the Marlowe team before deploying

PR #756, opened 2023-11-20
Support Windows Cross-Compilation

zeme-wana:


zeme-wana:
`marlowe-cardano` uses `hs-opentelemetry` everywhere and that depends on `unix`.  
We'd have to replace `hs-opentelemetry` (at least) to have windows builds; closing this PR as there is currently no intention to do that.

PR #758, opened 2023-11-21
Added Regression tests for Static Analysis before creating Incremental Static Analysis

ramsay-t:
Recording the results of assessing the current static analysis in the form of some example tests that stimulate each of the features that can currently be detected.

### DIFF
--- a/marlowe-test/test/Spec/Marlowe/StaticAnalysis.hs
+++ b/marlowe-test/test/Spec/Marlowe/StaticAnalysis.hs
@@ -1,0 +1,33 @@
+-----------------------------------------------------------------------------
+--
+-- Module      :  $Headers
+-- License     :  Apache 2.0
+--
+-- Stability   :  Experimental
+-- Portability :  Portable
+--
+
+-----------------------------------------------------------------------------
+
+-- | Tests of Marlowe semantics.
+module Spec.Marlowe.StaticAnalysis (
+  -- * Testing
+  main,
+  tests,
+) where
+
+import Test.Tasty (TestTree, defaultMain, testGroup)
+
+import qualified Spec.Marlowe.StaticAnalysis.Regression (tests)
+
+-- | Entry point for the tests.
+main :: IO ()
+main = defaultMain tests
+
+-- | Run the tests.
+tests :: TestTree
+tests =
+  testGroup
+    "StaticAnalysis"
+    [ Spec.Marlowe.StaticAnalysis.Regression.tests
+    ]
--- a/marlowe-test/marlowe-test.cabal
+++ b/marlowe-test/marlowe-test.cabal
@@ -192,6 +192,8 @@ test-suite marlowe-test
     Spec.Marlowe.Serialization.CoreJson
     Spec.Marlowe.Serialization.ExtendedJson
     Spec.Marlowe.Service.Isabelle
+    Spec.Marlowe.StaticAnalysis
+    Spec.Marlowe.StaticAnalysis.Regression
 
   build-depends:
     , aeson >=2 && <3
@@ -200,6 +202,7 @@ test-suite marlowe-test
     , cardano-api ^>=8.2
     , containers ^>=0.6.5
     , data-default ^>=0.7.1
+    , errors
     , filepath ^>=1.4
     , hint ^>=0.9
     , ilist ==0.4.0.1
--- a/marlowe-test/test/Spec.hs
+++ b/marlowe-test/test/Spec.hs
@@ -21,6 +21,7 @@ import qualified Spec.Marlowe.Plutus (tests)
 import qualified Spec.Marlowe.Semantics (tests)
 import qualified Spec.Marlowe.Serialization (tests)
 import qualified Spec.Marlowe.Service.Isabelle (tests)
+import qualified Spec.Marlowe.StaticAnalysis (tests)
 
 -- | Timeout seconds for static analysis, which can take so much time on a complex contract
 --   that it exceeds hydra/CI resource limits, see SCP-4267.
@@ -49,4 +50,5 @@ tests =
     , Spec.Marlowe.Semantics.tests
     , Spec.Marlowe.Plutus.tests
     , Spec.Marlowe.Service.Isabelle.tests
+    , Spec.Marlowe.StaticAnalysis.tests
     ]


marlowe-test/test/Spec/Marlowe/StaticAnalysis.hs
```diff
@@ -0,0 +1,33 @@
...
+-- | Entry point for the tests.
+main :: IO ()
+main = defaultMain tests
+
+-- | Run the tests.
+tests :: TestTree
+tests =
+  testGroup
+    "StaticAnalysis"
+    [ Spec.Marlowe.StaticAnalysis.Regression.tests
```
bwbush, line 32: This file needs to be added to the repository with `git add`.

ramsay-t: I was fairly sure that I was (although I should have taken out the `main` stuff...)

bwbush:
@ramsay-t see https://github.com/input-output-hk/marlowe-cardano/blob/main/CONTRIBUTING.adoc#changelogs for the changelog requirements. In brief, you'll need to do `scriv create` in the folder of the package where you've made changes, edit the file that `scriv` creates, and then add that file to the repository with `git add`.

bwbush (approved):
Looks good! It's nice to have these regression tests.

marlowe-test/test/Spec/Marlowe/StaticAnalysis.hs
```diff
@@ -0,0 +1,28 @@
...
+
+-----------------------------------------------------------------------------
+
+-- | Tests of Marlowe semantics.
+module Spec.Marlowe.StaticAnalysis (
+  -- * Testing
+  tests,
+) where
+
+import Test.Tasty (TestTree, defaultMain, testGroup)
```
bwbush, line 18:
```suggestion
import Test.Tasty (TestTree, testGroup)
```
Fixes https://github.com/input-output-hk/marlowe-cardano/pull/758/checks?check_run_id=18897246573.

marlowe-test/test/Spec/Marlowe/StaticAnalysis/Regression.hs
```diff
@@ -0,0 +1,213 @@
...
+  assertBool "Empty trace" $ isRight result && isNothing (fromRight Nothing result)
+
+nonPositivePay :: IO ()
+nonPositivePay = do
+  let contract = Pay alicePk (Party alicePk) ada (Constant (-100)) Close
+  result <- getWarning contract
+  assertBool "Detect negative pay" $ maybe False isTransactionNonPositivePay result
+  let contract2 = Pay alicePk (Party alicePk) ada (Constant 0) Close
+  result2 <- getWarning contract2
+  assertBool "Detect zero pay" $ maybe False isTransactionNonPositivePay result2
```
palas, line 93:
```suggestion
  let contract n = Pay alicePk (Party alicePk) ada (Constant n) Close
  result <- getWarning $ contract (-100)
  assertBool "Detect negative pay" $ maybe False isTransactionNonPositivePay result
  result2 <- getWarning $ contract 0
  assertBool "Detect zero pay" $ maybe False isTransactionNonPositivePay result2
```

marlowe-test/test/Spec/Marlowe/StaticAnalysis/Regression.hs
```diff
@@ -0,0 +1,213 @@
...
+transactionShadowing = do
+  let contract = Let "x" (Constant 1) (Let "x" (Constant 2) Close)
+  result <- getWarning contract
+  assertBool "Shadowing x" $ maybe False isTransactionShadowing result
+
+assertionFailed :: IO ()
+assertionFailed = do
+  let contract = Let "x" (Constant 1) (Assert (ValueEQ (UseValue "x") (Constant 2)) Close)
+  result <- getWarning contract
+  assertBool "Detect partial pay" $ maybe False isTransactionAssertionFailed result
```
palas, line 131:
```suggestion
  assertBool "Detect wrong assertion" $ maybe False isTransactionAssertionFailed result
```

marlowe-test/test/Spec/Marlowe/StaticAnalysis/Regression.hs
```diff
@@ -0,0 +1,213 @@
...
+
+-- The current UI drops in instrumentation in the form of assertions and then checks for AssertionFailed
```
palas, line 133: That reminds me there is a way of calling the static analysis `warningsTraceCustom` that takes a flag that allows you to only check for assertions (ignore warnings). And also there is a way to start with a particular state if you want, that is also accessible from `warningsTraceWithState`. The most generic one is `warningsTraceCustom`. We are not currently using the state functionality (or maybe we are for ensuring that minTime is at least 0), but we are using the ignore warnings bit for reachability (I think).

marlowe-test/test/Spec/Marlowe/StaticAnalysis/Regression.hs
```diff
@@ -0,0 +1,213 @@
...
+              ( Assert
+                  ( ValueLT
+                      (UseValue "x")
+                      (Constant 2)
+                  )
+                  (Assert FalseObs Close)
+              )
+          )
+  result <- getWarning contract
+  assertBool "Detect unreachable path" $ maybe False isTransactionAssertionFailed result
```
palas, line 161:
```suggestion
  assertBool "Detect reachable path" $ maybe False isTransactionAssertionFailed result
```

palas (changes requested):
I found a few small issues. But, in general, they are good tests. However, we must have both static analysis live together at the same time and test one against each other, that would be much more effective way of finding bugs (that is also what we did with the previous version of the static analysis), also it will make it much easier to benchmark if we want to measure the improvement (which is also what we did with the previous version of the static analysis)

marlowe-test/test/Spec/Marlowe/StaticAnalysis/Regression.hs
```diff
@@ -0,0 +1,213 @@
...
+          Close
+          (Pay alicePk (Party alicePk) ada (Constant (-100)) Close)
+  result <- warningsTrace (contract 10 11)
+  assertBool "Analyse a contract" $ isRight result
+
+emptyTrace :: IO ()
+emptyTrace = do
+  let contract = Close
+  result <- warningsTrace contract
+  assertBool "Empty trace" $ isRight result && isNothing (fromRight Nothing result)
```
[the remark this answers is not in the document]
palas: Also one for when it is failing for a reason, more readable than `maybe False`

marlowe-test/test/Spec/Marlowe/StaticAnalysis/Regression.hs
ramsay-t: We should record this documentation somewhere :) I saw these functions as I went through it, but there were no uses of some of them, so it was hard to tell what they were for.

palas (approved):

PR #761, opened 2023-11-23
PLT-8746: Generated schema conforms to OpenApi spec

bjornkihlberg:


jhbertra (changes requested):
While these changes make the spec compliant with OpenAPI 3.0 they also make it incorrect. The order of the items matters in these arrays and these changes obscure that.

Unfortunately, it appears there is no way to represent this in OpenAPI 3.0, but the existing schema _is_ valid in OpenAPI 3.1, which does support the tuple form of `items`. I think we should target OpenAPI 3.1 instead of 3.0.

PR #771, opened 2023-12-06
Removed RoleTokenConfig Serialization Alternatives

nhenin:
While testing https://github.com/input-output-hk/marlowe-ts-sdk/pull/54 , some serialization issues have been discovered. 
This is an attempt to fix this issue but by also removing roles configuration alternatives.

### DIFF
--- a/marlowe-runtime-web/changelog.d/20231206_110509_nicolas.henin.md
+++ b/marlowe-runtime-web/changelog.d/20231206_110509_nicolas.henin.md
@@ -1,0 +1,4 @@
+
+### Removed
+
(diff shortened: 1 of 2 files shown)


jhbertra (changes requested):
We only want to remove the alternatives for the object format - removing the string format will break backward compatibility

jhbertra (changes requested):
I have thought about this some more, and this change breaks backwards-compatibility. The format

```
{
  "address": "...",
  "metadata": ...
}
```

Is currently supported by 0.0.5, so this will break that. This change gets a :-1: from me

PR #812, opened 2024-01-29
PLT-9016 Benchmarking of arbitrarily complex queries.

bwbush:
Arbitrarily complex queries can be benchmarked via the benchmarking configuration file.

Pre-submit checklist:
- Branch
    - [ ] Operables are updated with changes to executable command line options.
    - [ ] Deploy charts updated with changes to operables.
- PR
    - [x] Self-reviewed the diff
        - Review required
        - [ ] Substantial changes to code, test, or documentation
        - Review not required
        - [x] Minor changes to non-critical code, documentation, nix derivations, configuration files, or scripts
    - [x] Reviewer requested


marlowe-benchmark/ReadMe.md
```diff
@@ -66,6 +67,7 @@ Here is an example configuration file:
 , "queryParallelism" : 4
 , "queryBatchSize" : 16
 , "queryPageSize" : 256
+, "complexQueries" : {}
```
bwbush, line 70: I'm working on some interest defaults for the complex queries.

bwbush:
> This looks like a good start, but I feel that this configuration gives too much control to the user (they have to give the exact test cases to run). I was hoping we could have some adjustable parameters that would control the generation of the query filters instead. I suppose that we could do this outside the benchmark runner and have that generate the config instead, but it feels a bit like punting on the decision of what our independent variables are in our performance tests for the queries.
> 
> Then again, maybe this is a more flexible architecture that supports a wider range of tests without requiring code changes.

Yes, this is a dilemma. Mostly I'd like to keep the benchmark deterministic because I think there'd be too much variation from randomly-generated queries. Someday we might want to create a generator for random queries of specified complexity, but those could be selected deterministically (perhaps via a seed). For now, I've created https://github.com/input-output-hk/marlowe-cardano/blob/584da201c98e44b2a71adef2f843b2053aa9b9ca/marlowe-benchmark/app/Language/Marlowe/Runtime/Benchmark/Query/Generate.hs#L33-L43, which has a stressfully representative set of queries for `mainnet`, `preprod`, and `preview`. If no configuration is provided on the command-line, then these representative queries are used; otherwise, user-specified ones are used. This keeps determinism, but lets uses define custom benchmarks. It also meets the requirement of full coverage of all sync and query types so that we can detect regressions in resource demand.

jhbertra (approved):

PR #813, opened 2024-01-30
chore: allow QA instance to use cardano-node-873

renebarbosafl:
This PR introduce an update to helm templates from `marlowe-runtime`, allowing instances to select which service of `cardano-node` they will use. 

Right now we have:

1. `cardano-node`  running on version 8.4.0
2. `cardano-node-873` running on version 8.7.3


jhbertra (approved):
thanks!

